PR #122, opened 2014-01-12
Add provider specific default_prefix option.

purpleidea:
This patch adds support for a provider specific default_prefix option.
When set, this value is used instead of the project's dir name as the
prefix for the machines being created.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ This provider exposes quite a few provider-specific configuration options:
 * `id_ssh_key_file` - The id ssh key file name to access Libvirt (eg: id_dsa or id_rsa or ... in the user .ssh directory)
 * `storage_pool_name` - Libvirt storage pool name, where box image and instance snapshots will be stored.
 * `default_network` - Libvirt default network name. If not specified default network name is 'default'.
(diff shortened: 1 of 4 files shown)


joejulian:
Thanks for this feature, @purpleidea. This is perfect for sharing a single libvirt server among several devs.

purpleidea:
@joejulian glad you like it! You might also want to have a look at https://github.com/purpleidea/oh-my-vagrant which is helpful for sharing a particular environment/setup with other devs. Best documentation is on my blog.

Cheers

raghavendra-talur:
I would like to make a little change in this behavior. If the default.prefix is set to '', can we remove the initial _ too? 
For example:
In my vagrantfile I have
config.vm.define "gluster-fedora22-#{i}"
and I set hostname as same with
fedora22.vm.hostname = "gluster-fedora22-#{i}"

But even with libvirt.default_prefix = '', I get the domain name in virt as _gluster-fedora22-#{i} whereas what I would like to have is gluster_fedora22-#{i}.

If nobody has an objection to it, I would like to send a patch for that.

raghavendra-talur:
I have sent a PR at https://github.com/pradels/vagrant-libvirt/pull/488 for the above requirement.

PR #256, opened 2014-10-16
allow preexisting disk volumes via new option

aspiers:


pronix:
merged and implemented logic for creation and removing

Issue #302, opened 2015-02-12
Is Vagrant 1.7 supported?


facastagnini:
As of today, the last version of Vagrant is 1.7.2, and according to the README the plugin is supported only over Vagrant 1.5 and 1.6.

Is this a real constrain or is the README outdated?
Anyone using this plugin with Vagrant 1.7?

Thanks

tkrille:
I'm running vagrant-libvirt with Vagrant 1.7.2 on multiple Debian Wheezy machines. So far I did not encounter any issues, hth.

sciurus:
Thanks for the report @tkrille !

PR #303, opened 2015-02-13
Improving documentation aftert #302

facastagnini:
According to @tkrille vagrant 1.7 is supported.


sciurus:
Thanks for updating this!

PR #305, opened 2015-02-18
Update README: Ruby 2.0 required

jakobadam:
Can't use my Ubuntu 14.04 bundled Ruby 1.9.3. Make a note about 2.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,6 +362,7 @@ $ git clone https://github.com/pradels/vagrant-libvirt.git
 $ cd vagrant-libvirt
 $ bundle install
 ```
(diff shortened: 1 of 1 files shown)


sciurus:
Could you tweak this slightly to say that ruby 2.0.0 should be used? That is the version shipped with vagrant. You could even set this [in the Gemfile](http://bundler.io/gemfile_ruby.html)

```
$ /opt/vagrant/embedded/bin/ruby --version
ruby 2.0.0p598 (2014-11-13 revision 48408) [x86_64-linux]
```

Issue #260, opened 2014-10-24
Port forwarding ignores protocol (udp)


mcaulfield:
Working with a virtual machine that hosts a DHCP server listening on UDP port 67. Trying to forward from host:8067 to guest:67. It looks like vagrant-libvirt sets up a regular SSH TCP tunnel and ignores the protocol attribute.

From my Vagrantfile:
```
config.vm.network "forwarded_port", guest: 67, host: 8067, protocol: 'udp'
```

From ps -elf  | grep 8067:
```
4 S root      4205     1  0  80   0 - 19125 poll_s 10:28 pts/8    00:00:00 ssh -o User=vagrant -o Port=22 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PasswordAuthentication=no -o ForwardX11=no -o IdentityFile=/root/.vagrant.d/insecure_private_key -L localhost:8067:192.168.121.102:67 -N 192.168.121.102
```

infernix:
Since this is an SSH limitation and it's documented, closing.

electrofelix:
Technically it is possible to route UDP over TCP, so it probably could be done by users that need this using the vagrant-trigger plugin to configure a command to setup socat on the host before vagrant up to redirect UDP traffic to a TCP port, have vagrant-libvirt do the TCP port forwarding, and a vagrant-trigger provision to execute the command in the guest to set up for TCP to UDP conversion to the correct port.

http://superuser.com/questions/53103/udp-traffic-through-ssh-tunnel suggests a few different approaches:
* fifo pipes with nc, or socat - programs need to remain running on both guest/host
* tun devices - must be created on start, and removed on destroy on both guest/host
* ssf - secure socket funnelling, dedicated program that is cmdline compatible with ssh, but again requires to available in guest/host.

Be a documentation only effort, I might see if it's possible to put something together to help users put together in their vagrant config to solve this.


One thing I did find that looks very interesting is that it might be possible to allow libvirt to handle all the port forwarding itself.

http://blog.adamspiers.org/2012/01/23/port-redirection-from-kvm-host-to-guest/, indicates that it might be possible to get libvirt itself to set up the necessary iptables rules on the host to do the necessary port forwarding, and support has been present since 0.8.8 (Feb 17 2011).

I'll see if I get a chance to look into this further.

infernix:
Sure, it's technically possible - but the question here is should we support it in vagrant-libvirt and if yes, then how?

[...]

PR #309, opened 2015-02-18
Update README: Only TCP port forwarding supported

facastagnini:
Documenting the inability to forward protocols other than TCP. 
See  #260


sciurus:
Thanks!

Issue #313, opened 2015-02-23
Sanitize memory and cpu 


facastagnini:
I was doing copy&paste to adapt a Vagrantfile using virtualbox to use libvirt and run into a variable type issue.

The Vagrantfile allow you to specify the memory and cpus parameters as strings:
```
  memory =  "4096"
  cpus = "2"
```

So when I copy the original lines and end up with this Vagrantfile config options:
```
    # Domain Specific Options
    bootstrap.vm.provider :libvirt do |domain|
      domain.memory = "4096"
      domain.cpus = "2"
    end
``` 

Run _vagrant up_
```
# vagrant up
Bringing machine 'bootstrap' up with 'libvirt' provider...
==> bootstrap: Creating image (snapshot of base box volume).
==> bootstrap: Creating domain with the following settings...
==> bootstrap:  -- Name:              myserver
==> bootstrap:  -- Domain type:       qemu
==> bootstrap:  -- Cpus:              1
/root/.vagrant.d/gems/gems/vagrant-libvirt-0.0.24/lib/vagrant-libvirt/action/create_domain.rb:101:in `call': undefined method `/' for #<String:0x00000002cdb930> (NoMethodError)
[...]
```

I fix the issue updating the Vagrantfile to:
```
    # Domain Specific Options
    bootstrap.vm.provider :libvirt do |domain|
      domain.memory = 4096
      domain.cpus = 2
    end
``` 

This is not a vagrant-libvirt issue, but it will be nice if we can make those two parameters compatible with the default vagrant provider style, allowing the use of quotes.

I will submit a PR soon.

PR #314, opened 2015-02-23
Sanitize memory and cpu parameters

facastagnini:
Making memory and cpu parameters more resilient. See #313


sciurus:
Looks good; thanks!

PR #328, opened 2015-04-01
Allow the situation where DHCP is not requested but where it is enabled on the virtual network

epienbroek:


lib/vagrant-libvirt/action/create_networks.rb
```diff
@@ -100,8 +100,10 @@ def lookup_bridge_by_name(bridge_name)
         # match what was configured in the vagrantfile
         # since we always enable dhcp for the management network
         # this ensures we wont start a vm vagrant cant reach
+        # Allow the situation where DHCP is not requested (:libvirt__dhcp_enabled == false)
+        # but where it is enabled on the virtual network
         def verify_dhcp
-          unless @options[:dhcp_enabled] == @interface_network[:dhcp_enabled]
+          if @interface_network[:dhcp_enabled] == true && @options[:dhcp_enabled] == false
```
pronix: hello @epienbroek  .
can you explain more ?
I not understand what and for what you change.

epienbroek:
@pronix: In vagrant-libvirt a lookup is done to fine out whether there are virtual networks available which match the network properties mentioned in the Vagrantfile (for example: config.vm.network :private_network, ip: "192.168.5.30"). After a match is found another check is done (the verify_dhcp method) to determine whether the DHCP setting of the already available virtual network matches the DHCP setting which is mentioned in the Vagrantfile (and which is 'true' by default as mentioned in https://github.com/pradels/vagrant-libvirt/blob/HEAD/lib/vagrant-libvirt/util/network_util.rb).

I think this is an invalid assumption. When the Vagrantfile contains the network property ":libvirt__dhcp_enabled => false" it shouldn't matter whether the virtual network contains a DHCP server or not and vagrant-libvirt should not fail on it. The only situation where the verify_dhcp method should fail is when DHCP is requested but isn't configured on a matching already existing virtual network so that is what this commit does

pronix:
@epienbroek got it.
please update also README, you can copy paste your explanation to network section.
after that i will accept it.

epienbroek:
@pronix: done

PR #365, opened 2015-05-12
fix extra disk removal

SimonPe:
I have no idea how that `libvirt_domain.name + "-" + disk[:device] + ".raw"` was suposed to work, but the disk gets created using `disk[:path]` so use that for destruction too

### DIFF
--- a/lib/vagrant-libvirt/action/destroy_domain.rb
+++ b/lib/vagrant-libvirt/action/destroy_domain.rb
@@ -39,10 +39,10 @@ module VagrantPlugins
             env[:machine].provider_config.disks.each do |disk|
               # shared disks remove only manualy or ???
               next if disk[:allow_existing]
-              diskname = libvirt_domain.name + '-' + disk[:device] + '.' + disk[:type].to_s
+              disk[:path] ||= "#{libvirt_domain.name}-#{disk[:device]}.#{disk[:type]}"
               # diskname is uniq
               libvirt_disk = env[:libvirt_compute].volumes.all.select do |x|
-                x.name == diskname
+                x.name == disk[:path]
               end.first
               libvirt_disk.destroy if libvirt_disk
             end


lib/vagrant-libvirt/action/destroy_domain.rb
```diff
@@ -37,9 +37,7 @@ def call(env)
 
             env[:machine].provider_config.disks.each do |disk|
               next if disk[:allow_existing] # shared disks remove only manualy or ???
-              diskname = libvirt_domain.name + "-" + disk[:device] + ".raw"
-              # diskname is uniq
-              env[:libvirt_compute].volumes.all.select{|x| x.name == diskname }.first.destroy
+              env[:libvirt_compute].volumes.all.select{|x| x.name == disk[:path] }.first.destroy
```
pronix: ```disk[:path]``` for my configuration is nil

pronix:
check https://github.com/pradels/vagrant-libvirt/commit/182809e7213cefb985c4db0cfe76c3ef79e4ce24
should be helpful for you.

SimonPe:
@pronix: oh, i missed an `disk[:path] ||= _disk_name(@name, disk)` when reading `create_domain.rb` adding that here to my patch should fix it, will do so and rebase.

SimonPe:
@pronix : fixed it now, care to have a look?

pronix:
can you show vagrantfile where realy require this logic ?

SimonPe:
@pronix: when you manually set the `disk[:path]`, which is allowed by the current api, you also have to use that path when removing the disk again, this is a regression since 0.0.25.
take a look at `create_domain.rb` where the disk creation happens.

SimonPe:
@pronix : could you please motivate that close? The issue still persists, and the fix is rather trivial, so what is the reason you close this?

pronix:
can you show vagrantfile where realy require this logic ?

SimonPe:
@pronix : 
```ruby
    libvirt.storage :file, :size => '40G', :path => 'datadisk.img'
  end
end
```
`:path` is documented in `README.md` on line 307

lib/vagrant-libvirt/action/destroy_domain.rb
```diff
@@ -39,10 +39,10 @@ def call(env)
...
               # shared disks remove only manualy or ???
               next if disk[:allow_existing]
-              diskname = libvirt_domain.name + '-' + disk[:device] + '.' + disk[:type].to_s
+              disk[:path] ||= "#{libvirt_domain.name}-#{disk[:device]}.#{disk[:type]}"
               # diskname is uniq
               libvirt_disk = env[:libvirt_compute].volumes.all.select do |x|
-                x.name == diskname
+                x.name == disk[:path]
               end.first
               libvirt_disk.destroy if libvirt_disk
```
pronix: you can implement use path as second option if ```libvirt_disk``` if null in this string.

SimonPe:
> you can implement use path as second option if libvirt_disk if null in this string.

what I do in the patch is mirroring the logic used to create the disk in the first place. Which is what is needed here

if you change it to your sugestion, and i have by accident an image in my pool wich matches `"#{libvirt_domain.name}-#{disk[:device]}.#{disk[:type]}"` (which could be the reason I have to set `:path` manualy) then with your sugestion you will end up deleting the wrong disk.

pronix:
sorry, i have problems with long phrases.
1. attemp find disk by patern  "#{libvirt_domain.name}-#{disk[:device]}.#{disk[:type]}"
2. if not found - use path.

SimonPe:
> 1. attemp find disk by patern "#{libvirt_domain.name}-#{disk[:device]}.#{disk[:type]}"
> 2. if not found - use path.

this is the wrong way around, don't you get it?
look at `create_domain.rb` on line 77.

pronix:
destroy working in destroy_domain.

pronix:
destroy working in destroy_domain.

SimonPe:
> destroy working in destroy_domain.

I have no idea how to parse that, could you rephrase?

pronix:
1. attemp find disk by patern "#{libvirt_domain.name}-#{disk[:device]}.#{disk[:type]}"
2. if not found - use path.

SimonPe:
> 1. attemp find disk by patern "#{libvirt_domain.name}-#{disk[:device]}.#{disk[:type]}"
> 2. if not found - use path.

like I already told you if you look at `create_domain.rb` line 77, you will see that this is the wrong way.

the correct way (as implemented by the patch) is:

1. try `disk[:path]`
2. if `disk[:path]` is `nil` use `"#{libvirt_domain.name}-#{disk[:device]}.#{disk[:type]}"`

SimonPe:
@pradels @sciurus could any of you take a look at this.

howleyt:
@SimonPe @pronix What's the current state of this patch?

pronix:
patch is crap but issue is fixed.

SimonPe:
@pronix @howleyt : issue still exists and suposed 'fix' is wrong.

SimonPe:
@pronix @howleyt patch is still valid, and suposed 'fix' is wrong

PR #368, opened 2015-05-15
Fixed wrong target spec in 9p-fs template

patricktruebe:
When using a id attribute to specify a synced folder the mount inside the guest fails, since the user-defined id will be used as mount_tag by vagrant-libvirt. Unfortunately this mount_tag has never been announced to libvirt/qemu as target via filesystem.xml.erb.

Inside filesystem.xml.erb the target was always the guestpath which is completely ok when no id was specified by the user and vagrant takes care of it.

### DIFF
--- a/lib/vagrant-libvirt/cap/synced_folder.rb
+++ b/lib/vagrant-libvirt/cap/synced_folder.rb
@@ -43,7 +43,8 @@ module VagrantPlugins
         begin
           # loop through folders
           folders.each do |id, folder_opts|
-            folder_opts.merge!({ :accessmode => "passthrough",
+            folder_opts.merge!({ :target => id,
+								 :accessmode => "passthrough",
                                  :readonly => nil }) { |_k, ov, _nv| ov }
             machine.ui.info "================\nMachine id: #{machine.id}\nShould be mounting folders\n #{id}, opts: #{folder_opts}"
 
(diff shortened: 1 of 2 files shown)


pronix:
can you show vagrantfile for checking your changes?

patricktruebe:
Sure. We are using a hombrew debian-wheezy box, but any debian 7 should do.

The failing mount issued by cachier is

```bash
mount -t 9p [mountopts] 'vagrant-cache' /tmp/vagrant-cache
```

```ruby
Vagrant.configure(2) do |config|
  config.vm.box = "debian-wheezy"
  config.vm.synced_folder './', '/vagrant', type: '9p', disabled: false
  config.vm.provider :libvirt do |lv|
	lv.driver = 'kvm'
	lv.disk_bus = 'scsi'
	lv.memory = 512
	lv.cpus = 2
	lv.nested = true
	lv.keymap = 'de'
  end

  if Vagrant.has_plugin?("vagrant-cachier")
    config.cache.scope = :machine
	config.cache.synced_folder_opts = {
	  type: '9p'
	}
  end
end
```

pronix:
thanks. today or tomorrow i will chek it and accept . schoenes wochenende

patricktruebe:
dankeschön! :-)

PR #373, opened 2015-05-17
Add support for mounting nfs remotely through bridged public network

tsl0922:
This works for me, I need more test to check if this will break existing functions, thanks!

PR #375, opened 2015-05-18
README: Add note for remote nfs synced folders

tsl0922:
The implementation on #373 only works for ssh connection to libvirt, since the management network is only available locally.


tsl0922:
I'll try to push a fix for this. For remote libvirt access, the management network should be created as public, any suggestions on this?

pronix:
I hope noone use vagrant in production and open networks. But security
notification will good

Issue #37, opened 2013-07-08
Support resizing of volume


twouters:
Currently it doesn't seem to be supported to define volumesize for a new guest.
It wouldn't be hard to support though.
In `lib/vagrant-libvirt/action/create_domain_volume.rb` there's this:

    # Virtual size of image. Same as box image size.
    @capacity = env[:machine].box.metadata['virtual_size'] #G

IMO this could easily be extended to support a setting in Vagrantfile that defines the size of the new image.
The defined size shouldn't be smaller than the size specified in the box metadata.

Manual intervention on the guest machine would still be required though. (resizing partitions/filesystems)

infernix:
This seems to have been merged a long time ago; I'm closing this, please create new issues if problems are found related to volume resizing.

electrofelix:
Would automatic resizing of the filesystems be a separate issue?

pronix:
@electrofelix  resizing fs is not problem of virtualisation tool. freebsd, linux, windows have different fs.

infernix:
@electrofelix it would be; the volume resize is just a matter of passing a larger value to fog when the volume is created in a libvirt storage pool. So that's entirely abstracted away from vagrant-libvirt.

Filesystem resizing would involve looking at the volume itself, repartitioning it, resizing the filesystem etc. I don't think libvirt itself supports this at all; I use libguestfs for that type of work.

That type of functionality could be added but if that results in a dependency on libguestfs, that may not be something we'd want. I'd personally consider that to be outside the scope. If you really needed it, you could create a custom box and have that box do the resizing on first (or every) boot.

PR #404, opened 2015-06-26
Support volume size resize configuration

electrofelix:
Allow control of the volume size to be increased from the box default
virtual_size value so that it is possible to use a box configured with a
minimal initial disk size and create virtual guests with larger disk
sizes.

Warn the user and ignore sizes that are less than the box size, and
inform them that a manual filesystem resize will be needed to take
advantage of the additional available disk space.

fixes: #37


pronix:
add description to readme please

electrofelix:
Updated.

PR #413, opened 2015-07-11
add libvirt interface tcp tunnel support.

skamithi:
Useful when configuring Virtualized Switch topologies using Switch VMs like Cumulus
Linux.

vagrant network interface auto_config is disabled in the code. This may be
re-enabled in a future update, once it is better understood how to
auto configure these types of links. All guestOS ports, for now, that are
connected to a tcp tunnel are in a link down state.

TCP tunnels allow guest OSes to exchange STP and LLDP information as
if they are directly connected to each other.

This is not possible with the default virtual switch network mode.

Reference:
https://libvirt.org/formatdomain.html#elementsNICSTCP

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,7 @@ An examples of network interface definitions:
 
   # Guest 2
   config.vm.define :test_vm2 do |test_vm2|
-    test_vm1.vm.network :private_network,
+    test_vm2.vm.network :private_network,
           :libvirt__tcp_tunnel_type => 'client',
           # default is 127.0.0.1 if omitted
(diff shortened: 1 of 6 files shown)


pronix:
thanks.looks interesting, but not useful for me. I hope you test it enough.

skamithi:
Thanks for merging. Tested on Ubuntu and Cumulus Linux VMs. if vagrant tests are required, please let me know and i can look into writing some.  planning on rewriting [my switch demo using ubuntu vms](https://github.com/skamithi/ptmd_demo), I wrote 2 yrs ago,  using vagrant-libvirt instead of hacking myself with ansible.

PR #417, opened 2015-07-13
Allow veryisolated networks

joejulian:
With veryisolated networks, there is no ip address assigned thus
matching the network name and ip address fails. This allows the null ip
address to match a named network when that forward_mode is set.

### DIFF
--- a/lib/vagrant-libvirt/action/create_networks.rb
+++ b/lib/vagrant-libvirt/action/create_networks.rb
@@ -63,9 +63,6 @@ module VagrantPlugins
 
               if @options[:ip]
                 handle_ip_option(env)
-              # in vagrant 1.2.3 and later it is not possible to take this branch
-              # because cannot have name without ip
-              # https://github.com/mitchellh/vagrant/commit/cf2f6da4dbcb4f57c9cdb3b94dcd0bba62c5f5fd
               elsif @options[:network_name]
                 handle_network_name_option(env)
               end
--- a/lib/vagrant-libvirt/action/create_networks.rb
+++ b/lib/vagrant-libvirt/action/create_networks.rb
@@ -116,13 +116,16 @@ module VagrantPlugins
         # @available_networks should be filled before calling this function.
         def handle_ip_option(env)
           return if !@options[:ip]
+          net_address = nil
+          if @options[:forward_mode] != 'veryisolated'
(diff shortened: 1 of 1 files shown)


pronix:
hello
I not use this type of network. can you show libvirt doc where described situation without ip ?
thanks

joejulian:
The last example on https://libvirt.org/formatnetwork.html shows a network with no forward_mode and no address.

When you create a veryisolated network, https://github.com/pradels/vagrant-libvirt/blob/master/lib/vagrant-libvirt/templates/private_network.xml.erb#L13-L28 produced a very similar network to that example with no forward mode and no ip.

pronix:
as i understand - it should be optional ?

joejulian:
I don't think so. It appears that forward_mode "veryisolated" differs from "none" in that the host does not have an address assigned.

R2ZER0:
I too am facing this issue, trying to create a veryisolated network without a gateway address. 
I agree with joejulian that this should be the behavior of veryisolated - but we should probably make it backwards-compatible with people using the existing setup.
Perhaps add an options to:
 * Override matching of networks to use the name instead of IP subnet - :libvirt__match_by_name => true (default false)
 * Enable/disable a gateway address - :libirt__enable_gateway => false  (default true)

Then this would give us the desired behaviour in a flexible way. What do you think?
In my own fork ( https://github.com/R2ZER0/vagrant-libvirt/commit/f0904941994403f04c40fb09f3dd29b1fa88d371 ) i implemented this behavior without the options just to get it working, though i'm sure it can be done it a less drastic way than i have!

joejulian:
I'm going to close this PR for now as I've found a problem still exists with tearing down the network.

I'll also put together a proper issue that explains the problem so the solution can be more clearly understood.

R2ZER0:
Hi @joejulian,

So I am testing your latest commit ( https://github.com/joejulian/vagrant-libvirt/commit/9b7c9def3c54b75653346532ded9c5aece96ad33 ), and am trying to create a network with two VMs with :auto_config *enabled*, but am getting the error:
http://pastebin.com/raw.php?i=88pBRBWc

Here is the Vagrantfile that was used: http://pastebin.com/FfSUvCJR

Not sure what's going on there?

Thanks

joejulian:
Thanks @R2ZER0. Fixed.

PR #422, opened 2015-07-20
Really Limit p9 tag to 31 bytes

### LINKED ISSUE #323: 9p sync can produce mount_tag which is too long
If you have a folder sync with a path longer than 31 bytes, creation will fail due to the mount_tag being too long.

Example failing config:
```
myvm.vm.synced_folder "/tmp/blah", "/etc/puppet/environments/vagrant", type: '9p'
```

Error:
```
Call to virDomainCreateWithFlags failed: internal error: process exited while connecting to monitor: qemu-system-x86_64: -device virtio-9p-pci,id=fs0,fsdev=fsdev-fs0,mount_tag=/etc/puppet/environments/vagrant,bus=pci.0,addr=0x6: mount tag '/etc/puppet/environments/vagrant' (32 bytes) is longer than maximum (31 bytes)
```

grahamhayes:
Libvirt p9 file system passthrough requires the tag be less than
31 bytes. Using the mount point as the tag can easily exceed this.

In KVM passthough the "target" attribuite in the XML is acutally the
mount tag.

Fixed the mounting script to use the same MD5 to mount the right
filesystem

Fixes #323

### DIFF
--- a/lib/vagrant-libvirt/cap/mount_p9.rb
+++ b/lib/vagrant-libvirt/cap/mount_p9.rb
@@ -17,7 +17,7 @@ module VagrantPlugins
             machine.communicate.sudo("mkdir -p #{expanded_guest_path}")
 
             # Mount
-            mount_tag = Digest::MD5.new.update(name).to_s[0,31]
+            mount_tag = Digest::MD5.new.update(opts[:hostpath]).to_s[0,31]
 
             mount_opts="-o trans=virtio"
             mount_opts += ",access=#{opts[:owner]}" if opts[:owner]
--- a/lib/vagrant-libvirt/cap/synced_folder.rb
+++ b/lib/vagrant-libvirt/cap/synced_folder.rb
@@ -1,6 +1,7 @@
 require 'log4r'
 require 'ostruct'
 require 'nokogiri'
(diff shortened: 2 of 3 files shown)


lib/vagrant-libvirt/cap/mount_p9.rb
```diff
@@ -17,12 +17,11 @@ def self.mount_p9_shared_folder(machine, folders)
             machine.communicate.sudo("mkdir -p #{expanded_guest_path}")
 
             # Mount
-            mount_tag = Digest::MD5.new.update(name).to_s[0,31]
+            mount_tag = Digest::MD5.new.update(opts[:hostpath]).to_s[0,31]
 
             mount_opts="-o trans=virtio"
             mount_opts += ",access=#{opts[:owner]}" if opts[:owner]
             mount_opts += ",version=#{opts[:version]}" if opts[:version]
-            mount_opts += ",#{opts[:mount_opts]}" if opts[:mount_opts]
```
pronix: why not allowed custom options ?

grahamhayes: Because I missed it. Let me fix that

PR #425, opened 2015-07-24
Use per device boot order when boot_order is set

infernix:
This allows for network boot with any of the NICs that are defined instead of
only from the primary NIC.


infernix:
The problem with the existing code is that when boot_order is set to 'network', it will only boot from the primary NIC, and will then try to boot off the first hard drive. With this change it will try to boot off of any defined NIC before it will boot from the first hard drive.

I considered implementing boot order specification in the Vagrant file but did not see a benefit there over this implementation, and this works without changes to existing Vagrantfiles.

pronix:
add description to readme please. better with example.

infernix:
Updated readme.

There are no examples needed because the given example in https://github.com/pradels/vagrant-libvirt/commit/56818642e567ade160515074fb68f59c5bd392d1 is unchanged.

The old behaviour:
* set ```domain.boot = network```
* at ```vagrant up``` the first NIC is tried with PXE boot, then first harddisk is tried - other NICS are never tried

The new behaviour:
* set ```domain.boot = network```
* at ```vagrant up```, all NICs are tried with PXE boot, then first harddisk is tried

pronix:
thanks

PR #429, opened 2015-07-28
Add a forwarded_ports capability

bexelbie:
This reports the forwarded_ports in use by the VM

### DIFF
--- a/lib/vagrant-libvirt/action.rb
+++ b/lib/vagrant-libvirt/action.rb
@@ -290,6 +290,14 @@ module VagrantPlugins
         end
       end
 
+      def self.action_forward_ports
+        Vagrant::Action::Builder.new.tap do |b|
+          b.use ConfigValidate
+          b.use ConnectLibvirt
+          b.use ForwardPorts
+        end
+      end
+
       def self.action_read_mac_addresses
         Vagrant::Action::Builder.new.tap do |b|
           b.use ConfigValidate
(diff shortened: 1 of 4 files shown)


pronix:
is it implemented in fog ?

bexelbie:
@pronix I honestly do not know.  I am new to vagrant and libvirt.  I needed this for parity with a virtualbox image.  Can you point me toward fog and I am happy to look?

pronix:
this patch working for you ?

bexelbie:
I tested it using a plugin I am working on and it returned the values as expected.

pronix:
ports are forwarded ?

bexelbie:
All this does is allow the forwarded_ports capability to be used to report what ports are forwarded .. port forwarding was already working

bexelbie:
This is similar to a cap provided by virtualbox

pronix:
I not understand what doing your patch.
add description to readme, please

electrofelix:
I believe the idea is to allow other plugins to query which ports have been forwarded based on checking the provider capabilities first to see if it can provide this information to do something with them, rather than checking if the provider is libvirt do x, virtualbox do y, docker do z, etc.

bexelbie:
@electrofelix has the correct answer.  However, I have discovered a bug in my patch and am going to be traveling for a few days.  So please give me some time to rework it.

PR #433, opened 2015-07-29
Enable travis CI testing of github PRs

electrofelix:
Add travis CI and support testing against a matrix of vagrant versions.

Initial framework to enable easier checking.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+---
+language: ruby
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get install -y libvirt-dev
+  - gem install bundler --version $BUNDLER_VERSION
+install: bundle _${BUNDLER_VERSION}_ install
+script: bundle _${BUNDLER_VERSION}_ exec rspec --color --format documentation
+notifications:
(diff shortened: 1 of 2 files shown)


electrofelix:
Forgot to mention, idea is pretty much lifted from oscar-stack/vagrant-auto_network will minimal adjustments.

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+script: bundle _${BUNDLER_VERSION}_ exec rspec --color --format documentation
+notifications:
+  email: false
+rvm:
+  - 2.0.0
+env:
+  global:
+    - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
+  matrix:
+    - VAGRANT_VERSION=v1.5.4 BUNDLER_VERSION=1.5.3
```
pronix: I think only 1.7.x make sense

electrofelix: Since the README states it works on 1.5, 1.6 & 1.7 seems like it should be at least have a limited test on them and then if it breaks, when removing from the matrix can also update the README to remove the versions listed as suported.

PR #436, opened 2015-07-31
allow query of ssh_info in parallel from other machines or user

### LINKED ISSUE #428: ssh_info and state can trigger machine locked errors when called in parallel from other machines
Previously logged as #420, an initial attempt at fixing with pr #421 solved the problem of the read_ssh_info and read_state being run against a machine that was locked, however they could still trigger exceptions in vagrant about the machine already being locked by another process.

(originally posted in #420 and edited for reposting under a separate issue)

I've done some further debugging and I believe I've established that #420 is not completely fixed, and traced the underlying issue.

With change #421 you now get the following message:

        Vagrant can't use the requested machine because it is locked! This
        means that another Vagrant process is currently reading or modifying
        the machine. Please wait for that Vagrant process to end and try
        again. Details about the machine are shown below:

        Name: ....
        Provider: libvirt

It turns out that there are some calls being made in both the read_ssh_info and read_state actions that attempt to modify part of the machine that vagrant will explicit lock and throw an error on.

The calls are:
`machine.state.id` at https://github.com/pradels/vagrant-libvirt/blob/master/lib/vagrant-libvirt/action/read_ssh_info.rb#L23
`machine.id = nil` at https://github.com/pradels/vagrant-libvirt/blob/master/lib/vagrant-libvirt/action/read_ssh_info.rb#L30
`machine.id = nil` at https://github.com/pradels/vagrant-libvirt/blob/master/lib/vagrant-libvirt/action/read_state.rb#L41
`machine.id = nil` at https://github.com/pradels/vagrant-libvirt/blob/master/lib/vagrant-libvirt/action/read_state.rb#L47

These are not thread safe and can result in an exception being thrown when running them with multiple machines in parallel.

Calling `machine.state.id` calls
https://github.com/mitchellh/vagrant/blob/master/lib/vagrant/machine.rb#L489 which calls
https://github.com/mitchellh/vagrant/blob/master/lib/vagrant/machine.rb#L497 which attempts to lock the machine and on failure will raise the exception at
https://github.com/mitchellh/vagrant/blob/master/lib/vagrant/machine_index.rb#L133

This would be fine, as long as ssh_info is only ever called on each machine individually, but instead it can be called by the ansible provisioner on all active machines, or by anyone wishing to ssh into the machine to watch what is happening.

Note the ansible provisioner for vagrant will attempt to call ssh_info on all active machines from each machine: https://github.com/mitchellh/vagrant/blob/master/plugins/provisioners/ansible/provisioner.rb#L125-L128

Similarly the problem when you attempt to set the `machine.id = nil` will result in calling
https://github.com/mitchellh/vagrant/blob/master/lib/vagrant/machine.rb#L263 which during it's execution will call:
https://github.com/mitchellh/vagrant/blob/master/lib/vagrant/machine.rb#L326-L327 due value being `nil` and subsequently result in the following being called:
https://github.com/mitchellh/vagrant/blob/master/lib/vagrant/machine.rb#L326-L327

Again, this is all fine when being called from the thread that owns the machine, but any other machines attempting to execution actions that do discovery of ssh_info or state of the other machines risk causing an exception being thrown about the machine being locked by another process.

I'm currently experimenting with a complete fix for this, which entails changing the provider.ssh_info and provider.state calls to use a driver class to talk to libvirt instead of using actions and also to avoid any calls to `machine.state` or attempting to set `machine.id`.

Current thought process is to create a separate driver class, move the libvirt connection management into this class along with calls to retrieve the ip address and current machine state. Remove all use of the ConnectLibvirt action replacing with a call to `machine.provider.driver.connection` instead (similar to the docker provider), and moving the remaining parts of ReadSSHInfo and ReadState actions into the provider class to in turn call the driver for the specific info.

`machine.state.id` from within ssh_info will be replaced with just `state.id` since it will now have access via the provider class, and there is no need to ever modify `machine.id` directly since vagrant will wipe that if calling `provider.state` returns `:not_created`.

I believe also that I can achieve some clean up around the states and remove the loop that waits for the state to stabilize from 'shutting-down' to 'terminated' since there really is no problem with returning a `:shutting_down` state similar to the other states that vagrant can return.

electrofelix:
This allows ssh_info to be queried in parallel as used by the ansible provisioner and also allows the user to ssh into a machine even when provisioners are being executed by being able to return the ssh_info.

It also avoids resetting of the machine id directly and instead allows the vagrant core to manage it based on the state returned by the driver. Also placing the logic in a driver class should make this easier to write unit tests.

Fixes #428


pronix:
thanks. huge pr. I will test it

PR #439, opened 2015-08-03
Include custom UEFI loader

gbriones1:
Domain XML has loader tag to read specific OVMF.fd path

Signed-off-by: Gabriel Briones Sayeg <gbriones.gdl@gmail.com>


pronix:
add description to README please

PR #440, opened 2015-08-05
Added ability to specify inputs for the VM via the Vagrantfile

d0c-s4vage:
Instead of a hard-coded

```xml
<input type="mouse" bus="ps2"/>
```

The user can now generate other types of inputs with

```ruby
libvirt.input :type => "tablet", :bus => "usb"
```

resulting in domain xml

```xml
<input type="tablet" bus="usb"/>
```

### DIFF
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -76,6 +76,9 @@ module VagrantPlugins
       attr_accessor :disks
   	  attr_accessor :cdroms
 
+	  # Inputs
+	  attr_accessor :inputs
+
       def initialize
         @uri               = UNSET_VALUE
         @driver            = UNSET_VALUE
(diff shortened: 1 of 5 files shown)


lib/vagrant-libvirt/config.rb
```diff
@@ -165,6 +171,21 @@ def _get_cdrom_dev(cdroms)
         raise "Only four cdroms may be attached at a time"
       end
 
+      def input(options={})
+        if options[:type] == nil or options[:bus] == nil
+          raise "Input type AND bus must be specified"
```
pronix: how possible not specified input if we have 
```@inputs = [{:type => "mouse", :bus => "ps2"}] if @inputs == UNSET_VALUE```

**Comment by d0c-s4vage** [comment]
hrmm good point

**Comment by d0c-s4vage** [comment]
Ah, I remember now. My thinking was that you would always want at least the default input of

```xml
<input type="mouse" bus="ps2"/>
```

Even if you didn't specify any inputs

PR #442, opened 2015-08-10
Make boxes optional and move boot ordering into a separate class

infernix:
The use case for this is a Vagrantfile in which one can deploy a box with a PXE server and boot PXE clients from it without having to manage those PXE clients outside Vagrant.

The assumption is that a "no box" VM is not managed in the Vagrant process at all. It is created but other than that, Vagrant is hands-off. There is some more discussion on this here: https://github.com/mitchellh/vagrant/issues/4487

It additionally moves the boot ordering into a separate class where the ordering is done just before a domain is created or started. Due to an issue with libvirt (https://bugzilla.redhat.com/show_bug.cgi?id=1248514), it is not possible to specify boot ordering when attaching a device to a domain that does not already have boot ordering. This is therefore the only way to do this currently.

### DIFF
--- a/lib/vagrant-libvirt/plugin.rb
+++ b/lib/vagrant-libvirt/plugin.rb
@@ -23,7 +23,7 @@ module VagrantPlugins
         Config
       end
 
-      provider('libvirt', parallel: true) do
+      provider('libvirt', parallel: true, box_optional: true) do
         require_relative 'provider'
         Provider
       end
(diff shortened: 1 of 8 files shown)


lib/vagrant-libvirt/plugin.rb
```diff
@@ -23,7 +23,7 @@ class Plugin < Vagrant.plugin('2')
         Config
       end
 
-      provider('libvirt', parallel: true) do
+      provider('libvirt', parallel: true, box_optional: true) do
```
electrofelix: I don't believe this will have any effect for vagrant 1.5, won't cause any compatibility issues AFAIK, will just be ignored. Maybe just mention in the docs that this requires vagrant 1.6+ to be usable?

infernix: Looks correct, commit is here: https://github.com/mitchellh/vagrant/commit/61ffa5313451c3705fe9e25c5d0167fae82fc865

1.5.0 will just error out if there is no box defined because it's not supported. I don't think we can print an error in vagrant-libvirt before vagrants validate runs. Will update the docs.

PR #464, opened 2015-09-24
Accept both symbol and string for network type

mandre:
Previously, following Vagrant documentation to setup private DHCP
network failed with:

  config.vm.network "private_network", type: "dhcp"

Now libvirt provider accepts both "dhcp" and :dhcp

### DIFF
--- a/lib/vagrant-libvirt/action/create_networks.rb
+++ b/lib/vagrant-libvirt/action/create_networks.rb
@@ -63,7 +63,7 @@ module VagrantPlugins
 
               if @options[:ip]
                 handle_ip_option(env)
-              elsif @options[:type] == :dhcp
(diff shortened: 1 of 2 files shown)


lib/vagrant-libvirt/action/create_networks.rb
```diff
@@ -63,7 +63,7 @@ def call(env)
 
               if @options[:ip]
                 handle_ip_option(env)
-              elsif @options[:type] == :dhcp
+              elsif @options[:type].to_s == 'dhcp'
```
pronix: why no ```.to_sym```

mandre: because `to_sym` method doesn't exist for NilClass, but `to_s` does.

so this code works in all cases, even for nil values.

PR #468, opened 2015-09-25
Add Table of Contents

Paulche:
Readme file is large, so navigation may be quite compilated.
Table of contents improves navigation but adds a little work when the content is changed.

Table of Contents was generated with http://doctoc.herokuapp.com/


pronix:
merged. but it is not done.

PR #470, opened 2015-09-27
Issue 411

mkutsevol:
This patch is based on PR#469, so it includes its commits. 
I think it's better to first merge PR#469. 

* Added the suspend_mode domain specific option. Now it can perform a
managed save/resume. Switching this domain specific option can be done
when the domain is supended, it will correctly resume the domain.

Kind regards,
Maxym


Et7f3:
Have you an example on how to set it ?

PR #471, opened 2015-09-28
Fix table of contents

Paulche:
In #468 I added Table of Contents but it doesn't actually work.
This commit adds desired functionality.

To keep TOC up to date:
```shell
doctoc  README.md
```


pronix:
thanks.fixed by previous PR

PR #474, opened 2015-10-05
If the disk size is specified as a decimal number the generated metad…

jcsmith:
…ata.json file is invalid resulting in a broken vagrant box.  Wrapped this in the awk int() function to only print the integer part of the disk size.

### DIFF
--- a/tools/create_box.sh
+++ b/tools/create_box.sh
@@ -81,6 +81,9 @@ fi
 
 cd "$TMP_DIR"
 
+#Using the awk int function here to truncate the virtual image size to an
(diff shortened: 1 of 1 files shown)


pronix:
add please comment to code 
it is important information

jcsmith:
Added a comment.  Let me know if its what you wanted to see or if you'd rather it be worded differently.

jcsmith:
Thanks.

--
Josh Smith
KD8HRX

Email/jabber: juicewvu@gmail.com
Phone: 304.237.9369(c)

PR #495, opened 2015-10-29
Check host only ip when ssh_info is ready

leoh0:
Sometimes ssh is not ready then ssh_info is nil.
So, we need to check that.
And I fix a one indentation line.

### DIFF
--- a/lib/vagrant-libvirt/action/prepare_nfs_settings.rb
+++ b/lib/vagrant-libvirt/action/prepare_nfs_settings.rb
@@ -39,7 +39,7 @@ module VagrantPlugins
         #
         # @param [Machine] machine
         # @return [String]
-	def read_host_ip(ip)
+        def read_host_ip(ip)
           UDPSocket.open do |s|
             if(ip.kind_of?(Array))
               s.connect(ip.last, 1)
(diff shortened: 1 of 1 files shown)


lib/vagrant-libvirt/action/prepare_nfs_settings.rb
```diff
@@ -54,9 +54,11 @@ def read_host_ip(ip)
         # @param [Machine] machine
         # @return [String]
         def read_machine_ip(machine)
-          # check host only ip
-          ssh_host = machine.ssh_info[:host]
-          return ssh_host if ping(ssh_host)
+          # check host only ip when ssh_info is ready
+          if machine.ssh_info
```
pronix: where is handle else ?

leoh0: I think it was my mistake.

leoh0:
I test openstack kolla by using vagrant-libvirt. 
They use 3 ethernet interface.
( https://github.com/openstack/kolla/blob/master/dev/vagrant/Vagrantfile#L49-L50 )
But, they use same bridge for eth0 and eth2.
So, arp can not address eth0's mac address.

So, I think that is my real problem.

PR #505, opened 2015-11-09
Allow disabling guest-to-guest ipv6 comms in libvirt networks

brk3:
Background
----------
The default private_network template we supply has the following tag:

    <network ipv6='yes'>

This enables ipv6 on guest interfaces for guest-to-guest communication.
This causes a problem on hosts that either do not have ipv6 enabled or
configured correctly, resulting in an error on 'vagrant up' as follows:

    Error while activating network: Call to virNetworkCreate failed: internal error: Failed to apply firewall rules /usr/sbin/ip6tables --table filter --insert FORWARD --in-interface virbr1 --jump REJECT: ip6tables v1.4.21: can't initialize ip6tables table `filter': Address family not supported by protocol
    Perhaps ip6tables or your kernel needs to be upgraded.

Reading here:
http://libvirt.org/git/?p=libvirt.git;a=commitdiff;h=705e67d40b09a905cd6a4b8b418d5cb94eaa95a8
would suggest the above option should really be off by default, however,
I don't want to break existing behaviour or setups. This patch adds two
new config items, one for the default management network, and one for
additional user defined private networks.

Usage
-----
To disable ipv6 for the managment network:

        libvirt.management_network_guest_ipv6 = "no"
    end

To disable for a custom private network:

    config.vm.network "private_network", ip: "192.168.33.10", libvirt__guest_ipv6: "no"

Extra Notes:
------------
Also note, if you get hit by the above error and want to apply the above
fix, you will need to clear out the management network manually:

    $ sudo virsh net-list --all
    Name                 State      Autostart     Persistent
    ----------------------------------------------------------
    default              active     yes           yes
    vagrant-libvirt      inactive   no            yes

    $ sudo virsh net-undefine vagrant-libvirt
    Network vagrant-libvirt has been undefined

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,6 +396,8 @@ starts with 'libvirt__' string. Here is a list of those options:
 * `:libvirt__tunnel_local_ip` - Sets the local IP used by the udp tunnel
     interface type. It populates the ip entry of the `<local address=XXX">` section of
     the interface xml configuration. _(This feature only works in libvirt 1.2.20 and higher)_
+* `:libvirt__guest_ipv6` - Enable or disable guest-to-guest IPv6 communication.
+    See [here](https://libvirt.org/formatnetwork.html#examplesPrivate6), and [here](http://libvirt.org/git/?p=libvirt.git;a=commitdiff;h=705e67d40b09a905cd6a4b8b418d5cb94eaa95a8) for for more information.
 * `:mac` - MAC address for the interface.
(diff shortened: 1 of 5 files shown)


garrettr:
This is currently a useful workaround when using vagrant-libvirt on [Subgraph OS](https://subgraph.com/sgos/), because Subgraph currently does not support IPv6.

PR #524, opened 2015-12-13
Removed direction for deprecated Vagrant option

lwh:
Removed and tested it, works fine
https://github.com/pradels/vagrant-libvirt/issues/509


pronix:
thanks. but better use think code https://github.com/litaio/development-environment/pull/2/files

lwh:
When you are running the dev version doesn't the Gemfile handle loading it without adding it to the Vagrantfile?

PR #558, opened 2016-02-17
Make it possible to set the path to the device model emulator.

fred-a-kemp:
On some systems (e.g. Debian) omitting this setting results in the
following error:

    Error while creating domain: Error saving the server: Call to
    virDomainDefineXML failed: invalid argument: could not find
    capabilities for domaintype=kvm

Which has been reported as:

    https://github.com/pradels/vagrant-libvirt/issues/539


fred-a-kemp:
I don't think this is the real solution for the problem, but at least it allows to work around it by setting emulator_path appropriately.

infernix:
Looks ok, can you add an update to the README.md with an example for this flag?

fred-a-kemp:
I hope that is good enough. :)

pronix:
@infernix accepted ?

infernix:
@pronix I am travelling, will test by the end of the week.

infernix:
@fred-a-kemp I've tried this in a few ways but it doesn't seem to pick it up. My ARM libvirt definition keeps choosing the default emulator (/usr/bin/qemu-system-arm in my case) over whatever I specify in emulator_path.

Could you provide a working vagrantfile?

fred-a-kemp:
Strange. Have you looked at the resulting libvirt XML? Is the `<emulator>` element there?

This is the exact Vagrant file I tested this with: http://git.tails.boum.org/tails/plain/vagrant/Vagrantfile?h=feature/6354-vagrant-libvirt

If you want to run it you should probably just drop the `require_relative` line and change the `domain.memory` line to something suitable.

infernix:
Yeah, not working in any way at all. Example:

```ruby
ENV['VAGRANT_DEFAULT_PROVIDER'] = 'libvirt'

  config.vm.box = 'tails-builder-amd64-jessie-20160226'
  config.vm.box_url = "http://dl.amnesia.boum.org/tails/project/vagrant/#{config.vm.box}.box"
  config.vm.box_download_checksum_type = "sha256"
  config.vm.box_download_checksum =
    'a63298b7cae7cbc2ad3e2c01763a11b29ee8ac879398e007ae22857da5b2c6de'
  config.vm.provision "shell", inline: <<-EOF
    sudo /vagrant/provision/setup-tails-builder
  EOF
    domain.driver = 'kvm'
    domain.connect_via_ssh = false
    domain.machine_arch = 'x86_64'
    domain.machine_type = 'q35'
    domain.emulator_path = '/usr/bin/qemu-system-i686'
    domain.memory = 1024
    cpus = 2
    domain.cpus = cpus unless cpus.nil?
  end
end
```

It doesn't matter what I pass in emulator_path - not even  `/bin/true` works.

What I'm actually trying to do is to pass it a custom compiled qemu-system-arm and if that doesn't work I'm not sure what the point is of emulator_path as that is what you'd expect it to do.

Can you test it on your side with /bin/true or some other random path and see what happens? What version of libvirt?

infernix:
I hardcoded `<emulator>/bin/true</emulator>` in `domain.xml.erb` and it still doesn't work, so it's not a vagrant specific issue. Either something filters it out or libvirt doesn't accept it/ignores it when the xml is being passed on. Very strange, needs some debugging.

infernix:
@fred-a-kemp doh, that was silly of me. Please apply this patch to your PR:

```diff
diff --git a/lib/vagrant-libvirt/action/create_domain.rb b/lib/vagrant-libvirt/action/create_domain.rb
index 0c34674..5be6bef 100644
--- a/lib/vagrant-libvirt/action/create_domain.rb
+++ b/lib/vagrant-libvirt/action/create_domain.rb
@@ -45,6 +45,7 @@ module VagrantPlugins
           @domain_volume_cache = config.volume_cache
           @kernel = config.kernel
           @cmd_line = config.cmd_line
+          @emulator_path = config.emulator_path
           @initrd = config.initrd
           @graphics_type = config.graphics_type
           @graphics_autoport = config.graphics_autoport
diff --git a/lib/vagrant-libvirt/config.rb b/lib/vagrant-libvirt/config.rb
index 82f517f..b9eafc8 100644
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -444,6 +444,7 @@ module VagrantPlugins
         @tpm_type = 'passthrough' if @tpm_type == UNSET_VALUE
         @tpm_path = nil if @tpm_path == UNSET_VALUE
         @nic_adapter_count = 8 if @nic_adapter_count == UNSET_VALUE
+        @emulator_path = nil if @emulator_path == UNSET_VALUE
 
         # Boot order
         @boot_order = [] if @boot_order == UNSET_VALUE
```

Then it should work as expected.

fred-a-kemp:
Done, thanks for the fix!

Now I wonder why this worked at all for me previously. In fact, without your fix the `<emulator>` element is present but set to `/usr/bin/kvm` (which works!) instead of the actual value I set. However, I remember looking at the XML right before submitting the PR, and then it was set as I liked it. I was playing around with different versions of many of the dependencies (libvirt, fog) at the the time, which might explain this, but I'd rather not try to get to the bottom of this if you're happy with the current state of the branch. :)

infernix:
Yeah this looks good now, @pronix ready to merge. FWIW in the process I found some additional work to do for supporting ARM CPUs which I will send a PR for soon-ish.

Thanks :+1:

PR #565, opened 2016-02-23
RFC: Drop incorrect NetworkNotAvailableError assumption for private networks.

dgoodwin:
This is probably more of an RFC than a real PR but I recently started having issues after an upgrade, my vms would no longer vagrant up. I use a custom libvirt network so I can keep them at a reliable address, and it was running afoul of this check that I'm removing with this commit. Taking it out lets me create/start my vms normally again.

Looking at the code it doesn't seem quite right to me, custom networks could easily already exist and will have an address as far as I can tell. Could easily be missing something but thought I would check and see if this should actually be removed.

More details on my issue (including Vagrantfile, custom network definition, and full error) in this bug report: https://bugzilla.redhat.com/show_bug.cgi?id=1310802

### DIFF
--- a/lib/vagrant-libvirt/action/create_networks.rb
+++ b/lib/vagrant-libvirt/action/create_networks.rb
@@ -210,12 +210,6 @@ module VagrantPlugins
           network = lookup_network_by_name(@options[:network_name])
           @interface_network = network if network
 
-          # if this interface has a network address, something's wrong.
-          if @interface_network[:network_address]
-            raise Errors::NetworkNotAvailableError,
-                  network_name: @options[:network_name]
-          end
-
           # Do we need to create new network?
           unless @interface_network[:created]
             @interface_network[:name] = @options[:network_name]


pronix:
@dgoodwin  explain please , why this code is wrong ? step by step please :)

dgoodwin:
I'm entirely not sure it's wrong but it seems it may be in my case.

I have a custom defined libvirt network created with commands like:

virsh net-define /etc/libvirt/qemu/networks/openshift.xml
virsh net-start openshift
virsh net-autostart openshift

The openshift.xml can be viewed here: https://bugzilla.redhat.com/attachment.cgi?id=1129409

I want to use this custom network for my vagrant vms as I like to be able to destroy and recreate them but always have them at a consistent IP address so it's effectively the same system, this is because I do a lot of cluster related work and it helps at lot to have the VMs reliably addressed.

So I specify in my vagrant file something like:

    origin.vm.network :private_network,
      :libvirt__network_name => 'openshift',
      :mac => '52:54:00:96:a1:a1'

As of vagrant-libvirt-0.0.32 (I think, this wasn't a problem until very recently when this rpm was upgraded) when this block of code triggers, it's looking for a network with name openshift, which exists, and the interface has a network address. Some debug logging on @interface_network prints:

DEBUG create_networks: Hash: {:name=>"openshift", :ip_address=>"192.168.133.1", :netmask=>"255.255.255.0", :network_address=>"192.168.133.0", :dhcp_enabled=>true, :bridge_name=>"virb
r1", :created=>true, :active=>true, :autostart=>true, :libvirt_network=>#<Libvirt::Network:0x0055f0afb3d058 @connection=#<Libvirt::Connect:0x0055f0afa661e8>>}

Because it has a network_address it errors out, I don't understand the reasoning for the check in the first place, but removing it allows me to vagrant up my systems again. It seems reasonable that the network interface will have an address, maybe even commonly. 

I cannot say for sure it's wrong but it feels like it may be in some cases.

infernix:
I don't think this check is wrong, but admittedly its intended purpose isn't exactly clear to me either.

Why don't you specify an `:ip => "x.x.x.x"`here? I think if you do that, which is the expected form, then it would just work fine?

infernix:
@dgoodwin any comments on this? Thanks!

dgoodwin:
Apologies @infernix I totally missed that you'd replied. This does seem to work! Thanks! Closing now.

PR #567, opened 2016-03-01
Fix "machine_type" documentation - was "machine" in README.md

lpenz:


infernix:
Thanks! I ran into the same recently but forgot about it :+1:

PR #577, opened 2016-03-14
Add cpu_model and cpu_fallback.

infernix:
This allows for specific CPU selection and enforcement when cpu_mode =
custom, which is important when emulating CPUs of different architecture
than the host.


infernix:
Yeah, I'm running into a lot of issues to get ARM support ready. Let's not merge this yet, I will add this to another series of commits.

PR #594, opened 2016-04-27
Skip delete cdrom

electrofelix:
Add rudimentary tests for destroy domain and ensure that when cdroms are present that disks are iterated explicitly to avoid attempting to delete iso files as they are also recognized as volumes.


infernix:
Looks good to me :+1:

PR #601, opened 2016-05-18
Allow network gateway attr (fix)

theoden9014:
Template of the network interface in vagrat can be set the gateway attribute, but you can not even want to set because CreateNetworkInterfaces class in "vagrant-libvirt" is filtering the attributes.
Therefore allow network gateway attribute.

https://github.com/mitchellh/vagrant/blob/master/templates/guests/fedora/network_static.erb#L9-L11


infernix:
Looks good to me :+1:

I don't know why it wasn't added in the first place but I tested it and seems fine.

PR #623, opened 2016-06-29
Debian Jessie init sample added

idvoretskyi:
For wider abilities to run multiple OS versions Debian Jessie init sample added.


idvoretskyi:
@pronix any objections?

Issue #627, opened 2016-07-05
domain.boot 'network' misses NICs of type 'UDP' when assigning Boot Order


ericpulvino:
### Steps to reproduce
1. Start the Vagrantfile provided below.
2. Dump the XML output of the libvirt domain to find that there is no boot order specified for interfaces of type UDP.
### Expected behaviour

Interfaces of type UDP should be considered for boot ordering just like any other interface.
### Actual behaviour

UDP interfaces are ignored in line 39 of the boot ordering code --> https://github.com/vagrant-libvirt/vagrant-libvirt/blob/master/lib/vagrant-libvirt/action/set_boot_order.rb#L39
### System configuration

**OS/Distro version:**:
Ubuntu 16.04

**Libvirt version:**
libvirtd (libvirt) 1.3.1

**Output of `vagrant version; vagrant plugin list`:**

```
Installed Version: 1.8.4
Latest Version: 1.8.4

You're running an up-to-date version of Vagrant!
vagrant-cumulus (0.1)
vagrant-libvirt (0.0.33)
  - Version Constraint: 0.0.33
vagrant-mutate (1.1.0)
vagrant-share (1.1.5, system)
```

**Output of `VAGRANT_LOG=debug vagrant ... --provider=libvirt`**
Gist showing the issue. https://gist.github.com/ericpulvino/2c08178dac2410872d8db2884bbec5be#file-vagrant-libvirt-debug-output-L514-L515
You can see in line 514,515 that the boot orders are never set for the UDP interfaces.

**A Vagrantfile to reproduce the issue:**

```
Vagrant.configure("2") do |config|

  config.vm.provider :libvirt do |domain|
    # increase nic adapter count to be greater than 8 for all VMs.
    domain.nic_adapter_count = 55
  end


  ##### DEFINE VM for pxehost #####
  config.vm.define "pxehost" do |device|
    device.vm.hostname = "pxehost" 
    #NO BOX USED FOR PXE DEVICE

    device.vm.provider :libvirt do |v|
      v.storage :file, :size => '100G', :type => 'qcow2', :bus => 'sata', :device => 'sda'
      v.boot 'hd'
      v.boot 'network'
      v.memory = 500
    end

    # NETWORK INTERFACES
      # link for eth0 --> server1:eth0
      device.vm.network "private_network",
            :mac => "443839000006",
            :libvirt__tunnel_type => 'udp',
            :libvirt__tunnel_local_ip => '127.0.0.1',
[...]

PR #628, opened 2016-07-05
Add UDP type NICs to boot order

ericpulvino:
Fixes Issue #627 by adding detection for NICs with type='udp'.

### DIFF
--- a/lib/vagrant-libvirt/action/set_boot_order.rb
+++ b/lib/vagrant-libvirt/action/set_boot_order.rb
@@ -36,7 +36,7 @@ module VagrantPlugins
             # Parse the XML and find each defined drive and network interfacee
             hd = xml.search("/domain/devices/disk[@device='disk']")
             cdrom = xml.search("/domain/devices/disk[@device='cdrom']")
-            network = xml.search("/domain/devices/interface[@type='network']")
+            network = xml.search("/domain/devices/interface[@type='network' or @type='udp']")
 
             # Generate an array per device group and a flattened array from all of those
             devices = {"hd" => hd, "cdrom" => cdrom, "network" => network}


lib/vagrant-libvirt/action/set_boot_order.rb
```diff
@@ -36,21 +36,25 @@ def call(env)
             # Parse the XML and find each defined drive and network interfacee
             hd = xml.search("/domain/devices/disk[@device='disk']")
             cdrom = xml.search("/domain/devices/disk[@device='cdrom']")
-            network = xml.search("/domain/devices/interface[@type='network']")
-
+            network = xml.search("/domain/devices/interface[@type='udp']")
+            network += xml.search("/domain/devices/interface[@type='network']")
```
electrofelix: On first glance would this not result in all udp devices being ordered before other network devices instead of the order defined?

I think you can use an xpath predicate of the following to get them in one go and retain the same order they were defined in:

```

            network = xml.search("/domain/devices/interface[@type='network' and @type='udp']")
```

infernix:
I would agree with the comment from @electrofelix - explicitly ordering udp before any other networks is not what you would expect.

infernix:
Also can you rebase this without the commit removing the debugging code?

ericpulvino:
Ideally there would be a way to have granular (per device/interface) boot ordering support at some point but it looks like a larger modification to the code base to do that since there needs to be a mapping between the configuration specified at the interface level in the Vagrantfile and the XML that is pulled-in from the already created interfaces. It would be easiest if this could be done when the interfaces are created.

I would like to take a stab at implementing this for network interfaces in a way that could later be extended for HDs and CDroms. If I were to implement this, how would you do it?

infernix:
The most practical way is to add a `boot_order_index: X`option to each type (nic/disk/cdrom) that one can boot from. We have `boot_order` already.

When boot_order_index is set but not boot_order, the index can be applied as is. When boot_order is also present it would be either a) use boot_order as primary order and boot_order_index as secondary order, or b) ignore boot_order entirely. Not sure on that one.

ericpulvino:
>When boot_order_index is set but not boot_order, the index can be applied as is.

The question for me is how do you map the interface/HD/cdrom identified in the xml.search output with the one specified in the Vagrantfile to apply the boot_order_index setting (if present). I've been toying with this and can't figure out how to import the boot_order_index parameter I created in the create_network_interfaces.rb but also can't figure out a good way to map them to one another... any thoughts?

>When boot_order is also present it would be either a) use boot_order as primary order and boot_order_index as secondary order, or b) ignore boot_order entirely. Not sure on that one.

I would say make them mutually exclusive because they could conflict. 
Use the order specified by the boot_order logic if boot_order is defined, else, use the custom boot_order_index if present.

infernix:
> can't figure out how to import the boot_order_index parameter I created in the create_network_interfaces.rb but also can't figure out a good way to map them to one another... any thoughts?

Since we cannot set the boot order before all devices are added (see comments in set_boot_order.rb)  I see no other way than to parse `config` in `set_boot_order.rb` for each device type and map them into a new `final_boot_order`. 

Also, maybe use `boot_index`instead of `boot_order_index`.

Issue #631, opened 2016-07-21
 Vagrant libvirt no longer able to create/re-start vms with custom network 


matt8754:
See more details at 

https://bugzilla.redhat.com/show_bug.cgi?id=1310802

It seems to be a regression in #402.

infernix:
Please reproduce with current vagrant and vagrant-libvirt and post logs as described in the issue template: https://github.com/vagrant-libvirt/vagrant-libvirt/blob/master/.github/issue_template.md

matt8754:
Steps to reproduce
1. Define a custom libvirt network "abc"
2. put  server.vm.network "private_network", libvirt__network_name: "abc" in the Vagrantfile
3. vagrant up
Expected behaviour
Vagrant libvirt  can create a vm with this custom network
Tell us what should happen
Actual behaviour
$ sudo vagrant up     
Bringing machine 'origin' up with 'libvirt' provider...
==> origin: Uploading base box image as volume into libvirt storage...
==> origin: Creating image (snapshot of base box volume).
==> origin: Creating domain with the following settings...
==> origin:  -- Name:              fedoradev_origin
==> origin:  -- Domain type:       kvm
==> origin:  -- Cpus:              1
==> origin:  -- Memory:            4096M
==> origin:  -- Management MAC:    
==> origin:  -- Loader:            
==> origin:  -- Base box:          fedora-23
==> origin:  -- Storage pool:      default
==> origin:  -- Image:             /var/lib/libvirt/images/fedoradev_origin.img (41G)
==> origin:  -- Volume Cache:      default
==> origin:  -- Kernel:            
==> origin:  -- Initrd:            
==> origin:  -- Graphics Type:     vnc
==> origin:  -- Graphics Port:     5900
==> origin:  -- Graphics IP:       127.0.0.1
==> origin:  -- Graphics Password: Not defined
==> origin:  -- Video Type:        cirrus
==> origin:  -- Video VRAM:        9216
==> origin:  -- Keymap:            en-us
==> origin:  -- Disks:         vdb(qcow2,10G)
==> origin:  -- Disk(vdb):     /var/lib/libvirt/images/fedoradev_origin-vdb.qcow2
==> origin:  -- INPUT:             type=mouse, bus=ps2
==> origin:  -- Command line : 
==> origin: Creating shared folders metadata...
Network abc is not available. Specify available network
name, or an ip address if you want to create a new network.

OS/Distro version:: fedora-23

Libvirt version:
[...]

PR #632, opened 2016-07-22
fix a regression introduced in 546c98c

matt8754:


matt8754:
This patch is aimed to fix #631 when creating a non-veryisolated network.

pronix:
create test please with described behaviour

matt8754:
I have been very busy since last month and will find out some time to create a test for this, even though I have noticed that the test coverage for this is so poor.

infernix:
@pronix perhaps merge and add test later?

PR #639, opened 2016-08-08
Validator for network MAC configuration options.

artworklv:
Implements #638 feature.
Validates network interface mac address if it is configured.
MAC should be in canonical form 'xx:xx:xx:xx:xx:xx', otherwise vagrant throws configuration error.


infernix:
It should also be enforced lowercase, so please drop A-F and leave only a-f, or add some code that always converts it to lowercase after parsing Vagrantfile.

infernix:
LGTM, @pronix?

petems:
@artworklv Thanks for this! 👍

PR #646, opened 2016-09-07
Add NUMA topology support

stephenfin:
I'm not a ruby developer so apologies if I've done something daft. Suggestions of how I can test this would be appreciated.

### DIFF
--- a/lib/vagrant-libvirt/action/create_domain.rb
+++ b/lib/vagrant-libvirt/action/create_domain.rb
@@ -37,6 +37,7 @@ module VagrantPlugins
           @cpu_mode = config.cpu_mode
           @cpu_model = config.cpu_model
           @cpu_fallback = config.cpu_fallback
+          @numa_nodes = config.numa_nodes
           @loader = config.loader
           @machine_type = config.machine_type
           @machine_arch = config.machine_arch
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -62,6 +62,7 @@ module VagrantPlugins
       attr_accessor :cpu_model
       attr_accessor :cpu_fallback
       attr_accessor :cpu_features
+      attr_accessor :numa_nodes
       attr_accessor :loader
       attr_accessor :boot_order
       attr_accessor :machine_type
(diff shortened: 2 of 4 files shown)


infernix:
Think there's an error in your _generate_numa since Travis build fails; can you try to fix it?

Also, please remove the tab fix commit in non-related code; happy to take cosmetic fixes but shouldn't be in a PR that adds functionality :)

lib/vagrant-libvirt/action/create_domain.rb
```diff
@@ -12,7 +12,7 @@ def initialize(app, _env)
         end
 
         def _disk_name(name, disk)
-          "#{name}-#{disk[:device]}.#{disk[:type]}"	# disk name
+          "#{name}-#{disk[:device]}.#{disk[:type]}"  # disk name
```
pronix: this change is not required.

lib/vagrant-libvirt/config.rb
```diff
@@ -198,15 +200,15 @@ def initialize
       end
 
       def boot(device)
-        @boot_order << device	# append
+        @boot_order << device  # append
```
pronix: this, 209 211 - changes for nothing.

lib/vagrant-libvirt/config.rb
```diff
@@ -457,6 +482,8 @@ def finalize!
         @cpu_model = 'qemu64' if @cpu_model == UNSET_VALUE
         @cpu_fallback = 'allow' if @cpu_fallback == UNSET_VALUE
         @cpu_features = [] if @cpu_features == UNSET_VALUE
+        @numa_nodes = nil if @numa_nodes == UNSET_VALUE
```
pronix: why require redefine varaiable if value not set ?

infernix:
@stephenfin Does this work for any value of mem, even not cleanly divisable in numa_nodes?

stephenfin:
@infernix No, the memory would also have to be cleanly divisible into numa_nodes. I guess we could either assume everyone's a consenting adult and won't pass in something daft, or we could raise an exception if it didn't divide cleanly. If we opt for the latter, I'd appreciate a little help as, again, not a ruby dev :smile:

infernix:
@stephenfin always assume user isn't bright ;)

PR #649, opened 2016-09-07
Add disk property 'shareable'

### LINKED ISSUE #648: Shared disks not configured properly when following the vagrant-libvirt documentation
There is section in the vagrant-libvirt documentation (https://github.com/vagrant-libvirt/vagrant-libvirt#additional-disks) which mentions the use of shared disks:
```
allow_existing - Set to true if you want to allow the VM to use a pre-existing disk.
This is useful for sharing disks between VMs, e.g. in order to simulate shared SAN storage.
Shared disks removed only manually. If not exists - will created. If exists - using existed.
```

In our situation we want to simulate shared SAN storage so we decided to set 'allow_existing' to true.

It turns out that two additional changes are needed to get simulated SAN storage working correctly with vagrant-libvirt:
1. Each disk which needs to be shared also needs to have the libvirt disk property 'shareable' set as mentioned in http://libvirt.org/formatdomain.html#elementsDisks
2. Testing has indicated that shared disks using qcow2 as storage type are not working (changes done in VM1 are not directly visible in VM2). When using the raw storage type shared disks work as expected (changes are directly visible in all VM's). This should be mentioned in the documentation

To fix issue 1 I had to apply the following change against vagrant-libvirt:
```patch
--- vagrant-libvirt/config.rb.orig	2016-09-06 17:06:03.481730741 +0200
+++ vagrant-libvirt/config.rb	2016-09-06 17:07:06.487551438 +0200
@@ -252,6 +252,7 @@
           :bus => options[:bus],
           :cache => options[:cache] || 'default',
           :allow_existing => options[:allow_existing],
+          :shareable => options[:shareable],
         }
 
         @disks << disk	# append
--- vagrant-libvirt/templates/domain.xml.erb.orig	2016-09-06 17:02:17.699174595 +0200
+++ vagrant-libvirt/templates/domain.xml.erb	2016-09-06 17:27:16.395037315 +0200
@@ -61,6 +61,9 @@
       <driver name='qemu' type='<%= d[:type] %>' cache='<%= d[:cache] %>'/>
       <source file='<%= d[:absolute_path] %>'/>
       <target dev='<%= d[:device] %>' bus='<%= d[:bus] %>'/>
+      <% if d[:shareable] %>
+        <shareable/>
+      <% end %>
 <%# this will get auto generated by libvirt
       <address type='pci' domain='0x0000' bus='0x00' slot='???' function='0x0'/>
 -%>
```
This adds an additional property which users can set in the Vagrantfile for each disk.

To properly create shared disks the following should be sufficient to add to the Vagrantfile:
```
libvirt.storage :file, :size => '1G', :path => 'my_shared_disk.img', :allow_existing => true, :bus => 'scsi', :shareable => true, :type => 'raw'
```

If you consider this approach to be correct then I'll create a pull request for it which also updates the documentation

epienbroek:
To simulate shared SAN storage an additional libvirt
disk property needs to be set which disables caching.

Also updated the documentation to document this new
property and added an example on how to simulate
shared SAN storage

Closes #648


infernix:
LGTM; @pronix?

PR #654, opened 2016-09-15
Add support for RNG passthrough

rmohr:
Allow passing through /dev/random from the host to the VM. This speeds
up pretty much everything which needs entropy inside the VM.

### DIFF
--- a/lib/vagrant-libvirt/action/create_domain.rb
+++ b/lib/vagrant-libvirt/action/create_domain.rb
@@ -89,6 +89,9 @@ module VagrantPlugins
           # USB device passthrough
           @usbs = config.usbs
 
+          # RNG device passthrough
+          @rng =config.rng
+
           config = env[:machine].provider_config
           @domain_type = config.driver
 
@@ -217,6 +220,10 @@ module VagrantPlugins
 
           @pcis.each do |pci|
             env[:ui].info(" -- PCI passthrough:   #{pci[:bus]}:#{pci[:slot]}.#{pci[:function]}")
+          end
+
+          if !@rng[:model].nil?
+            env[:ui].info(" -- RNG device model:  #{@rng[:model]}")
           end
 
           @usbs.each do |usb|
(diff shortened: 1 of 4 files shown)


infernix:
A few typos, but LGTM. @pronix ?

lib/vagrant-libvirt/config.rb
```diff
@@ -317,6 +323,18 @@ def channel(options={})
         })
       end
 
+      def random(options={})
+        if !options[:model].nil? && options[:model] != "random"
+          raise 'The only supported rng backend is "random".'
+        end
+
+        if @rng == UNSET_VALUE
```
pronix: it should be first line of method

pronix: and function is not used.

rmohr: > and function is not used.

The function is. When specifying the random device in the Vagrantfile like mentioned in the readme:

```
      domain.random :model => 'random'
  end
```

it is called.

> it should be first line of method

But all other functions have exactly this structure. First input validation, then checking if the target config property is UNSET_VALUE and acting accordingly.

But maybe I miss something here ...

rmohr:
> A few typos, but LGTM. @pronix ?

Fixed the typos where I saw them.

@pronix @infernix thanks for the fast feedback :+1:

lib/vagrant-libvirt/action/create_domain.rb
```diff
@@ -89,6 +89,9 @@ def call(env)
           # USB device passthrough
           @usbs = config.usbs
 
+          # RNG device passthrough
+          @rng =config.rng
```
infernix: Nitpicking, but... `@rng = config.rng`

:)

aspiers:
Is it just me or is this documented wrong?  It's documented as

    libvirt.random :model => 'random'

but AFAICS from the code it should be

    libvirt.rng :model => 'random'

aspiers:
Nope, it was just me - never mind :-)

PR #662, opened 2016-10-05
patch to include virtual sound card support

lcampbel-rh:


lib/vagrant-libvirt/config.rb
```diff
@@ -529,6 +530,7 @@ def finalize!
         @graphics_ip = '127.0.0.1' if @graphics_ip == UNSET_VALUE
         @video_type = 'cirrus' if @video_type == UNSET_VALUE
         @video_vram = 9216 if @video_vram == UNSET_VALUE
+        @sound_type = 'ich6' if @sound_type == UNSET_VALUE
```
infernix: I don't think we should default to adding a sound card. In most cases this will not be used.

pronix: :+1:  we don't need sound by default. will merge after fix

infernix:
@lcampbel-rh a new release is due shortly; if you can push a fix we can merge this beforehand :)

lcampbel-rh:
Working on it. Having some issues with the sound device get passed through
to fog as a nil type. Will have time to look at this over the weekend.

lib/vagrant-libvirt/templates/domain.xml.erb
```diff
@@ -131,7 +131,12 @@
 <% @inputs.each do |input| %>
     <input type='<%= input[:type] %>' bus='<%= input[:bus] %>'/>
 <% end %>
-
+    <% if @sound_type != 'none' %>
```
electrofelix: I'd switch to a simple check of:
```
    <% if !@sound_type.nil? %>
```
And then set the default in finalize! above to `nil` when it's current `UNSET_VALUE`.

That should skip adding the sound block unless sound is configured.

infernix:
@lcampbel-rh any luck with this?

PR #672, opened 2016-10-12
Retry mac to ip search when libvirt throws an error

infernix:
Libvirt will return 'Call to virNetworkGetDHCPLeases failed: internal
error: no lease with matching MAC address' when a mac is not found.
fog-libvirt behaviour changed in 0.2.0 but since we immediately start to
poll for an IP, we need to retry despite the libvirt error.

Workaround fix for #668

### DIFF
--- a/lib/vagrant-libvirt/action/wait_till_up.rb
+++ b/lib/vagrant-libvirt/action/wait_till_up.rb
@@ -41,8 +41,14 @@ module VagrantPlugins
 
               # Wait for domain to obtain an ip address
               domain.wait_for(2) {
-                addresses.each_pair do |type, ip|
-                  env[:ip_address] = ip[0] if ip[0] != nil
+                begin
+                  addresses.each_pair do |type, ip|
+                    env[:ip_address] = ip[0] if ip[0] != nil
+                  end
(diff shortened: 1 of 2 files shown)


infernix:
@pronix this look OK to you? I'd like this to be fixed in fog-libvirt but this should help in the meantime.

lib/vagrant-libvirt/action/wait_till_up.rb
```diff
@@ -41,8 +41,14 @@ def call(env)
...
-                addresses.each_pair do |type, ip|
-                  env[:ip_address] = ip[0] if ip[0] != nil
+                begin
+                  addresses.each_pair do |type, ip|
+                    env[:ip_address] = ip[0] if ip[0] != nil
+                  end
+                # Workaround for fog-libvirt terminating when mac is not found
+                rescue Libvirt::Error
+                  sleep(2)
+                  retry
```
pronix: if i correct understart , retry will ubique with timeout 2 seconf

infernix: I've found a better way already. Will redo entire PR.

PR #675, opened 2016-10-18
[Docs] Add instructions for Arch Linux

petRUShka:
There are some troubles (two at the moment) with installation process of `vagrant-libvirt` on Archlinux. I think it is good idea to documented link to up-to-date workarounds.


infernix:
thanks!

PR #678, opened 2016-10-23
set cpu model only for custom cpu mode

evgeni:
according to https://libvirt.org/formatdomain.html#elementsCPU setting the model is not supported when using "host-model" and with recent libvirt this actually results in errors like this:

```
There was an error talking to Libvirt. The error message is shown
below:

Call to virDomainCreateWithFlags failed: the CPU is incompatible with host CPU: Host CPU does not provide required features: svm
```

### DIFF
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -502,7 +502,8 @@ module VagrantPlugins
         @memory = 512 if @memory == UNSET_VALUE
         @cpus = 1 if @cpus == UNSET_VALUE
         @cpu_mode = 'host-model' if @cpu_mode == UNSET_VALUE
-        @cpu_model = 'qemu64' if @cpu_model == UNSET_VALUE
+        @cpu_model = 'qemu64' if (@cpu_model == UNSET_VALUE and @cpu_mode == 'custom')
+        @cpu_model = '' if (@cpu_mode != 'custom')
         @cpu_fallback = 'allow' if @cpu_fallback == UNSET_VALUE
         @cpu_features = [] if @cpu_features == UNSET_VALUE
         @numa_nodes = @numa_nodes == UNSET_VALUE ? nil : _generate_numa()
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -502,8 +502,11 @@ module VagrantPlugins
         @memory = 512 if @memory == UNSET_VALUE
         @cpus = 1 if @cpus == UNSET_VALUE
         @cpu_mode = 'host-model' if @cpu_mode == UNSET_VALUE
-        @cpu_model = 'qemu64' if (@cpu_model == UNSET_VALUE and @cpu_mode == 'custom')
-        @cpu_model = '' if (@cpu_mode != 'custom')
+        @cpu_model = if (@cpu_model == UNSET_VALUE and @cpu_mode == 'custom')
+            'qemu64'
+          elsif (@cpu_mode != 'custom')
+            ''
+          end
         @cpu_fallback = 'allow' if @cpu_fallback == UNSET_VALUE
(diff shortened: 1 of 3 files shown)


lib/vagrant-libvirt/config.rb
```diff
@@ -502,7 +502,8 @@ def finalize!
         @memory = 512 if @memory == UNSET_VALUE
         @cpus = 1 if @cpus == UNSET_VALUE
         @cpu_mode = 'host-model' if @cpu_mode == UNSET_VALUE
-        @cpu_model = 'qemu64' if @cpu_model == UNSET_VALUE
+        @cpu_model = 'qemu64' if (@cpu_model == UNSET_VALUE and @cpu_mode == 'custom')
+        @cpu_model = '' if (@cpu_mode != 'custom')
```
pronix: twice assigned variable.

lib/vagrant-libvirt/templates/domain.xml.erb
```diff
@@ -7,7 +7,7 @@
 
   <cpu mode='<%= @cpu_mode %>'>
     <% if @cpu_mode != 'host-passthrough' %>
-      <model fallback='<%= @cpu_fallback %>'><%= @cpu_model %></model>
+      <model fallback='<%= @cpu_fallback %>'><% if @cpu_mode == 'custom' %><%= @cpu_model %><% end %></model>
```
pronix: if ```custom``` then ```@cpu_model``` also custom

lib/vagrant-libvirt/config.rb
evgeni: would you prefer:
```ruby
if (@cpu_model == UNSET_VALUE and @cpu_mode == 'custom')
  @cpu_model = 'qemu64'
elsif (@cpu_model == UNSET_VALUE)
  @cpu_model = ''
end
```
instead?

lib/vagrant-libvirt/templates/domain.xml.erb
evgeni: not sure I undertand your comment :(

lib/vagrant-libvirt/config.rb
pronix:
``` @cpu_model = if ...
elsif ....
else raise....
end```

File: lib/vagrant-libvirt/templates/domain.xml.erb (hunk 2 above)
Comment (pronix) [inline lib/vagrant-libvirt/templates/domain.xml.erb]: you are right. ignore it.

File: lib/vagrant-libvirt/config.rb [hunk 3]
```diff
@@ -502,7 +502,11 @@ def finalize!
         @memory = 512 if @memory == UNSET_VALUE
         @cpus = 1 if @cpus == UNSET_VALUE
         @cpu_mode = 'host-model' if @cpu_mode == UNSET_VALUE
-        @cpu_model = 'qemu64' if @cpu_model == UNSET_VALUE
+        @cpu_model = if (@cpu_model == UNSET_VALUE and @cpu_mode == 'custom')
+            'qemu64'
```
pronix: previous default value was qemu64
right now for keep same settings require define cpu_mode to "custom"
i think default settings and behaviour should not be changed.

evgeni: That's the point of this PR ;-)

While it is okay to set "random" values for `cpu_model` when `cpu_mode` is `custom`, but when set to `host-model` or `host-passthrough` it is not really supported (https://libvirt.org/formatdomain.html#elementsCPU) and may even make the VM fail to start (https://www.redhat.com/archives/libvir-list/2016-May/msg01940.html).

Thus the idea was to to keep the `qemu64` model while in custom mode, but reset it to an empty string in any other mode.

pronix:
@evgeni thanks

lelutin:
I'm still experiencing this issue on latest release (which was done before this PR was merged). I'd encourage project maintainers to make a new release in order to get this fix out the door :)

PR #691, opened 2016-11-25
Spec test for domain template to verify rendering

electrofelix:
Add basic spec tests for the domain template to test rendering through
simple wrapping of a config object to provide values for class instance
variables used by the template as it mostly mirrors those required.


infernix:
LGTM; you can merge spec tests yourself all you want as far as I'm concerned.

electrofelix:
Good to know, will do.

PR #694, opened 2016-11-25
Fix storage config blocks

electrofelix:
Ensures that multiple storage config blocks are merged correctly for disk and cdrom by delaying device assignment until the finalize! call.


infernix:
Very nice. LGTM! @pronix?

electrofelix:
"Spec test for domain template to verify rendering" has already been merged since it's just a spec test, but it appears that doesn't get reflected when looking at this, possibly because GH does an explicit merge even if a fast-forward is available. I'll rebase and push up to update PR to remove the extra noise.

PR #695, opened 2016-11-25
[WIP] Add basic support to control serial console settings

electrofelix:
Support limited user defined serial console settings to redirect boot
time messages to a log file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ can help a lot :-)
 - [Input](#input)
 - [PCI device passthrough](#pci-device-passthrough)
 - [USB Redirector Devices](#usb-redirector-devices)
+- [Serial Console Devices](#serial-console-devices)
 - [Random number generator passthrough](#random-number-generator-passthrough)
 - [CPU Features](#cpu-features)
 - [No box and PXE boot](#no-box-and-pxe-boot)
@@ -357,6 +358,9 @@ end
   for use by the [qemu guest
   agent](http://wiki.libvirt.org/page/Qemu_guest_agent) and the Spice/QXL
   graphics type.
+* `serial` - [libvirt serial devices](https://libvirt.org/formatdomain.html#elementsConsole).
+  Configure a serial/console port to communicate with the guest. Can be used
+  to log to file boot time messages sent to ttyS0 console by the guest.
 
 Specific domain settings can be set for each domain separately in multi-VM
 environment. Example below shows a part of Vagrantfile, where specific options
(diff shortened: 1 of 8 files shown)


electrofelix:
I expect this needs more work before it'll be merge ready, but thought I'd pop it up so that even if I don't get it finished someone else might be able to complete it.

Main use case we've had for this locally is for capturing boot time messages from CI jobs running on systems so that people could inspect for kernel messages logged after completion, as typically unlikely to be connected into the build slave to watch consoles directly.

infernix:
Nice! I will test this sometime next week, great work :+1:

ira-lb:
is this abandoned?

mark-i-m:
This would be super useful to have. Are there any plans to complete it?

jwcmd:
I don't normally @ people just for an update but this has been idle for >3 years now and in another comment (https://github.com/vagrant-libvirt/vagrant-libvirt/issues/901) it's hinted that additional work needs to be done but there's no mention of that work here.

Is there anything we can do to help move this PR along @electrofelix @infernix ?

Thanks kindly for your work on this! Loving it! Looking forward to more, and happy to help if I can!

infernix:
I'm unable to spend any time on vagrant-libvirt in the foreseeable future unfortunately. We could use more maintainers :)

jwcmd:
To bring the discussion here from the other ticket,

> Take a look at #695 which contains the base support, but it needs to be updated and synced with the latest codebase. 2 main issues resulted, in that the log file would be owned by root, and you lost access to connect into the serial console so had been meaning to look into a better default that could forward the output to a user process that could two wait serial communication to log to a file for what was outputted while still permitting access for users needing to connect into the console.

> the log file would be owned by root

How problematic is this? I would begrudgingly accept with this myself if it meant being able to get console logs at all, perhaps someone with a better idea of how to solve that could post a PR down the road as well to make them user readable for qemu:///session.

> you lost access to connect into the serial console

To clarify, the situation you had in mind here was also one where the box was created with qemu:///session, correct? And that enabling the console logs would cause the user to not be able to connect to the console of the VM? I would accept that as well if it means being able to get console logging from the VM into a logfile, again hoping that someone with a better approach would post another PR later on if it bothers them enough.

I have Ruby experience and am interested in picking this up 🎁 
Have found the Development section in the readme, 👍

jwcmd:
Closing the loop on this: based on the problems we're unable to get past in https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1032 I can't invest more time in fixing this like I'd hoped, I'll have to try OpenStack instead. Thank you to everyone who's worked on this so far!

electrofelix:
@jwcmd I'm sorry, unfortunately I don't have the same time I used to previously to fix issues, I did respond on the other issue that it looks like you were using a base image that is broken, and using the ones produced by Fedora directly works much better.

For debugging bootstrap issues, I'd generally recommend connecting a vncviewer or using virt-viewer to look at the VM booted as by default the graphics port should be enabled to be automatically assigned.

I'll see what I can do about sorting this PR out to just print a caveat about the root permissions to alert end-users, at the time I think I was hoping to redirect to a process via a tcp connection so that we could launch a tmux instance as a user connected to read from the corresponding port which would allow users to connect directly as well. But probably better as you suggested to just land this and then improve in the future.

PR #700, opened 2016-12-01
catch -h/--help passed to create_box.sh and output usage

evgeni:
instead of passing it to readlink, confisung the user with strange output


infernix:
:+1:

PR #701, opened 2016-12-01
add a way to inject data into the Vagrantfile

evgeni:
for example, I have to set a few settings for my box:

    % cat Vagrantfile.add
        libvirt.nic_model_type = "e1000"
        libvirt.disk_bus = "sata"
      end

      config.vm.guest = :tinycore
    % create_box.sh my.qcow2 my.box Vagrantfile.add

### DIFF
--- a/tools/create_box.sh
+++ b/tools/create_box.sh
@@ -8,7 +8,7 @@ error() {
 }
 
 usage() {
-    echo "Usage: ${0} IMAGE [BOX]"
+    echo "Usage: ${0} IMAGE [BOX] [Vagrantfile.add]"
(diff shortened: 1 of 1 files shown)


infernix:
How about supplying the entire Vagrantfile instead of just a part to add? I mean if this needs to be done, it makes more sense, since there will be cases where the generated Vagrantfile data inside `create_box.sh` is wrong. 

So instead of a `Vagrantfile.add` just take the complete contents of a custom defined `Vagrantfile`?

evgeni:
Sure, yeah, that would work too.

PR #707, opened 2016-12-13
Make mgmt network detachable from no box VMs 

michalskalski:
Pxe nodes are not provisioned, mgmt network is not required for them.
Also take into account bridge interfaces when set boot order, to support
scenario when we would like to use vagrant VM to pxe boot physical nodes
in existing network.

### DIFF
--- a/lib/vagrant-libvirt/util/network_util.rb
+++ b/lib/vagrant-libvirt/util/network_util.rb
@@ -44,8 +44,18 @@ module VagrantPlugins
             management_network_options[:mac] = management_network_mac
           end
 
+          if (env[:machine].config.vm.box &&
+              !env[:machine].provider_config.mgmt_attach)
+            raise Errors::ManagementNetworkRequired
+          end
+
           # add management network to list of networks to check
-          networks = [management_network_options]
+          # unless mgmt_attach set to false
+          networks = if env[:machine].provider_config.mgmt_attach
+                       [management_network_options]
+                     else
+                       []
+                     end
 
           env[:machine].config.vm.networks.each do |type, original_options|
             logger.debug "In config found network type #{type} options #{original_options}"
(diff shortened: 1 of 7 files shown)


lib/vagrant-libvirt/util/network_util.rb
```diff
@@ -45,7 +45,8 @@ def configured_networks(env, logger)
           end
 
           # add management network to list of networks to check
-          networks = [management_network_options]
+          # except pxe boot (no box) VMs which won't be provisioned
+          networks = ([management_network_options] if env[:machine].config.vm.box) || []
```
pronix: why not ? may be i want call later ```vagrant ssh``` and connect to it ?

infernix:
Hm. I'm not sure if this is ideal in all cases. No management network will break `vagrant ssh` (even if it's unlikely being used for pxe booted VMs, it's possible some people are using that).

Same with folks that might have altered their `management_network_mode` and run their own flavour of dhcp.

michalskalski:
I think that with this change pxe nodes behaves like it is described in Readme file:

- No provisioning scripts are ran
- No network configuration is being applied to the VM
- No SSH connection can be made

My point of view was that configuration of the VMs are done by pxe server and it is not expected to use vagrant ssh to connect them, but of course I don't know all use cases. I wanted to eliminate mgmt interface from VM to prevent situation when VM is trying to boot from it.

infernix:
I wrote that code with the assumption that nobody reconfigures the management network, but that doesn't mean nobody does. Right now qemu probes eth0 for dhcp/pxe but simply skips it if it doesn't get any pxe responses, so it still works.

I'm not against disabling the management network, but perhaps better not break default behaviour. Maybe make it a per vm option?

infernix:
@michalskalski thanks! :+1:

PR #730, opened 2017-02-02
Use slot number to report MAC address

jarrpa:
Use a network device's slot number to detect and store the generated MAC address if one hasn't already been provided.

This accommodates scenarios where multiple devices are defined on the same network.

Closes: #645

### DIFF
--- a/lib/vagrant-libvirt/action/create_network_interfaces.rb
+++ b/lib/vagrant-libvirt/action/create_network_interfaces.rb
@@ -122,20 +122,17 @@ module VagrantPlugins
             end
 
             # Re-read the network configuration and grab the MAC address
-            next if @mac
-            xml = Nokogiri::XML(domain.xml_desc)
-            if iface_configuration[:iface_type] == :public_network
+            if iface_configuration[:iface_type] == :public_network and not @mac
+              xml = Nokogiri::XML(domain.xml_desc)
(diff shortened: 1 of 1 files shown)


jarrpa:
....if so, then [this](https://github.com/vagrant-libvirt/vagrant-libvirt/blob/master/lib/vagrant-libvirt/action/create_network_interfaces.rb#L46-L67) seems extremely broken. Is it possible to create a domain with two devices and number one of them as adapter 3? Then you'd have an empty slot between, it would seem, and that should cause an error when trying to iterate through each_with_index. Unless I'm still too new to Ruby.

pronix:
this patch will broke existed behaviour, please update it.

electrofelix:
@jarrpa agreed leaving a hole seems incorrect but when you have multiple networks defined the ability to explicitly control device and network order is supported so need to retain it while fixing here.

Perhaps it's not sufficiently useful that it could be deprecated and removed in the future, but would have to go through a period where users are notified of the plan to remove support.

jarrpa:
@pronix @electrofelix Okay, changed it up a bit, going off the comment made [here](https://github.com/vagrant-libvirt/vagrant-libvirt/issues/645#issuecomment-263236075). Also cleane dup a little bit of logic to make the change easier to read.

PR #741, opened 2017-02-27
Introduce qemu:commandline (for osx quirks)

azat:
Hello everyone,

It will be great to have support for creating configs like this:
```
<qemu:commandline>
    <qemu:arg value='-device'/>
    <qemu:arg value='isa-applesmc,osk=secret-here'/>
    <qemu:arg value='-smbios'/>
    <qemu:arg value='type=2'/>
    <qemu:arg value='-cpu'/>
    <qemu:arg value='Penryn,vendor=GenuineIntel'/>
</qemu:commandline>
```

For now it is not possible to pass `-device` and `-smbios` at least, and I'm not sure about `-cpu` (at least `vendor` for it).

And here is my proposal, though I guess it is ugly (since I don't write in `ruby`)

Thanks,
Azat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1220,6 +1220,20 @@ Vagrant.configure(2) do |config|
 end
 ```
 
+## Custom command line arguments
+You can also specify multiple qemuargs arguments for qemu-system
+
+* `value` - Value
+
+```ruby
+Vagrant.configure("2") do |config|
+  config.vm.provider :libvirt do |libvirt|
+    libvirt.qemuargs :value => "-device"
+    libvirt.qemuargs :value => "intel-iommu"
+  end
+end
+```
+
 ## Box Format
 
 You can view an example box in the
--- a/lib/vagrant-libvirt/action/create_domain.rb
+++ b/lib/vagrant-libvirt/action/create_domain.rb
@@ -88,9 +88,9 @@ module VagrantPlugins
 
           # PCI device passthrough
           @pcis = config.pcis
-          
+
           # Watchdog device
-          @watchdog_dev = config.watchdog_dev          
+          @watchdog_dev = config.watchdog_dev
 
           # USB device passthrough
           @usbs = config.usbs
@@ -101,6 +101,9 @@ module VagrantPlugins
 
           # smartcard device
           @smartcard_dev = config.smartcard_dev
+
+          # qemuargs
+          @qemuargs = config.qemuargs
 
           # RNG device passthrough
           @rng = config.rng
@@ -278,6 +281,14 @@ module VagrantPlugins
 
           if not @smartcard_dev.empty?
             env[:ui].info(" -- smartcard device:  mode=#{@smartcard_dev[:mode]}, type=#{@smartcard_dev[:type]}")
+          end
+
+          unless @qemuargs.empty?
+            env[:ui].info(' -- Command line args: ')
+            @qemuargs.each do |arg|
+              msg = "    -> value=#{arg[:value]}, "
+              env[:ui].info(msg)
+            end
           end
 
           env[:ui].info(" -- Command line : #{@cmd_line}")
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -141,6 +141,9 @@ module VagrantPlugins
       # Attach mgmt network
       attr_accessor :mgmt_attach
 
+      # Additional qemuargs arguments
+      attr_accessor :qemuargs
+
       def initialize
         @uri               = UNSET_VALUE
         @driver            = UNSET_VALUE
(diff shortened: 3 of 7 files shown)


pronix:
hello @azat . thanks and require fix this pr for merge.

README.md
```diff
@@ -1159,6 +1159,20 @@ Vagrant.configure(2) do |config|
...
+
+```ruby
+Vagrant.configure("2") do |config|
+  config.vm.provider :libvirt do |libvirt|
+    libvirt.commandline :value => "-device"
+    libvirt.commandline :value => "isa-applesmc,osk=ourhardworkbythesewordsguardedpleasedontsteal(c)AppleComputerInc"
+  end
+end
+```
+
```
electrofelix: I believe you're missing an addition to the merge method in the `VagrantPlugins::ProviderLibvirt::Confg` class to ensure that additional blocks will merge the `commandline` settings together.

Haven't check it myself, but I'd expect the following to have unexpected behaviour
```
    libvirt.commandline :value => "-device"
    libvirt.commandline :value => "isa-applesmc,osk=ourhardworkbythesewordsguardedpleasedontsteal(c)AppleComputerInc"
  end

  # this block will override the previous block for commandline
    libvirt.commandline :value => "-device"
    libvirt.commandline :value => "dummy-device"
  end
end
```

If you want subsequent blocks to always override previous within the same Vagrantfile, this is fine, if you want blocks to append, not so much.

mxl:
@azat Will you fix the PR? I also need this feature.

README.md
azat: I though that this was enough, am I missing something?

```
      def commandline(options = {})
        @commandline_args = [] if @commandline_args == UNSET_VALUE
        @commandline_args.push(value: options[:value])
      end
```

Besides I checked it (and iff I packaged it right), it works.

azat:
Hi @mxl, great, I've just rebased it.

README.md
azat: Nope, that is not true, /me digging

mxl:
@azat Is it possible to provide all command line arguments in single `libvirt.commandline :value => "multiple args here"`? Also may be `libvirt.qemuargs` is better name because it's not libvirt command line args but QEMU's, that also will be more consistent with xml element name `<qemu:arg>` and will look familiar for those who used Packer to build Vagrant box (https://www.packer.io/docs/builders/qemu.html#qemuargs).

mxl:
@azat Any progress with this? How can I help you?

pronix:
@mxl форкаете, дописываете и мержим за 5 минут

azat:
@mxl Yeah, I don't have enough time right now, so if you want to fix it, that will be great!

azat:
Just rebased and `pushed --force` (I hope that hunk in `merge` is enough for merging).

mxl:
@electrofelix @pronix Could you review, please?

mxl:
Guys, is there something that should be fixed here?

electrofelix:
Haven't had a chance to test it out, the suggestion about being able to set multiple args via an array in one go appeals to me but it doesn't look like the code does that. If it can handle doing so as a subsequent change without changing the current proposed behavior that would work for me.

azat:
@electrofelix you mean smth like this:
```ruby
libvirt.qemuargs = ['-device', 'dummy', ...]
```

electrofelix:
Yep, that would make it easier instead of needing to add a config entry per value. If it can be extended to support that without breaking the proposed behavior that would work for me, leave it up to you whether to roll into this PR or separate.

azat:
If you are talking about just
```ruby
libvirt.qemuargs = ['-device', 'dummy', ...]
```
Then I guess I can do this (like for `libvirt.features`).

But if you are talking about supporting both, then let's make this a separate PR (and I'm not sure that I will have time for this, maybe @mxl has? that will be great!)

mxl:
@azat I can do this.

mxl:
Guys, could you merge it? I will open separate PR for next changes.

README.md
```diff
@@ -1220,6 +1220,20 @@ Vagrant.configure(2) do |config|
...
+## Custom command line arguments
+You can also specify multiple qemuargs arguments for qemu-system
+
+* `value` - Value
+
+```ruby
+Vagrant.configure("2") do |config|
+  config.vm.provider :libvirt do |libvirt|
+    libvirt.qemuargs :value => "-device"
+    libvirt.qemuargs :value => "isa-applesmc,osk=ourhardworkbythesewordsguardedpleasedontsteal(c)AppleComputerInc"
```
infernix: Please do not put this example string in the documentation; I'm sure you can find other examples to put there.

infernix:
LGTM after documentation change

README.md
azat: Indeed! Just push forced.

electrofelix:
OK, let's improve it with a subsequent PR.

PR #742, opened 2017-03-01
Always call provisioners logic for booted machines

electrofelix:
Allows `vagrant up` command to handle whether or not to run the
provisioners should the machine already be up. This allows re-running
certain provisioners in parallel using:

  vagrant up --provision --provision-with puppet_server

Which in turn allows for certain scripts that are bootstrap related
only to be skipped.

In general however it is better to let the middleware from vagrant
decided whether provisioners should run or not, as enhancements to that
functionality will then be automatically picked up.

With this change if a user runs `vagrant up` on already running
machines, without the `--provision` option they will be greeted with a
message from vagrant stating "Machine already provisioned. Run ...",
basically letting them know what to do next.

Without this change you simply see a message about "Bringing machine
'<name>' up with 'libvirt' provider..." and then nothing further.

Change-Id: I73ca4262abaa6431a838e4cf6468c06223643dea

### DIFF
--- a/lib/vagrant-libvirt/action.rb
+++ b/lib/vagrant-libvirt/action.rb
@@ -67,8 +67,11 @@ module VagrantPlugins
         Vagrant::Action::Builder.new.tap do |b|
           b.use ConfigValidate
           b.use Call, IsRunning do |env, b2|
-            # If the VM is running, then our work here is done, exit
-            next if env[:result]
+            # If the VM is running, run the necessary provisioners
+            if env[:result]
+              b2.use action_provision
+              next
+            end
 
             b2.use Call, IsSuspended do |env2, b3|
               # if vm is suspended resume it then exit


electrofelix:
Unfortunately I haven't been able to find a sensible way to add unit tests for actions to make it easier to check that for up, halt, resume, etc that the correct set of actions take place so relied on some hand testing that 'vagrant up' repeatedly would do the right thing along with 'vagrant up --provision' and 'vagrant provision'.

pronix:
hello @electrofelix 
i doubt. if vm is running and require provisioning, call ```vagrant provision```.

electrofelix:
@pronix that is limiting in that it can't do the provision in parallel, and also the current behaviour is not the expected behaviour.

Vagrant supports re-calling provision through the `up` command using `up --provision` and thus it is a reasonable expectation for any provider to support doing this. In fact it was a surprise to find it missing from vagrant-libvirt as it exists with other providers such as docker and virtualbox.

Since this is the expected behaviour from vagrant plugin providers, it seems reasonable to support it, particularly as it's not a particular invasive change.

The only impact for those re-running `vagrant up` without the option `--provision` is that they will also finally get the message about the machine already having been brought up and that vagrant is skipping the provisioning. The current behaviour of simply printing "Bringing machine '<name>' up with 'libvirt' provider..." and then nothing more can look like an unexpected bug.

electrofelix:
@pronix was my response helpful?

Seems like a good idea to follow the expected behaviour as it will allow for the standard messages from vagrant to be displayed to users when using the `up` command, as well as supporting the behaviour of the vagrant up command as described at https://www.vagrantup.com/docs/cli/up.html#_no_provision

pronix:
hello @electrofelix 
thanks. Convinced.

Issue #763, opened 2017-04-17
connect_via_ssh does not work if I manually set the uri


dustymabe:
This may not be a bug, but it is definitely not obvious from the documentation and should be clarified if so. 

When I manually set the URI to connect over ssh to a remote libvirt host then `connect_via_ssh = true` does not work and bringing the machine up fails because vagrant can't ssh in. 

### Steps to reproduce
1. set uri to something like: `uri=qemu+ssh://dustymabe@192.168.X.XXX/system`
2. set `connect_via_ssh = true`
3. observe failure

### System configuration
**OS/Distro version:**
Fedora 25 - vagrant-libvirt-0.0.35-3.fc25.noarch

**Libvirt version:**
libvirt-daemon-2.2.0-2.fc25.x86_64

**Output of `vagrant version; vagrant plugin list`:**
$ vagrant version
Installed Version: 1.8.5
Latest Version: 1.9.3

electrofelix:
Looks like the problems is that while the code can build `@uri` from the individual settings, it doesn't support de-constructing a provided `@uri` into the component parts.

Can you confirm that setting the individual pieces used to construct the `@uri` work? Just like to confirm that is the problem.

dustymabe:
> Looks like the problems is that while the code can build @uri from the individual settings, it doesn't support de-constructing a provided @uri into the component parts.

yes, that is what I expected is the issue. 

> 
> Can you confirm that setting the individual pieces used to construct the @uri work? Just like to confirm that is the problem.

yes, i was able to workaround the problem the other day by specifying the different parts of the uri individually to the other options.

itisnotdone:
For me, using 'uri' setting works fine. I think you should probably check debug log using --debug option.
What you should see on the debug log will look like this.
"INFO driver: Connecting to Libvirt (qemu+ssh://USER@IP_OR_DOMAIN/system) ..."

infernix:
Please provide the required information asked when creating an isssue.

dustymabe:
> Please provide the required information asked when creating an isssue.

I guess I was thinking from the details I gave that it would be pretty easy to reproduce. I'll try to get the rest of the information for you, but if there is something specific you want I might be able to get that to you faster.

infernix:
With infinite free time everything can be reproduced ;) so please provide a debug log, thanks.

electrofelix:
Should be fixed by #1225

PR #764, opened 2017-04-19
allow setting the URI from the environment

evgeni:
this allows running vagrant-libvirt against a remote libvirt without modifying the Vagrantfile.

example:
  VAGRANT_LIBVIRT_URI="qemu+ssh://user@192.0.2.1/system" vagrant up

this basically behaves as libvirt's own LIBVIRT_DEFAULT_URI

the patch also automates setting `connect_via_ssh` if the URI contains `+ssh://`

Fixes: #763

### DIFF
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -538,6 +538,15 @@ module VagrantPlugins
       end
 
       def finalize!
+        @uri = ENV['VAGRANT_LIBVIRT_URI'] if (@uri == UNSET_VALUE && ENV['VAGRANT_LIBVIRT_URI'])
+        if (@uri && @uri != UNSET_VALUE)
+          uri = URI(@uri)
+          @driver = u.scheme.sub('+ssh','') if @driver == UNSET_VALUE
+          @host = uri.host if @host == UNSET_VALUE
+          @connect_via_ssh = uri.scheme.include?("+ssh") if @connect_via_ssh == UNSET_VALUE
+          @username = uri.user if @username == UNSET_VALUE
+          @password = uri.password if @password == UNSET_VALUE
+        end
         @driver = 'kvm' if @driver == UNSET_VALUE
         @host = nil if @host == UNSET_VALUE
         @connect_via_ssh = false if @connect_via_ssh == UNSET_VALUE


lib/vagrant-libvirt/config.rb
```diff
@@ -538,6 +538,15 @@ def _generate_uri
...
       def finalize!
+        @uri = ENV['VAGRANT_LIBVIRT_URI'] if (@uri == UNSET_VALUE && ENV['VAGRANT_LIBVIRT_URI'])
+        if (@uri && @uri != UNSET_VALUE)
+          uri = URI(@uri)
+          @driver = u.scheme.sub('+ssh','') if @driver == UNSET_VALUE
+          @host = uri.host if @host == UNSET_VALUE
+          @connect_via_ssh = uri.scheme.include?("+ssh") if @connect_via_ssh == UNSET_VALUE
+          @username = uri.user if @username == UNSET_VALUE
+          @password = uri.password if @password == UNSET_VALUE
+        end
```
electrofelix: This will prevent the generation of the `@uri` below, see line 566 any time VAGRANT_LIBVIRT_URI is defined, even if a Vagrantfile defined all the other pieces of information for `@driver`, `@host`, `@connect_via_ssh`, `@username` and `@password`.

At the same time it will fail to make sure that the individual values have been set based on the provided `@uri`. This seems like a potential for unexpected behaviour.

I would suggest that if the `ENV['VAGRANT_LIBVIRT_URI']` should provide the `@uri`, then all the other settings should be constructed from and replace any provided values. Or the `ENV['VAGRANT_LIBVIRT_URI']` value should be ignored if the other values have been specifically provided.

Also critically important, if using the environment variable and ignoring what has been set in the Vagrantfile, it seems pretty important to report this to user.

Putting this all in a separate method will also make it easier to add some tests to just check the variables relating to `@uri`, which I think a change like this will need.

lib/vagrant-libvirt/config.rb
```diff
@@ -538,6 +538,15 @@ def _generate_uri
       end
 
       def finalize!
+        @uri = ENV['VAGRANT_LIBVIRT_URI'] if (@uri == UNSET_VALUE && ENV['VAGRANT_LIBVIRT_URI'])
+        if (@uri && @uri != UNSET_VALUE)
+          uri = URI(@uri)
+          @driver = u.scheme.sub('+ssh','') if @driver == UNSET_VALUE
```
electrofelix: Btw, this is probably wrong, as driver controls what type of virtualization should be used, qemu, qemu accelerated or kvm, but that is not inferable from the uri.

See https://libvirt.org/uri.html#URI_qemu for more info, and the `_generate_uri` method above in that usage of either kvm or qemu as the driver will result in a `qemu` scheme for the connection.

Course there is also a typo here in that `u.scheme` should be `uri.scheme`.

pronix:
you can use env variables in Vagrantfile.

lib/vagrant-libvirt/config.rb
evgeni: Ha, thanks for spotting the typo!

So would you suggest not to override the driver from the URI at all?

electrofelix: Yes, I can't see that there is any easy connection back from the uri to which driver should be used, so I'd leave it to the config in the Vagrant file to set that. Maybe the documentation should state that use of any other libvirt backend with a full `@uri` being provided will still need to set `@driver`, otherwise it will default to 'kvm'.

As @pronix mentions you can also reference environment variables within a Vagrantfile, so if someone wanted to provide support for this in a Vagrantfile they can do so. A little dubious about needing to support use of `ENV['VAGRANT_LIBVIRT_URI']` directly.

However what is still definitely required is to ensure that if a `@uri` is provided that the other components are extracted correctly and set the necessary config options that are used elsewhere in the code base, or that the code using those is switched to extract them from the `@uri` instead. So there is still something to be improved in this area.

dustymabe:
> you can use env variables in Vagrantfile.

yeah, i think this plugin should either honor `LIBVIRT_DEFAULT_URI` orr just leave the behavior as is and set the `uri = $YOUR_CHOICE_OF_ENV_VAR`. 

what is the use case for having a separate environment variable for this?

evgeni:
The var name was chosen "randomly", I am happy to just use `LIBVIRT_DEFAULT_URI`.

evgeni:
Why did you close? Just because I did not have time to update yet?

pronix:
you can use env variables in Vagrantfile.

evgeni:
but my idea was to override the URI without having to touch the vagrant file (as it lives in a shared repo)

PR #766, opened 2017-04-21
Support for --include and --vagrantfile in package action

reidrac:
This provides the same behaviour seen in other providers (e.g. virtualbox).

Files are placed into a "_include" subdirectory and the user provided Vagrantfile (if any) will be loaded from the main Vagranfile in the box.

Also fixed a typo in the `assemble_box` method.

### DIFF
--- a/lib/vagrant-libvirt/action/package_domain.rb
+++ b/lib/vagrant-libvirt/action/package_domain.rb
@@ -39,12 +39,29 @@ module VagrantPlugins
           # remove hw association with interface
           # working for centos with lvs default disks
           `virt-sysprep --no-logfile --operations defaults,-ssh-userdir -a #{@tmp_img}`
+          # add any user provided file
+          extra = ''
+          @tmp_include = @tmp_dir + '/_include'
+          if env['package.include']
+            extra = './_include'
+            Dir.mkdir(@tmp_include)
+            env['package.include'].each do |f|
+              env[:ui].info("Including user file: #{f}")
+              FileUtils.cp(f, @tmp_include)
+            end
+          end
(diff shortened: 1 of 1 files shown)


pronix:
i need docs for understand what is it.

reidrac:
The vagrant package command supports `--include x,y,x` and `--vagrantfile file`; but currently vagrant-libvirt isn't doing anything with these options.

This patch implements the functionality after looking at how the virtualbox provider works.

1. Copy any included files (for whatever reason the user wants them; for example I include a YAML file with extra information that is read from my Vagrantfile).
2. Copy user's Vagranfile.

To avoid filename collisions files are copied into `./_include` directory. Because your Vagrantfile already sets some values for KVM, I follow the same strategy the virtualbox provider uses to load user's Vagranfile at the end of vagrant-libvirt's Vagrantfile

I hope this helps!

References:

- Vagrant docs: https://www.vagrantup.com/docs/cli/package.html

reidrac:
One thing that we can improve compared to the virtualbox support is that we could add documentation; I can add something to the README.md file if you want (in this or in a different pull request).

I didn't do it in the PR because I just implemented what is already described in Vagrant docs.

pronix:
thanks. it is new for me.

reidrac:
Thanks for vagrant-libvirt!

reidrac:
Quick question; what are your plans to release a new version? I'd love this patch to be available via `vagrant plugin install`.

Thanks!

pronix:
i can make it today.

reidrac:
That would be amazing! Thanks again.

Issue #739, opened 2017-02-25
-append only allowed with -kernel option


strarsis:
For using Ubuntu (mutated ubuntu/xenial64) with Vagrant (and vagrant-libvirt), 
"old" network interface names must be used, this can be achieved by 
setting cmd_line to "net.ifnames=0 biosdevname=0" as described in this [issue](https://github.com/vagrant-libvirt/vagrant-libvirt/issues/609):
````
[...]
Vagrant.configure("2") do |config|
[...]
  config.vm.provider :libvirt do |libvirt|
    libvirt.cmd_line = 'net.ifnames=0 biosdevname=0'
  end
[...]
end
````

However, an error occurs on startup:
````
Call to virDomainCreateWithFlags failed: internal error: process exited while connecting to monitor: [...] qemu-system-x86_64: -append only allowed with -kernel option
````

infernix:
When booting box images, you boot a full BIOS with bootloader, and therefore `-kernel` flags are not usable.  You either need to use the `kernel` and `initrd` options in your vagrantfile (bad idea) or use a [vagrant-libvirt xenial box image](https://atlas.hashicorp.com/boxes/search?utf8=%E2%9C%93&sort=&provider=libvirt&q=xenial) ([see documentation](https://github.com/vagrant-libvirt/vagrant-libvirt#add-box))

Helenbj:
@infernix hello, I face the same problem 

> Call to virDomainCreateWithFlags failed: internal error: process exited while connecting to monitor: 2018-09-12T11:32:25.240178Z qemu-system-x86_64: -append only allowed with -kernel option

can you tell me what you mean by saying 

> use the kernel and initrd options in your vagrantfile (bad idea)

PR #775, opened 2017-05-08
Explain `cmd_line` little further

norpol:
Other people (#739) are confused by the feature too. This pull request makes it hopefully a little clearer that `cmd_line` works only in combination with `initrd` and `kernel`.


norpol:
Thank you!

pronix:
@norpol thank you

PR #777, opened 2017-05-11
Making adapter exhaustion error message more helpful.

ericpulvino:
Consistently forget what this error message means. Trying to make it more useful to others who hit this. I'm using between 130-150 interfaces to simulate network switches and so I hit this error more often.

### DIFF
--- a/locales/en.yml
+++ b/locales/en.yml
@@ -76,8 +76,7 @@ en:
       no_domain_volume: |-
         Volume for domain is missing. Try to run 'vagrant up' again.
       interface_slot_not_available: |-
-        Interface adapter number is already in use. Please specify other adapter
-        number.
+        Available interface adapters have been exhausted. Please increase the nic_adapter_count.
       rsync_error: |-
         There was an error when attempting to rsync a share folder.
         Please inspect the error message below for more info.


locales/en.yml
```diff
@@ -76,8 +76,7 @@ en:
       no_domain_volume: |-
         Volume for domain is missing. Try to run 'vagrant up' again.
       interface_slot_not_available: |-
-        Interface adapter number is already in use. Please specify other adapter
-        number.
+        Available interface adapters have been exhausted. Please increase the nic_adapter_count.
```
electrofelix: I thought it was possible to specify specific slot numbers for the nic adapters, hence the error about the slot in use, looking at the code https://github.com/vagrant-libvirt/vagrant-libvirt/blob/43bd92670df7b6769611f5771dc1264948302eef/lib/vagrant-libvirt/action/create_network_interfaces.rb#L56 it seems to me that it's the line I've referenced that should have had a different error raised with the new message you've defined above.

The previous error message would have been exactly what should be seen when triggered by the code at https://github.com/vagrant-libvirt/vagrant-libvirt/blob/43bd92670df7b6769611f5771dc1264948302eef/lib/vagrant-libvirt/action/create_network_interfaces.rb#L47-L49

Seems like these are actually two different error cases and it needed two different error messages instead of changing the existing one.

ericpulvino: I agree, didn't dig that deep initially, thank you for the commentary. I've submitted #779 to correct this.

PR #779, opened 2017-05-14
Reverts PR#777 and adds a new error message for a different case

ericpulvino:
This PR captures commentary from @electrofelix in #777  about creating a new error message that accurately represents the exhausted slot state instead of re-using another error message inappropriately.


electrofelix:
Tested this locally with a fix for the failing tests (not due to this change), looks good to me.

electrofelix:
see #783 for a fix for the failing tests.

PR #781, opened 2017-05-16
Remove NFS exports on domain destroy

omahn:
When using vagrant with the default virtualbox provider, NFS exports are removed as guests are destroyed. With the libvirt provider, NFS exports are left in `/etc/exports` following a destroy. They are only removed if a `vagrant up` operation subsequently takes place using the libvirt provider.

This leaves the host machine exporting the local filesystem to the IP(s) used previously by the libvirt domains. I think this behaviour is undesirable and potentially a security risk as those IP(s) may now be used for other purposes.

This PR duplicates the existing code used during NFS pruning on a `vagrant up` and places it in the `DestroyDomain` class. NFS exports will therefore be removed as domains are destroyed. I suggest we also leave the existing NFS pruning code in-place to catch any situations where vagrant/libvirt fails during the destroy steps.

### DIFF
--- a/lib/vagrant-libvirt/action.rb
+++ b/lib/vagrant-libvirt/action.rb
@@ -187,7 +187,7 @@ module VagrantPlugins
             b2.use Call, DestroyConfirm do |env2, b3|
               if env2[:result]
                 b3.use ClearForwardedPorts
-                # b3.use PruneNFSExports
+                b3.use PruneNFSExports
                 b3.use DestroyDomain
                 b3.use DestroyNetworks
                 b3.use ProvisionerCleanup
--- a/lib/vagrant-libvirt/action/prepare_nfs_settings.rb
+++ b/lib/vagrant-libvirt/action/prepare_nfs_settings.rb
@@ -6,7 +6,7 @@ module VagrantPlugins
   module ProviderLibvirt
     module Action
       class PrepareNFSSettings
-        include Vagrant::Action::Builtin::MixinSyncedFolders
+        include VagrantPlugins::ProviderLibvirt::Util::Nfs
 
         def initialize(app, _env)
           @app = app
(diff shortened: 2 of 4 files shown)


lib/vagrant-libvirt/action/destroy_domain.rb
```diff
@@ -75,6 +75,14 @@ def call(env)
...
           end
 
+          # Remove NFS exports that are no longer valid following destroy
+          uuid = env[:machine].id
+          uuids = env[:machine].provider.driver.connection.servers.all.map(&:id)
+          uuids.delete(uuid)
+          env[:host].capability(
+            :nfs_prune, env[:machine].ui, uuids
+          )
+
```
electrofelix: This seems like we should simply be adding back in the action PruneNFSExports to the self.action_destroy method in lib/vagrant-libvirt/action.rb.

omahn:
I didn't spot the commented out entry in `action.rb` - I've tested it locally and it appears to work fine so I've removed my change from `destroy_domain.rb` and removed the comment for the pruning code in `action.rb`. Just looks like an oversight to me.

electrofelix:
Tested locally with a patch for the failing tests (see #783), and everything passes.

I've managed to trace the source PR responsible for this change #170 , and the reason appears to be an issue during teardown about pruning NFS exports requiring root privileges.
https://github.com/vagrant-libvirt/vagrant-libvirt/pull/170#issuecomment-42093945

That should be fine if the user decided to enable NFS, but I think the reason this was disabled was because it was getting called even when using 9p, instead of only being limited to whether the machine had NFS syncing in the first place.

Looking at the code in Vagrant for the prune_nfs capability, it did seem like it should only remove entries no longer associated, https://github.com/mitchellh/vagrant/blob/master/plugins/hosts/linux/cap/nfs.rb#L68-L74, but there may have been some subtle issues there, when running multiple vagrant envs on the same system, as well as systems that are locked down (no others read on /etc/exports) will trigger requiring a sudo password to determine whether to remove entries or not.

Perhaps we should add a check to see if the machine has been configured to use NFS before allowing this to be called? Maybe just copy the test from the lib/vagrant-libvirt/action/prepare_nfs_settings.rb for the method `using_nfs?`?

Other than a check on whether to skip this action altogether I think this looks fine.

electrofelix:
I need to test this fully, but I think it should avoid asking for sudo on destroy when using only 9p or rsync for synced folders.

electrofelix:
closing/opening to trigger tests

PR #787, opened 2017-06-06
Do not destroy domain when starting existing domain

mkhon:
WaitTillUp is re-used in action_start but failures (and interruptions)
during action_start are not expected to cause domain destruction:
do halt instead of destroy in WaitTillUp if called from action_start

### DIFF
--- a/lib/vagrant-libvirt/action.rb
+++ b/lib/vagrant-libvirt/action.rb
@@ -54,6 +54,7 @@ module VagrantPlugins
                 # b2.use SyncFolders
               end
             else
+              env[:halt_on_error] = true
               b2.use action_start
             end
           end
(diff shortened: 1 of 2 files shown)


lib/vagrant-libvirt/action.rb
```diff
@@ -54,7 +54,9 @@ def self.action_up
                 # b2.use SyncFolders
               end
             else
-              b2.use action_start
+              start_env = env.dup
+              start_env[:halt_on_error] = true
+              env[:action_runner].run(Action.action_start, start_env)
```
electrofelix: I don't think you need to duplicate the env and use the action runner here, could be wrong though. Believe that is only needed for the halt/destroy on error actions because the env needs to be isolated for the destroy action to remove the interrupted key so it doesn't cause the middleware to abort the destroy/halt action.

Don't need the same for the start action, just set the env key and call as before.

mkhon:
@electrofelix please review an updated pull-request

PR #800, opened 2017-08-02
Feature/numa nodes

homeski:
This refactors and fixes the numa_nodes domain option. 

It also allows for more flexibility when defining NUMA nodes, as asymmetrical topologies are now supported. 

Note the removal of lines 127-133 in `start_domain.xml`. These lines worked to remove all nested XML  objects within `<cpu></cpu>` if `cpu_mode` was set to `host-passthrough`. This doesn't make sense, as we may still want to define NUMA nodes here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,7 +300,7 @@ end
 * `cpu_fallback` - Whether to allow libvirt to fall back to a CPU model close
   to the specified model if features in the guest CPU are not supported on the
   host. Defaults to 'allow' if not set. Allowed values: `allow`, `forbid`.
-* `numa_nodes` - Specify an array of NUMA nodes for the guest. The syntax is similar to what would be set in the domain XML. `memory` must be in MB. Symmetrical and asymmetrical topologies are supported but ake sure your total count of defined CPUs adds up to v.cpus. 
+* `numa_nodes` - Specify an array of NUMA nodes for the guest. The syntax is similar to what would be set in the domain XML. `memory` must be in MB. Symmetrical and asymmetrical topologies are supported but make sure your total count of defined CPUs adds up to v.cpus. 
 
   The sum of all the memory defined here will act as your total memory for your guest VM. **This sum will override what is set in `v.memory`**
   ```
--- a/README.md
+++ b/README.md
@@ -300,7 +300,7 @@ end
 * `cpu_fallback` - Whether to allow libvirt to fall back to a CPU model close
   to the specified model if features in the guest CPU are not supported on the
   host. Defaults to 'allow' if not set. Allowed values: `allow`, `forbid`.
-* `numa_nodes` - Specify an array of NUMA nodes for the guest. The syntax is similar to what would be set in the domain XML. `memory` must be in MB. Symmetrical and asymmetrical topologies are supported but make sure your total count of defined CPUs adds up to v.cpus. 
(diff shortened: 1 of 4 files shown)


lib/vagrant-libvirt/action/start_domain.rb
```diff
@@ -124,13 +128,6 @@ def call(env)
                     cpu.delete_element(svm_feature)
                   end
                 end
-              else
-                unless cpu.elements.to_a.empty?
```
infernix: Should this not be conditional on numa, e.g. do not skip this block `if  config.numa_nodes == nil`?

lib/vagrant-libvirt/config.rb
```diff
@@ -279,26 +279,12 @@ def _get_cdrom_dev(cdroms)
       end
 
       def _generate_numa
-        raise 'NUMA nodes must be a factor of CPUs' if @cpus % @numa_nodes != 0
+        # todo: compare count of NUMA defined CPUs and throw warning
+        # if config.cpu != NUMA.cpu
```
infernix: Either add this check now, or make config.cpu the total of NUMA.cpu (and specify in the docs that this is the case).

homeski:
Updated this pull request to reflect changes from comments.

README.md
```diff
@@ -300,7 +300,16 @@ end
...
   to the specified model if features in the guest CPU are not supported on the
   host. Defaults to 'allow' if not set. Allowed values: `allow`, `forbid`.
-* `numa_nodes` - Number of NUMA nodes on guest. Must be a factor of `cpu`.
+* `numa_nodes` - Specify an array of NUMA nodes for the guest. The syntax is similar to what would be set in the domain XML. `memory` must be in MB. Symmetrical and asymmetrical topologies are supported but make sure your total count of defined CPUs adds up to `v.cpus`. 
+
+  The sum of all the memory defined here will act as your total memory for your guest VM. **This sum will override what is set in `v.memory`**
+  ```
+  v.cpus = 4
+  v.numa_nodes = [
+    {:id => 0, :cpus => "0-1", :memory => "1024"},
```
electrofelix: is ':id' always incrementing? or would it make sense for someone to leave a hole?

Just wondering if it makes sense to calculate it automatically based on the index of each hash within the given array.

Alternatively requiring numbers from users may make merging across blocks easier.

lib/vagrant-libvirt/config.rb
```diff
@@ -279,26 +279,26 @@ def _get_cdrom_dev(cdroms)
       end
```
electrofelix: This file should probably also contain a block added to the 'merge' method to handle numa_nodes being defined in different files & blocks.

README.md
homeski: Looks like id needs to be contiguous range. Tried leaving a hole and libvirt responded with:

Error while creating domain: Error saving the server: Call to virDomainDefineXML failed: XML error: Exactly one 'cell' element per guest NUMA cell allowed, non-contiguous ranges or ranges not starting from 0 are not allowed

lib/vagrant-libvirt/config.rb
homeski: @electrofelix can you help me with understanding the merge method? Reach me on gitter if that is easier.

homeski: Merge behavior isn't wanted for numa_nodes. Override should happen if numa_nodes is defined in different files and blocks.

PR #819, opened 2017-10-14
[Documentation] Add links for understanding 9p

petRUShka:
There is lack of documentation on 9p so it could be useful to extend that part of docs with some links.


infernix:
Thanks!

PR #820, opened 2017-10-20
Add mtu option for private networks and interfaces

Ntr0:
In order to create virtual networks with different MTU sizes
than the libvirt default of 1500, this patch adds a
libvirt__mtu option. This will add an mtu option to the libvirt
network definition and to the domain interface definition as well.

However, the MTU only comes into effect, if the interface on the vm was configured correctly.

### DIFF
--- a/lib/vagrant-libvirt/action/create_network_interfaces.rb
+++ b/lib/vagrant-libvirt/action/create_network_interfaces.rb
@@ -76,6 +76,7 @@ module VagrantPlugins
             @driver_name = iface_configuration.fetch(:driver_name, false)
             @driver_queues = iface_configuration.fetch(:driver_queues, false)
             @device_name = iface_configuration.fetch(:iface_name, false)
+            @mtu = iface_configuration.fetch(:mtu, nil)
             template_name = 'interface'
             # Configuration for public interfaces which use the macvtap driver
             if iface_configuration[:iface_type] == :public_network
@@ -143,6 +144,7 @@ module VagrantPlugins
                                     @device_name,
                                     @iface_number,
                                     @model_type,
+                                    @mtu,
                                     driver_options,
                                     @udp_tunnel)
                     else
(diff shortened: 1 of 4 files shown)


lib/vagrant-libvirt/action/create_network_interfaces.rb
```diff
@@ -245,6 +247,7 @@ def interface_xml(type, source_options, mac, device_name,
               xml.target(dev: target_dev_name(device_name, type, iface_number))
               xml.alias(name: "net#{iface_number}")
               xml.model(type: model_type.to_s)
+              xml.mtu(size: mtu) if mtu
```
electrofelix: Suggest using `mtu.to_i` to make sure this is an integer at this point, or validate that it is either `nil` or an integer in the configuration in the config object.

Ntr0: I decided to use the Integer Constructor, since to_i would cast any string to 0 if its not a valid integer. I prefer an exception, if the input is not a valid integer instead. If that is not what you had in mind, I would write a validator function instead (e.g. new module util/validators & function validate_integer) to raise a custom validation error.

Ntr0:
@electrofelix could you please reexamine?

lib/vagrant-libvirt/action/create_networks.rb
```diff
@@ -286,6 +286,8 @@ def create_private_network(env)
           @network_bridge_name = @interface_network[:bridge_name]
           @network_address = @interface_network[:ip_address]
           @network_netmask = @interface_network[:netmask]
+          @network_mtu = Integer(@options[:mtu]) unless @options[:mtu].nil?
```
pronix: simplify ``` unless @options[:mtu].nil?```

lib/vagrant-libvirt/templates/private_network.xml.erb
```diff
@@ -6,6 +6,10 @@
     <domain name="<%= @network_domain_name %>" localOnly="yes" />
   <% end %>
 
+  <% if !@network_mtu.nil? %>
```
pronix: simplify

PR #824, opened 2017-11-03
add Vagrant Guest Halt capability to vagrant-libvirt

skamithi:
add guest halt capability to vagrant-libvirt . Reference: [Builtin Guest Halt capability](https://github.com/hashicorp/vagrant/blob/master/lib/vagrant/action/builtin/graceful_halt.rb#L50)

### DIFF
--- a/lib/vagrant-libvirt/action/halt_domain.rb
+++ b/lib/vagrant-libvirt/action/halt_domain.rb
@@ -16,8 +16,14 @@ module VagrantPlugins
           domain = env[:machine].provider.driver.connection.servers.get(env[:machine].id.to_s)
           raise Errors::NoDomainError if domain.nil?
 
-          @logger.info('Trying gracefull shutdown.')
-          domain.shutdown
+          begin
+            env[:machine].guest.capability(:halt)
+          rescue
+            @logger.info('Trying libvirt graceful shutdown.')
+            domain.shutdown
+          end
+
+
           begin
             domain.wait_for(30) do
               !ready?


infernix:
AFAIK this just does an `ssh $vm "shutdown -h now"`, as opposed to a `virsh shutdown`. The thing is that a `virsh shutdown` already does a graceful halt by sending an ACPI poweroff to the VM which should be handled by any properly built box.

An ACPI poweroff even works if the VMs networking is not set up correctly or SSH access to it is broken for some reason. So I'm not sure in which case this is better?

linuxsimba:
I use it for vagrant setups where ACPI poweroff doesn't work like for Vyos VMs or Cumulus Linux VMs. Since the virtualbox provider has it, I figured it would make sense for the libvirt provider to have it to.

infernix:
We also support VMs with no `vagrant ssh` support; for example PXE booted VMs. But I guess the rescue covers it, have you tested that though?

linuxsimba:
I will confirm next week that it doesn't affected PXE booted VMs. I have seen no negative effects of this patch on windows VMs, centos and debian based VMs.

linuxsimba:
Not causing a problem with PXE VMs. [Full Output](https://gist.github.com/linuxsimba/2c62b33e7c01e19d71a9a19f85567bff)  from ``vagrant --debug halt `` on the PXE VM. 

Summary:
```
=> pxevm: Halting domain...
DEBUG ssh: Checking whether SSH is ready...
 INFO driver: Timeout at waiting for an ip address for machine pxevm
 INFO driver: No arp table entry found for machine pxevm
 INFO ssh: SSH not up: #<Vagrant::Errors::SSHNotReady: The provider for this Vagrant-managed machine is reporting that it
is not yet ready for SSH. Depending on your provider this can carry
different meanings. Make sure your machine is created and running and
try again. Additionally, check the output of `vagrant status` to verify
that the machine is in the state that you expect. If you continue to
get this error message, please view the documentation for the provider
you're using.>
 INFO halt_domain: Trying libvirt graceful shutdown.
```

infernix:
Thanks!

Issue #776, opened 2017-05-11
Unsupported XML configs?


polarathene:
I currently use virt-manager GUI to setup my VMs with libvirt. For GPU passthrough and some other tweaks I need to edit the XML directly. I am looking into migrating my config into being managed by this plugin, glancing over the docs some elements don't seem to be supported. 

Is there a possibility to support them or assuming behind the scenes a libvirt xml is created at `/etc/libvirt/qemu` should I manually add these tweaks myself?

Here are some of the tweaks:

### qemu commandline
This element lets you pass on CLI args to the final qemu command that libvirt creates(at least that's my understanding of what happens), the example is setting up audio routing from the guest to my host(an audio device still needs to be set such as ICH9).

```
  <qemu:commandline>
    <qemu:env name='QEMU_AUDIO_DRV' value='pa'/>
    <qemu:env name='QEMU_PA_SAMPLES' value='1024'/>
    <qemu:env name='QEMU_AUDIO_TIMER_PERIOD' value='150'/>
    <qemu:env name='QEMU_PA_SERVER' value='/run/user/1000/pulse/native'/>
  </qemu:commandline>
```

Supporting this element requires the domain element to be changed like this 

```
<domain type='kvm' xmlns:qemu='http://libvirt.org/schemas/domain/qemu/1.0'>
```

### vendor_id for nvidia hardware

In addition to hiding kvm which is mentioned in the docs, this change is also required, there isn't a specific value to use, so long as it's not whatever value(s) the driver rejects:

```
<hyperv>
      <relaxed state='on'/>
      <vapic state='on'/>
      <spinlocks state='on' retries='8191'/>
      <vendor_id state='on' value='1234567890ab'/>
</hyperv>
```

### clock offset (hpet)
```
  <clock offset='localtime'>
    <timer name='rtc' tickpolicy='catchup' track='guest'/>
    <timer name='pit' tickpolicy='delay'/>
    <timer name='hpet' present='yes'/>
    <timer name='hypervclock' present='yes'/>
  </clock>
```

Slightly different from the defaults I think. I originally had an issue with my host where resuming from suspend broke the `tsc` timesource falling back to `hpet` which was disabled by default iirc. Sometimes it's useful to disable `hpet` but this caused immense stuttering/lag when running the VM without `tsc` available, a BIOS update after contacting the vendor fixed that, not everyone may be as lucky to get a BIOS update so being able to enable hpet would be useful.

[...]

PR #827, opened 2017-11-12
Add memory backing configuration

petracvv:
I came across #776 when looking for hugepage support in this vagrant plugin and decided to implement that myself. 

This pull request gives users the ability to configure memoryBacking settings for their vagrant VM with a general form of:

```ruby
libvirt.memorybacking :Option, {Settings}
```

where `:Option` is one of 
* `:hugepages`
* `:nosharepages`
* `:locked`
* `:source`
* `:access`
* `:allocation`

and `Settings` is an Hash of settings and their values for the configured `:Option`. This Hash is only required for certain options as shown in the [libvirt documentation](https://libvirt.org/formatdomain.html#elementsMemoryBacking)

A full example is found in the README that is part of this request.

The `<page>` option under `<hugepages>` is not implemented as it does not seem to be a very common use case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -974,6 +974,24 @@ Vagrant.configure("2") do |config|
 end
 ```
 
+## Memory Backing
+
+You can specify memoryBacking options via `libvirt.memorybacking`. Available options are shown below. Full documentation is available at the [libvirt _memoryBacking_ section](https://libvirt.org/formatdomain.html#elementsMemoryBacking).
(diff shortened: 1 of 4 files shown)


infernix:
LGTM

petracvv:
Thanks for the merge!

pronix:
@petracvv thanks for PR :)

Issue #765, opened 2017-04-21
package failed to output to a directory


reidrac:
### Steps to reproduce
1.  mkdir `boxes`
2. run `vagrant package myvm --output boxes/test.box`

### Expected behaviour
The Vagrant box is saved into the `boxes/` directory.

### Actual behaviour
vagrant-libvirt crashes on rename:
```
tar (child): boxes/test.box: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
/opt/vagrant/embedded/lib/ruby/2.2.0/fileutils.rb:527:in `rename': No such file or directory @ sys_fail2 - (/home/user/vagrant/_tmp_package/boxes/test.box, ../boxes/test.box) (Errno::ENOENT)
        from /opt/vagrant/embedded/lib/ruby/2.2.0/fileutils.rb:527:in `block in mv'
        from /opt/vagrant/embedded/lib/ruby/2.2.0/fileutils.rb:1570:in `block in fu_each_src_dest'
        from /opt/vagrant/embedded/lib/ruby/2.2.0/fileutils.rb:1586:in `fu_each_src_dest0'
        from /opt/vagrant/embedded/lib/ruby/2.2.0/fileutils.rb:1568:in `fu_each_src_dest'
        from /opt/vagrant/embedded/lib/ruby/2.2.0/fileutils.rb:516:in `mv'
        from /home/user/.vagrant.d/gems/2.2.5/gems/vagrant-libvirt-0.0.37/lib/vagrant-libvirt/action/package_domain.rb:47:in `call'
        from /opt/vagrant/embedded/gems/gems/vagrant-1.9.3/lib/vagrant/action/warden.rb:34:in `call'
        from /opt/vagrant/embedded/gems/gems/vagrant-1.9.3/lib/vagrant/action/builtin/config_validate.rb:25:in `call'
        from /opt/vagrant/embedded/gems/gems/vagrant-1.9.3/lib/vagrant/action/warden.rb:34:in `call'
        from /opt/vagrant/embedded/gems/gems/vagrant-1.9.3/lib/vagrant/action/builder.rb:116:in `call'
        from /opt/vagrant/embedded/gems/gems/vagrant-1.9.3/lib/vagrant/action/runner.rb:66:in `block in run'
        from /opt/vagrant/embedded/gems/gems/vagrant-1.9.3/lib/vagrant/util/busy.rb:19:in `busy'
[...]

PR #836, opened 2017-11-23
Fix error in package command

n-rodriguez:
Fix https://github.com/vagrant-libvirt/vagrant-libvirt/issues/765

### DIFF
--- a/lib/vagrant-libvirt/action/package_domain.rb
+++ b/lib/vagrant-libvirt/action/package_domain.rb
@@ -62,7 +62,9 @@ module VagrantPlugins
           File.write(@tmp_dir + '/metadata.json', metadata_content(img_size))
           File.write(@tmp_dir + '/Vagrantfile', vagrantfile_content)
           assemble_box(boxname, extra)
-          FileUtils.mv(@tmp_dir + '/' + boxname, '../' + boxname)
+          source_file = @tmp_dir + '/' + boxname
+          dest_file = '../' + boxname
+          FileUtils.mv(source_file, dest_file) if File.exist?(source_file)
           FileUtils.rm_rf(@tmp_dir)
           env[:ui].info('Box created')
           env[:ui].info('You can now add the box:')
--- a/locales/en.yml
+++ b/locales/en.yml
@@ -49,6 +49,8 @@ en:
       Machine is booted and ready for use!
     remove_stale_volume: |-
       Remove stale volume...
+    package_domain: |-
+      Packaging domain...
 
     warnings:
       ignoring_virtual_size_too_small: |-


n-rodriguez:
Hi! Actually this patch fixes a bug that happens with [test-kitchen](https://github.com/test-kitchen/test-kitchen) and [kitchen-vagrant](https://github.com/test-kitchen/kitchen-vagrant) : 

```
nicolas@nicolas-laptop:~/PROJECTS/saltstates$ kitchen package

-----> Package on sys-web-fb-stretch-amd64.
-----> Packaging remote instance
$$$$$$ Disable vagrant ssh key replacement to preserve the default key!
       Failed to load /etc/gemrc because it doesn't contain valid YAML hash
       ==> default: translation missing: en.vagrant_libvirt.package_domain
       qemu-img: Could not open '/var/lib/libvirt/images/sys-web-fb-stretch-amd64_default_1511452676_407e49c1588909474850.img': Failed to get shared "write" lock
       ==> default: Image has backing image, copying image and rebasing ...
       Nombre total d'octets écrits: 3045990400 (2,9GiB, 22MiB/s)
       /usr/lib/ruby/2.4.0/fileutils.rb:469:in `rename': No such file or directory @ rb_file_s_rename - (/home/nicolas/PROJECTS/saltstates/.kitchen/kitchen-vagrant/sys-web-fb-stretch-amd64/_tmp_package//home/nicolas/PROJECTS/saltstates/sys-web-fb-stretch-amd64.box, ..//home/nicolas/PROJECTS/saltstates/sys-web-fb-stretch-amd64.box) (Errno::ENOENT)
       	from /usr/lib/ruby/2.4.0/fileutils.rb:469:in `block in mv'
       	from /usr/lib/ruby/2.4.0/fileutils.rb:1461:in `block in fu_each_src_dest'
       	from /usr/lib/ruby/2.4.0/fileutils.rb:1477:in `fu_each_src_dest0'
       	from /usr/lib/ruby/2.4.0/fileutils.rb:1459:in `fu_each_src_dest'
       	from /usr/lib/ruby/2.4.0/fileutils.rb:458:in `mv'
```

kitchen-vagrant calls `package` command with the `--output` param : https://github.com/test-kitchen/kitchen-vagrant/blob/master/lib/kitchen/driver/vagrant.rb#L153.
It packages a new image in the current directory, not the temps dir, thus the error when moving the generated file.

n-rodriguez:
I guess the patch is to check for file presence before moving it ;)

infernix:
kitchen uses `        box_name = File.join(Dir.pwd, instance.name + ".box")` but we cannot rely on that. Users might not (and probably won't) specify a full path. So the proposed fix cannot be merged, but I'll take any patch that makes it work in these cases:

`vagrant package`
`vagrant package --output foo.box`
`vagrant package --output relative_path/foo.box`
`vagrant package --output /absolute_path/foo.box`

n-rodriguez:
I've updated the patch, do you think it's mergeable?

infernix:
This makes the code only more unreadable and I think it would fail with an absolute path?

n-rodriguez:
> This makes the code only more unreadable and I think it would fail with an absolute path?

Ok, so what do you suggest to fix the bug?

infernix:
> In general this code seems kludgy, but I think the right approach is to parse env['package.output'] into a boxpath and boxname variable (using File.split for example) and to use the full boxpath+boxname in the right places (such as the tar cvzf "#{boxpath}/#{boxname}")

Issue #831, opened 2017-11-20
Vagrant up gives me error: Call to virDomainDefineXML failed: internal error: Unknown source mode '' when using "org.qemu.guest_agent.0"


ruzickap:
Hello

I'm not able to run ```vagrant up``` when using the latest Fedroa / libvirt / vagrant with this parameters in Vagrantfile:

```
libvirt.channel :type => 'unix',     :target_name => 'org.qemu.guest_agent.0', :target_type => 'virtio'
```

I'm getting the following error (details are below):

```
Error while creating domain: Error saving the server: Call to virDomainDefineXML failed: internal error: Unknown source mode ''
```

### Steps to reproduce
1. ```vagrant init centos/7```
2. Add the following to the Vagrantfile:
```
  config.vm.provider :libvirt do |libvirt|
    libvirt.channel :type => 'unix',     :target_name => 'org.qemu.guest_agent.0', :target_type => 'virtio'
  end
```
3. ```vagrant up```

### Expected behaviour
VM should start like normal when using ```vagrant up``` 

### Actual behaviour
Getting error and Vagrant stops

### System configuration
**OS/Distro version:** 
Fedora 27

**Libvirt version:**
```
$ rpm -qa libvirt-libs
libvirt-libs-3.7.0-2.fc27.x86_64
```

**Output of `vagrant version; vagrant plugin list`:**
```
$ vagrant version; vagrant plugin list
Installed Version: 2.0.1
Latest Version: 2.0.1

You're running an up-to-date version of Vagrant!
vagrant-hostmanager (1.8.7)
vagrant-libvirt (0.0.40)
vagrant-share (1.1.9, system)
vagrant-winrm (0.7.0)
```

**Output of `VAGRANT_LOG=debug vagrant up --provider=libvirt`**
```
VAGRANT_LOG=debug vagrant up --provider=libvirt
...
...
...
==> default:  -- TPM Path:          
 INFO interface: info:  -- INPUT:             type=mouse, bus=ps2
 INFO interface: info: ==> default:  -- INPUT:             type=mouse, bus=ps2
==> default:  -- INPUT:             type=mouse, bus=ps2
 INFO interface: info:  -- CHANNEL:             type=unix, mode=
 INFO interface: info: ==> default:  -- CHANNEL:             type=unix, mode=
==> default:  -- CHANNEL:             type=unix, mode=
[...]

PR #838, opened 2017-11-28
Fix for #831

ableischwitz:
This patch-set should fix the issue with recent libvirt on Fedora 27.

I did not validate on older versions, but shouldn't do any harm as it removes the empty "mode=''" entries from the XML-definition which libvirt now is complaining about.


infernix:
LGTM, thanks!

PR #852, opened 2018-01-17
Add option to specify CPU topology

chulup:
By default libvirt creates very strange CPU topology of multiple sockets of one core each. 

This pull request allows user to write an exact topology in Vagrantfile in the form of ```libvirt.cputopology``` call:

  ```ruby
      libvirt.cpus = 4
      libvirt.cputopology :sockets => '2', :cores => '2', :threads => '1'
    end
  end
  ```
User have to provide ```libvirt.cpus``` option and make sure its value is equal to multiple of all numbers given to ```libvirt.cputopology``` call, just as required by libvirt.


pronix:
thanks. looks good. i will check today evening.

Issue #272, opened 2014-11-08
Support user session libvirtd (qemu:///session)


mdekstrand:
Right now, vagrant-libvirt seems to basically require the system libvirt instance (`qemu:///system`). It would be nice to be able to use my user session libvirt (`qemu:///session`) instead of the systemwide one (fewer password prompts, for one thing).

There are two ways this problem manifests at present:

- No obvious setting, besides `uri`, to specify the session instance.
- When specifying `uri = "qemu:///session"`, `vagrant up` fails. The current failure is basically #247, although I already have version 0.23 installed (which should contain the fix?).

bersace:
With qemu-bridge-helper, it is possible to plug to an existing bridge. This is how GNOME Boxes is supposed to work.

purpleidea:
So vagrant+vagrant-libvirt look like they're finally getting into Fedora 22. w00t.
https://bugzilla.redhat.com/show_bug.cgi?id=1168333

One issue is that it was decided that the polkit file will not be included, meaning many password prompts by default. As a result, it makes a lot of sense to have vagrant-libvirt work with the qemu:///session instead of the system one.

As @sciurus commented, we have the network issue, but perhaps this could use the solution proposed by @bersace.

Any patches or recommendations welcome. If someone can help on the network side, I can probably patch the ruby side to support the qemu:///session aspect.

Cheers!

purpleidea:
Ping :)

Bounty for this issue now contains some beverages...

purpleidea:
Ping pong, ding dong. Bounty now also includes a cool t-shirt and/or a hat or something.

purpleidea:
FYI: this issue getting fixed would directly help: https://bugzilla.redhat.com/show_bug.cgi?id=1187019

stefwalter:
@zeenix Can gnome-boxes users use non user-mode networks? Is it something you support?

@purpleidea I found this. http://xkahn.zoned.net/blog/2013/11/26/networking-and-gnome-boxes/

In vagrant-libvirt we probably need to use virsh to setup the network on the system libvirt, and then use the bridge network interface from the session defined domain (ie: vm).

purpleidea:
@stefwalter setting up a single network for the machine discovery and that is shared with every vagrant environment would be doable, however each user vagrant environment would need it's own custom static network, and this would need some sort of user mode network which I'm not quite sure how to build at the moment.

Maybe @strzibny is looking into this?

stefwalter:
[...]

PR #868, opened 2018-02-23
Qemu Session Support

qazokm:
Added basic qemu support in response to issue #272. The patch supports creating storage pools under user directories and using existing networks created under the qemu:///system connection. Using existing networks requires the qemu-bridge-helper to be installed and the relevant permissions granted (this is the default on some distros but others may need to configure this).

To assist with this the following configuration options where added
-  **qemu_use_session** 
    - Enable usage of qemu session, changes default connection URL to qemu:///session and enables support for qemu session
-  **storage_pool_path**
    -  Storage pool path to use for libvirt images 
    - Defaults to /var/lib/libvirt/images
    - If session enabled defaults to ~/.local/share/libvirt/images)
-  **management_network_device**
    - Device to use for management network
    - Defaults to 'virbr0'
-  **system_uri**
    - QEMU system URL to connect to as read-only to retrieve DHCP leases for obtaining IP addresses

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ can help a lot :-)
 - [SSH Access To VM](#ssh-access-to-vm)
 - [Forwarded Ports](#forwarded-ports)
 - [Synced Folders](#synced-folders)
+- [QEMU Session Support](#qemu-session-support)
 - [Customized Graphics](#customized-graphics)
 - [Box Format](#box-format)
 - [Create Box](#create-box)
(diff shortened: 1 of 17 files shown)


pronix:
hello @qazokm 
thanks for huge impact. it will take time for testing.
can you also add some information to README, for example same as in description.
thank you

qazokm:
No problem, added a basic example with links to QEMU bridge helper information

pronix:
@qazokm thank you. check please some changes https://github.com/vagrant-libvirt/vagrant-libvirt/commit/99371fee2b24e5f58793b7ce86824a47491d0372

voxik:
@qazokm nice work :+1: but have you considered to submit part of your changes to fog-libvirt? E.g. default_storage_volume.xml.erb would not be needed here if the owner and group tags were supported by fog-libvirt and on the first look, it just looks it would be just a minor change.

PR #869, opened 2018-03-02
QEMU libvirt nvram support

herzogt:
Some UEFI firmwares may want to use a non-volatile memory to store
variables.
This requires to specify loader and nvram to use UEFI boot in QEMU.
Specifying loader and nvram at the same time will set loader to
type 'pflash' instead of 'rom'.
If loader is used without nvram option type will remain 'rom'.

Further information can be found at libvirt documentation:
https://libvirt.org/formatdomain.html#elementsOS


eoli3n:
@herzogt  Is this working ?

eoli3n:
Yes it is.

PR #870, opened 2018-03-03
Allow specifying Hypervisor HyperV features

tuxillo:


tuxillo:
Hi,

This is related to issue #776 . It is a first attempt to allow specifying the HyperV features for the hypervisor which seems to be quite convenient for Windows guests.

The PR is quite not perfect because it can only set the feature's state but it can't set values and I didn't want to hardcode libvirt specifics in here, any pointer is appreciated :-)

Regards,
Antonio Huete

lib/vagrant-libvirt/config.rb
```diff
@@ -321,6 +323,17 @@ def cpu_feature(options = {})
...
 
+      def hyperv_feature(options = {})
+        if options[:name].nil? || options[:state].nil?
+          raise 'Feature name AND state must be specified'
+        end
+
+        @features_hyperv = [] if @features_hyperv == UNSET_VALUE
+
+        @features_hyperv.push(name:  options[:name],
+                              state: options[:state])
```
pronix: simplification
```
@features_hyperv = [{name:  options[:name],state: options[:state]}]  if @features_hyperv == UNSET_VALUE
```

tuxillo:
I've added the simplification you specified.

Issue #866, opened 2018-02-19
CPU feature 'vmx' can't be overwritten when libvirt.nested = true


anastasiiavlaskina:
### Steps to reproduce
1. libvirt.nested = true 
    libvirt.cpu_feature :name => 'vmx', :policy => 'require'  in Vagrantfile
2. vagrant up

### Expected behaviour
A vm shoud be brought up and support nested virtualization with vmx require parameter.

### Actual behaviour
Error while creating domain: Error saving the server: Call to virDomainDefineXML failed: XML error: CPU feature 'vmx' specified more than once

### System configuration
**OS/Distro version:**: Ubuntu 16.04

**Libvirt version:** libvirtd (libvirt) 2.0.0

**Output of `vagrant version; vagrant plugin list`:**
hostmanager (1.2.2)
  - Version Constraint: > 0
vagrant-hostmanager (1.8.7)
  - Version Constraint: > 0
vagrant-libvirt (0.0.40)
  - Version Constraint: > 0
vagrant-share (1.1.9, system)
  - Version Constraint: > 0

**Output of `VAGRANT_LOG=debug vagrant up --provider=libvirt`**
```
ERROR vagrant: Error while creating domain: Error saving the server: Call to virDomainDefineXML failed: XML error: CPU feature 'vmx' specified more than once
ERROR vagrant: /home/vlaskina/.vagrant.d/gems/2.4.2/gems/vagrant-libvirt-0.0.40/lib/vagrant-libvirt/action/create_domain.rb:302:in `rescue in call'
/home/vlaskina/.vagrant.d/gems/2.4.2/gems/vagrant-libvirt-0.0.40/lib/vagrant-libvirt/action/create_domain.rb:297:in `call'
/opt/vagrant/embedded/gems/gems/vagrant-2.0.1/lib/vagrant/action/warden.rb:34:in `call'
/home/vlaskina/.vagrant.d/gems/2.4.2/gems/vagrant-libvirt-0.0.40/lib/vagrant-libvirt/action/create_domain_volume.rb:82:in `call'
/opt/vagrant/embedded/gems/gems/vagrant-2.0.1/lib/vagrant/action/warden.rb:34:in `call'
/home/vlaskina/.vagrant.d/gems/2.4.2/gems/vagrant-libvirt-0.0.40/lib/vagrant-libvirt/action/handle_box_image.rb:113:in `call'
[...]

PR #871, opened 2018-03-05
Remove nested section in domain template

anastasiiavlaskina:
An unnecessary nested xml section was removed in the domain template, since code in start_domain.rb handles this logic.


pronix:
https://github.com/vagrant-libvirt/vagrant-libvirt/issues/866

Issue #759, opened 2017-04-12
packaged box got no sshd running because of virt-sysprep


ch0k0bn:
### Steps to reproduce
1. create a linux box
2. package it
3. use your newly created packaged box with vagrant

### Expected behaviour

vagrant should be able to connect to the vm and interact with it

### Actual behaviour

==> vm-test: Creating shared folders metadata...
==> vm-test: Starting domain.
==> vm-test: Waiting for domain to get an IP address...
==> vm-test: Waiting for SSH to become available...
==> vm-test: Removing domain...
/opt/vagrant/embedded/gems/gems/net-ssh-3.0.2/lib/net/ssh/proxy/command.rb:93:in `read_nonblock': end of file reached (EOFError)

### System configuration
**OS/Distro version:**: Ubuntu 16.04.2 LTS

**Libvirt version:** Version: 1.3.1-1ubuntu10.8

virt-sysprep remove hostkeys by default.
Don't know if it is new or not in debian8 (the target os) but without hostkeys, sshd will never start making the packaged box not functionnal.

I temporary fixed it by removing the specific action (ssh-hostkeys) from virt-sysprep.

ch0k0bn:
my patch:
in lib/vagrant-libvirt/action/package_domain.rb

`virt-sysprep --no-logfile --operations defaults,-ssh-userdir,-ssh-hostkeys -a #{@tmp_img} `

infernix:
Although that works it is not ideal since the guest should be creating ssh keys for each new instantiation, so leaving this open for now.

pronix:
it should be customized options for package. i will pick it up soon

electrofelix:
https://github.com/libguestfs/libguestfs/blame/f4ca4de207f33847871a95e686997476b82ec1af/sysprep/sysprep_operation_ssh_hostkeys.ml suggests that the operation ssh-hostkeys has been enabled by default for 4 years.

ericpulvino:
I hit this issue as well when packaging a box. The workaround from @ch0k0bn worked for me. I cannot speak to whether this worked in previous releases.

zakame:
This seems to be due to Vagrant expecting to find the known [insecure keypair](https://github.com/hashicorp/vagrant/tree/master/keys) tested in https://github.com/hashicorp/vagrant/blob/b0fc33e9b728f8575890a5a9e358394983b5298c/plugins/communicators/ssh/communicator.rb#L702-L707

[...]

PR #873, opened 2018-03-22
Enable provisioning by providing ssh access -ssh-hostkeys parameter to virt-sysprep step

gpenaud:
The actual configuration for virt-sysprep step (lib/vagrant-libvirt/action/package_domain.rb) removes all of ssh server hostkeys. The vagrant machines are unable to connect through ssh, and hangs indefinitely.

I just added the keys generation to virt-sysprep command, like shown on https://github.com/vagrant-libvirt/vagrant-libvirt/issues/759

Thx

### DIFF
--- a/lib/vagrant-libvirt/action/package_domain.rb
+++ b/lib/vagrant-libvirt/action/package_domain.rb
@@ -38,7 +38,7 @@ module VagrantPlugins
           `qemu-img rebase -p -b "" #{@tmp_img}`
           # remove hw association with interface
           # working for centos with lvs default disks
-          `virt-sysprep --no-logfile --operations defaults,-ssh-userdir -a #{@tmp_img}`
+          `virt-sysprep --no-logfile --operations defaults,-ssh-userdir,-ssh-hostkeys -a #{@tmp_img}`
           # add any user provided file
           extra = ''
           @tmp_include = @tmp_dir + '/_include'


lib/vagrant-libvirt/action/package_domain.rb
```diff
@@ -38,7 +38,7 @@ def call(env)
           `qemu-img rebase -p -b "" #{@tmp_img}`
           # remove hw association with interface
           # working for centos with lvs default disks
-          `virt-sysprep --no-logfile --operations defaults,-ssh-userdir -a #{@tmp_img}`
+          `virt-sysprep --no-logfile --operations defaults,-ssh-userdir,-ssh-hostkeys -a #{@tmp_img}`
```
electrofelix: It would be more useful to allow a use to provide some options, but also the problem with including this is that after packaging, `vagrant up` should result in different hostkeys being generated for each run.

If you can change this to read from the environment maybe `VAGRANT_LIBVIRT_VIRT_SYSPREP_OPERATONS` is a reasonable option to use?

Something like the following seems like it should work?
```
          operations = ENV.get('VAGRANT_LIBVIRT_VIRT_SYSPREP_OPERATONS', 'defaults,-ssh-userdir')
          `virt-sysprep --no-logfile --operations #{operations} -a #{@tmp_img}`
```

electrofelix: Will also want to update the README.md to mention this being available, it doesn't really solve the problem because the ssh hostkeys should get re-created, but at least it means that anyone hitting various problems around the packaging with virt-sysprep can adjust as needed.

electrofelix:
Looks like there is now a solution for this with #955 so unless there is another requirement here, believe this can be closed.

PR #875, opened 2018-03-24
Allow usage of public/private networks as management networks

gpenaud:
Hello !

I was very bored by the fact i had to use a management network to setup my VM ; I already have a public network which could be used for that.
Here's what i did. I know this is very dirty (i'm far from being a ruby expert), but I'm sure you'll see the idea. 

Thanks for your returns

### DIFF
--- a/lib/vagrant-libvirt/action/wait_till_up.rb
+++ b/lib/vagrant-libvirt/action/wait_till_up.rb
@@ -17,11 +17,15 @@ module VagrantPlugins
         end
 
         def call(env)
+
           # Initialize metrics if they haven't been
           env[:metrics] ||= {}
 
+          # Get driver object
+          driver = env[:machine].provider.driver
           # Get domain object
-          domain = env[:machine].provider.driver.get_domain(env[:machine].id.to_s)
+          domain = driver.get_domain(env[:machine].id.to_s)
+
           if domain.nil?
             raise Errors::NoDomainError,
                   error_message: "Domain #{env[:machine].id} not found"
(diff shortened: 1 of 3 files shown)


lib/vagrant-libvirt/action/wait_till_up.rb
```diff
@@ -31,39 +35,28 @@ def call(env)
...
+
+              # Wait for domain to obtain an ip address
+              domain.wait_for(2) do
+                addresses.each_pair do |_type, ip|
+                  ip_address = driver.get_ipaddress(env[:machine])
+
+                  if !ip[0].nil?
+                    env[:ip_address] = ip[0]
+                  elsif ip_address != ""
+                    env[:ip_address] = ip_address
```
pronix:
```
env[:ip_address] = if !ip[0].nil?
                                     ip[0]
                                   elsif ip_address != ""
                                     ip_address
                                   end

```

lib/vagrant-libvirt/driver.rb
```diff
@@ -99,7 +99,14 @@ def get_ipaddress(machine)
             addresses.each_pair do |_type, ip|
               # Multiple leases are separated with a newline, return only
               # the most recent address
-              ip_address = ip[0].split("\n").first unless ip[0].nil?
+              unless ip[0].nil?
```
pronix: use if and change 1st and 2nd branches order

lib/vagrant-libvirt/util/network_util.rb
```diff
@@ -70,10 +70,10 @@ def configured_networks(env, logger)
...
           end
 
-          if (env[:machine].config.vm.box &&
-              !env[:machine].provider_config.mgmt_attach)
-            raise Errors::ManagementNetworkRequired
-          end
+          # if (env[:machine].config.vm.box &&
+          #     !env[:machine].provider_config.mgmt_attach)
+          #   raise Errors::ManagementNetworkRequired
+          # end
```
pronix: it should be option for switch behaviour and default should be old one.

jabreity:
This is an awesome feature, I would like to see soon!

electrofelix:
Replaced by #1411 as I can't update the original PR any more.

Sorry it took so long, effectively needed qemu agent to avoid various issues here if attempting to perform address discovery and not relying on dhcp addresses.

Future improvement would be to use hostfwd so that it would be possible to remove the need for the management network completely but would also require being able to support connecting to remote machines to completely be able to remove the dependency.

PR #884, opened 2018-04-18
Only use 'configure_networks' capability if there actually are networks to be configured

anton-dessiatov:


syndbg:
@electrofelix Can we get this released?

electrofelix:
well oops

PR #889, opened 2018-05-07
README (Debian install): Don't autorun dnsmasq

dabreese00:
Current install instructions for Debian/Ubuntu install the package
"dnsmasq", which results in dnsmasq being immediately started and
configured to run as a system service.  This is probably not something
vagrant-libvirt users typically want.

Instead, package "dnsmasq-base" exists, which achieves precisely what is
probably desired here.


electrofelix:
ah, yes, noticed I don't have `dnsmasq` installed locally, just `dnsmasq-base`, thanks

PR #895, opened 2018-05-22
Default Port Forwarding to All Host Addresses

mattandersen:
This change would make it so that the default for forwarded ports would be to listen on all ports. This change brings vagrant-libvirt in line with the default set of plugins, that function this way. It also makes sharing configurations between multiple hosts easier, as the VirtualBox provider (at least on Windows 10) will fail to start the guest VM if "host_ip: *" is used.


electrofelix:
@pradels @pronix either have any objections to switching to this behaviour? Does it require that the next version released requires a bigger version bump to note that there has been a behaviour change?

PR #906, opened 2018-07-31
change dependency for fog-core to ~> 2.1

johanneskastl:
This solves #905.

Needs to be tested if there are regressions due to the new major version.


electrofelix:
Duplicated in #970 if you can update the PR to use a more flexible version to `'>= 1.43.0', '<2.2.0'` which covers the entire span, that can be merged, at least until I can make sure that dropping support for the older fog-core, which still works, doesn't impact anything.

electrofelix:
Looks like #970 is being updated, and having looked at when ruby 2.0 was the default for distros, and the fact that vagrant 1.6 which is the oldest listed supported version for vagrant-libvirt, I think it should be safe to skip the additional supported versions I mentioned last week.

So either if this or #970 get updated to match the existing code without conflicts and use the pin `~> 2.1` I'll merge the first to ready.

electrofelix:
resolved by #970

PR #907, opened 2018-08-06
Allow setting the CDROM image type

iam-TJ:
Although it is possible to attach additional images (disk or cdrom) the cdrom functionality doesn't allow setting the :type - it defaults to 'raw' since vagrant-libvirt is assuming cdrom images are in ISO9660 format.

This assumption breaks down when using vagrant-mutate to convert a box with multiple storage devices. A common situation is the Ubuntu boxes built for VirtualBox that contain a cloud-init ISO9660 image with user-data. vagrant-mutate converts the image to qcow* format (not ISO9660) in order to not have to deal with what is inside.

This means the resulting libvirt domain needs to be told the type is 'qcow2' in order for the guest to see it as an ISO9660 image. 

This patch adds the required options and allows, in a Vagrantfile, something like:
```

  libvirt.storage :file, :device => :cdrom, :type => 'qcow2', :path => '/home/tj/.vagrant.d/boxes/ubuntu-VAGRANTSLASH-bionic0/0/libvirt/box2.img'
 ...
end
```


spec/unit/templates/domain_all_settings.xml
```diff
@@ -48,11 +48,13 @@
     </disk>
 
     <disk type='file' device='cdrom'>
+      <driver name='qemu' type='' />
```
electrofelix: This doesn't seem correct, surely it's required to specify a valid type? Have a look at https://libvirt.org/formatdomain.html#elementsDisks, while it's valid to have `type` set to `qcow2` as you want to be able to, if specifying `type` it reads as though it needs to be a valid value. Making it default to `raw` as mentioned in the config code should be sufficient.

lib/vagrant-libvirt/config.rb
```diff
@@ -545,6 +545,7 @@ def _handle_cdrom_storage(options = {})
         }.merge(options)
 
         cdrom = {
+          type: options[:type],
```
electrofelix: It would seem like a default `type` of `raw` should be defined in the options hash above.

electrofelix (dismissed):

PR #923, opened 2018-10-02
Allow specification of cpuset

QuintenJohnson:
Allows the pinning of vcpus to physical cpus.

### DIFF
--- a/lib/vagrant-libvirt/templates/domain.xml.erb
+++ b/lib/vagrant-libvirt/templates/domain.xml.erb
@@ -2,8 +2,11 @@
   <name><%= @name %></name>
   <uuid><%= @uuid %></uuid>
   <memory><%= @memory_size %></memory>
-  <vcpu><%= @cpus %></vcpu>
-
+  <vcpu
+      <% if @cpuset %>
+          cpuset='<%= @cpuset %>'
+      <% end %>
+  ><%= @cpus %></vcpu>
 
   <cpu mode='<%= @cpu_mode %>'>
     <% if @cpu_mode != 'host-passthrough' %>
--- a/lib/vagrant-libvirt/templates/domain.xml.erb
+++ b/lib/vagrant-libvirt/templates/domain.xml.erb
@@ -2,11 +2,8 @@
   <name><%= @name %></name>
   <uuid><%= @uuid %></uuid>
   <memory><%= @memory_size %></memory>
-  <vcpu
-      <% if @cpuset %>
-          cpuset='<%= @cpuset %>'
-      <% end %>
-  ><%= @cpus %></vcpu>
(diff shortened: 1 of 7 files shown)


lib/vagrant-libvirt/templates/domain.xml.erb
```diff
@@ -2,8 +2,11 @@
   <name><%= @name %></name>
   <uuid><%= @uuid %></uuid>
   <memory><%= @memory_size %></memory>
-  <vcpu><%= @cpus %></vcpu>
-
+  <vcpu
+      <% if @cpuset %>
+          cpuset='<%= @cpuset %>'
+      <% end %>
```
electrofelix: if you change the above to:
```
    <%- if @cpuset -%>
        cpuset='<%= @cpuset %>'
    <%- end -%>
```
it should correctly pass the template rendering test.

QuintenJohnson: There were still newlines after those modifications, so I changed it up until I produced the result expected by the tests. I also added some expected output to the unit test for all settings.

QuintenJohnson:
I made those changes and added a new commit, but I found a different issue with this patch. It appears that setting vcpu above 1 occasionally causes issues. The fault occurs in lib/vagrant-libvirt/action/start_domain.rb at line 71. I've fixed it by making this change, but I will say that while it fixes the issue, I haven't verified that it gives us back the original behavior:

      # vCpu count
    - if config.cpus.to_i != libvirt_domain.vcpus.length
    + if config.cpus.to_i != libvirt_domain.num_vcpus(0)

electrofelix:
@QuintenJohnson Looking at the code where you pointed out the issue with https://github.com/vagrant-libvirt/vagrant-libvirt/blob/master/lib/vagrant-libvirt/action/start_domain.rb#L71 do you know what was getting returned by `libvirt_domain.vcpus` in the first place before it checks it's length?

I think will need to inspect that and see what's returned in order to determine whether it will result in the same behaviour.

QuintenJohnson:
I believe libvirt_domain.vcpus returns structures for each vcpu in the domain, so .length would get the number of those structures. It wasn't returning anything when I hit the issues, so .length caused a fault. .num_vcpus() probably does what was originally intended more directly, but it doesn't explain why .vcpus was returning nothing.

electrofelix:
That would make sense as to why it should be working, definitely odd that the existing code failed, but `num_vcpus` does seem to be more accurate as to what is needed and the API call appears to have existed for 8 years in ruby-libvirt, so I'm guessing it's been around long enough that it shouldn't be an issue.

I think maybe update to use `num_vcpus(0)` as that seems more in line with the original intent.

PR #929, opened 2018-10-18
Revert back to development version of .gemspec file.

voxik:
Reverts commit 34a0def7d1d369a692ea5a431e127061403f6ec6, because it seems to be completely accidental change, when the file was overridden by Bundler or so.

At minimum, it does not make sense to specify the development dependencies like this and reference ```installed_by_version``` etc.


voxik:
Just FTR, I stumbled upon this when I was right about to submit PR relaxing fog-core dependency.

electrofelix:
@pronix this looks like it was caused by an accidental change?

PR #939, opened 2018-11-09
remove gemspec use of `git ls-files`

eighthave:
In 5158b0e733ba22aeb61266335b7901b347eaab0d, @pronix added `git ls-files` to list the _files_ and _test_files_.  This will never work for any process that uses a source tarball, like Debian packages and I think gems.  Using `git ls-files` requires _git_ is installed on the build machine and requires that the _.git/_ dir is present in the source.

Example fixes:
* jfelchner/ruby-progressbar#54
* rubygems/rubygems#2064
* wpscanteam/CMSScanner#64


vagrant-libvirt.gemspec
```diff
@@ -8,8 +8,8 @@ Gem::Specification.new do |s|
   s.required_rubygems_version = Gem::Requirement.new(">= 0".freeze) if s.respond_to? :required_rubygems_version=
   s.require_paths = ["lib".freeze]
   s.authors = ["Lukas Stanek".freeze, "Dima Vasilets".freeze, "Brian Pitts".freeze]
-  s.files         = `git ls-files`.split($\)
-  s.test_files    = s.files.grep(%r{^(test|spec|features)/})
+  s.files = Dir.glob("{lib,locales}/**/*.*")
```
electrofelix: Think this needs to include `LICENSE` & `README.md`.

PR #941, opened 2018-11-10
add domain name for the management network

evgeni:
this allows users to have a dedicated domain for their vagrant VMs and point their regular resolver at the libvirt dnsmasq for that domain to make them reachable by a full name instead of the IP when accessing the services on said VM.

 an example for such a setup can be found in [1] and [2]

 [1] https://liquidat.wordpress.com/2017/03/03/howto-automated-dns-resolution-for-kvmlibvirt-guests-with-a-local-domain/
 [2] https://m0dlx.com/blog/Automatic_DNS_updates_from_libvirt_guests.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -699,9 +699,10 @@ used by this network are configurable at the provider level.
   for for more information.
 * `management_network_autostart` - Automatic startup of mgmt network, if not
   specified the default is 'false'.
-* `:management_network_pci_bus` -  The bus of the PCI device.
-* `:management_network_pci_slot` -  The slot of the PCI device.
+* `management_network_pci_bus` -  The bus of the PCI device.
+* `management_network_pci_slot` -  The slot of the PCI device.
 * `management_network_mac` - MAC address of management network interface.
+* `management_network_domain` - Domain name assigned to the management network.
 
 You may wonder how vagrant-libvirt knows the IP address a VM received.  Libvirt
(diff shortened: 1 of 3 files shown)


electrofelix:
@evgeni what's the benefit of this? Could you update the commit message so that someone in the future would know why it's been added. And I'll look to merge it subsequently then.

evgeni:
@electrofelix updated commit message and PR description. does it make sense now?

electrofelix:
@evgeni thanks, I suspected it might have something to do with that. Is it common to want to set this on the management network, which is intended as a private communication channel into the VM as opposed to the user defined networks which allows this to be set via `:libvirt__domain_name`?

Is it a case that the option wasn't well known, or is this just to cover the additional case where no networks are defined because you'd just like to reuse the management network?

evgeni:
The option is known, yeah. But in my/our usecase we don't define extra networks as the management one is enough and we already get it configured "for free" :)

(https://github.com/theforeman/forklift if you're curious ;))

electrofelix:
After asking for an update the PR description, I forgot that Github doesn't use it for the merge automatically, I'm clearly spoiled by my workplace tooling...

evgeni:
Not for the merge, yes, but my original commit contained the same message: https://github.com/vagrant-libvirt/vagrant-libvirt/commit/53539f6ad4f434f55145cb83f68b5c553b08f165

Issue #952, opened 2018-12-04
support different storage pool for base box snapshot


abbbi:
Currently there is only one option to specify a storage pool where the box base images and the snapshots are stored. The base box image and the snapshot are stored alongside in the same storage pool.

It would be nice if one could also specify a special storage pool for the snapshot that is created. As it is an external snapshot from the libvirt/qemu view, this shall be possible.

Example:

1) base image is in the "default" storage pool /var/lib/libvirt/images/base_box_image.img
2) one sets something like 

snapshot_pool_name = "snapshots"
snapshot_pool_path = "/tmp/snapshot/"

3) The resulting base image snapshot is placed in /tmp/snapshot/base_box_image_xxx.img

This can have effect on performance, as the base box images for example can be stored on regular disks, and the snapshots which actually have to handle write operations shall be stored on SSD disks.

abbbi:
See commit in my branch.

It allows to set option like this:
  config.vm.provider :libvirt do |domain|
        domain.snapshot_pool_name = "snapshots"
  end

this will create base box snapshot in libvirt pool "snapshots", if it exists:

==> default:  -- Base box:          generic/debian9
==> default:  -- Storage pool:      default
==> default:  -- Image:             /tmp/snapshots/mynet_default.img (32G)

this way one can split base box volume and snapshot volume.

electrofelix:
@pronix @abbbi I think #974 might be needed for consistency around avoiding users needing to configure both options if they set storage_pool_name.

I do wonder though should we change the options around and instead of having:
* storage_pool_name => default base box image volume storage
* snapshot_pool_name => live VM writes

to use:
* storage_pool_name => live VM writes
* vmbox_pool_name => default base box image volume storage

As there might be some confusion as to the use of the word snapshot as there is the concept of snapshotting a VM that has been running, and it seems more intuitive that storage_pool_name should control where the machine uses rather than going forward it focuses on controlling where the base image is placed and instead to control where the machine writes to you need to do something else.

Thoughts?

abbbi:
hi, i think this one has been taken care of.

PR #953, opened 2018-12-04
#952 : support different storage pool for base box snapshot

abbbi:


electrofelix:
@pronix Seems like this is missing handling the snapshot storage pool similar to the main storage pool see https://github.com/vagrant-libvirt/vagrant-libvirt/blob/39f9a9d0b455a358917290417bb447396c245034/lib/vagrant-libvirt/action/handle_storage_pool.rb, should be included before this gets released?

electrofelix:
@abbbi can you give some consideration to the naming used here, would you be OK if I inverted the behaviour so that storage_pool_name was retained as the default pool that writes for vm's occurred and instead switched to using an option to allow the base box to be located in a separate pool that can be common?

PR #954, opened 2018-12-08
Adding disk_address_type option.

bdegreve:
Hi,

This patch allows to specify the virtio-mmio address type, which is needed
for debian guests on virt machines that don't support the PCI address type.

See also [here](https://translatedcode.wordpress.com/2016/11/03/installing-debian-on-qemus-32-bit-arm-virt-board/) where it is explicitly said to use `virtio-blk-device` and `virtio-net-device` instead of `virtio-blk-pci` and `virtio-net-pci`, for that reason.

Apparently, libvirt will use the `virtio-blk-pci` and `virtio-net-pci` by default. By setting address type to `virtio-mmio`, it uses `virtio-blk-device` instead.  It seems not necessary to do the same for the network controller, since libvirt will also use `virtio-net-device` if the disk address type is set to `virtio-mmio`

PS: this is my first dabbling with ruby :-)

Relates-to: #1608


lib/vagrant-libvirt/templates/domain.xml.erb
```diff
@@ -103,6 +103,9 @@
       <source file='<%= @domain_volume_path %>'/>
       <%# we need to ensure a unique target dev -%>
       <target dev='<%= @disk_device %>' bus='<%= @disk_bus %>'/>
+      <% if @disk_address_type %>
+      <address type='<%= @disk_address_type %>'/>
+      <% end %>
```
electrofelix: Can you add the corresponding block for the additional disks?

In the next block should be able to add:
```
<% if d[:address_type] %>
<address type='<%= d[:address_type] %'/>
<% end %>
```
Likely will need an update to the finalize method to ensure that as the disks are finalized, the default for `address_type` is taken from `disk_address_type` within the `disks.map!` block.

electrofelix (dismissed):

lib/vagrant-libvirt/templates/domain.xml.erb
bdegreve: I've added the extra block as requested, but I don't know how to modify the finalize method to do what you ask (i'm not familiar with Ruby). However, it seems that libvirt already uses the same address type as the primary disk by default. At leasts, that's what my experiments seem to indicate.

One question though, when I added an extra disk in my Vagrantfile, this extra disk was attached to /dev/vda, and the primary disk to /dev/vdb. Even if I've set disk_device='vda' for the primary disk. Is this normal behaviour?

bdegreve:
About the failing CI build, is there anything I can/need to do?

```
Installing nokogiri 1.10.1 with native extensions
Gem::RuntimeRequirementNotMetError: nokogiri requires Ruby version >= 2.3.0. The
current ruby version is 2.2.0.
An error occurred while installing nokogiri (1.10.1), and Bundler
cannot continue.
```

electrofelix:
@bdegreve if you rebase against the latest master, you can pick up the fixes for Travis CI

lib/vagrant-libvirt/templates/domain.xml.erb
electrofelix: Sounds like it's current behaviour, but not desirable, the method _get_device is not including the primary disk settings when iterating over existing devices to assign letters, see https://github.com/vagrant-libvirt/vagrant-libvirt/blob/master/lib/vagrant-libvirt/config.rb#L289-L298

The settings from https://github.com/vagrant-libvirt/vagrant-libvirt/blob/c02905be14f95ddbc13d91ae39ff3061f7021fd5/lib/vagrant-libvirt/config.rb#L691-L692 should probably be injected at the point the rest of the disks are finalized https://github.com/vagrant-libvirt/vagrant-libvirt/blob/c02905be14f95ddbc13d91ae39ff3061f7021fd5/lib/vagrant-libvirt/config.rb#L724-L728
Which would ensure that devices are correct ordered after the primary.

Consequently when the XML is generated it's spotting that 'vda' has already been assigned to the other disks and changes the value to be the next available letter.

The other alternative would be to delay device assignment until after all configuration has merged, I need to go back and look closely at when `finalize!` is called relative to `merge` for handling the Vagrantfile from the box, the project, and user custom overrides from ~/.vagrant.d/Vagrantfile.

Super annoying that I can't comment on areas of the code that are not modified in order to be able to point to have a conversation around the exact lines of code involved.

As libvirt is defaulting to the same type as the primary disk, this should suffice for now, but definitely need to try and get some tests in around how disks are added and managed to ensure all this configuration works as expected.

bdegreve: You mean I shoud add something like this?
```
         @disks = [] if @disks == UNSET_VALUE
         @disks.map! do |disk|
           disk[:device] = _get_device(@disks) if disk[:device].nil?
+          disk[:address_type] = @disk_address_type if disk[:address_type].nil?
           disk
         end
 ```

Comment (bdegreve) [reply lib/vagrant-libvirt/templates/domain.xml.erb]: Hi @electrofelix, I'm not sure if you still expect me to make changes on this PR. I've added tests as requested. But I don't know if you want me to address the device ordering as well ...

I guess I could add this line to explicitly set the address type of the secondary disks to the primary one, but I'd like your opinion first:
```
         @disks = [] if @disks == UNSET_VALUE
         @disks.map! do |disk|
           disk[:device] = _get_device(@disks) if disk[:device].nil?
+          disk[:address_type] = @disk_address_type if disk[:address_type].nil?
           disk
         end
```

PR #955, opened 2018-12-10
Do not remove SSH hostkeys or update LVM UUIDs on package

zakame:
Fix `vagrant package` of e.g. [bento](https://github.com/chef/bento)-derived boxes that require SSH hostkeys already set (as in the case of bento/debian-7 and above,) as well as leave existing LVM UUIDs untouched (e.g. in bento/ubuntu-18.04.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ can help a lot :-)
 - [Customized Graphics](#customized-graphics)
 - [Box Format](#box-format)
 - [Create Box](#create-box)
+- [Package Box from VM](#package-box-from-vm)
 - [Development](#development)
 - [Contributing](#contributing)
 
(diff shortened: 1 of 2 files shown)


electrofelix:
This seems like it would break packaging of machines that are subsequently used for multi-machine environments in not generating different host keys, similarly I would suspect issues if the same LVM UUIDs are used across multi-vm vagrant environments.

Seems like this should be optional behaviour for where it is the right thing to do, opposed to being applied to all use cases.

zakame:
> This seems like it would break packaging of machines that are subsequently used for multi-machine environments in not generating different host keys, similarly I would suspect issues if the same LVM UUIDs are used across multi-vm vagrant environments.
> 
> Seems like this should be optional behaviour for where it is the right thing to do, opposed to being applied to all use cases.

Agreed, actually I'd love to see if we can pass virt-sysprep options through `vagrant package` or some other means rather than hard-coding in the first place.

electrofelix:
There are two PRs in this area, if you could look at some of the suggests I made on #873 on using an environment variable to allow a user to override the default sysprep options, will merge whichever implements that as a solution that should allow you to manage additional options.

zakame:
@electrofelix `VAGRANT_LIBVIRT_VIRT_SYSPREP_OPERATONS` env (with existing `defaults,-ssh-userdir` defaults) sounds like a good option for me; for my workflow (which is mainly to support Chef/Test-Kitchen,) I rarely have to `vagrant package` so I can just set this env in a script/alias.  I'll update this PR, thanks!

zakame:
@electrofelix thanks again, updated this one based on your comment https://github.com/vagrant-libvirt/vagrant-libvirt/pull/873/files#r256423392 and added a small doc/example.

zakame:
@electrofelix thanks! :tada:

PR #961, opened 2019-01-07
Add qemu commandline environment variable support.

ju2wheels:
- Also cleans up weird variable indirection used for qemu commandline args variable through `qargs` in various constructors.
- Addresses some functionality discussed in #776.


ju2wheels:
The integration is failing in setting up ruby for the tests for one of the rvm versions, not sure if I should just update the rvm version or not...

lib/vagrant-libvirt/config.rb
```diff
@@ -766,7 +782,11 @@ def finalize!
         # Attach mgmt network
         @mgmt_attach = true if @mgmt_attach == UNSET_VALUE
 
+        # Additional qemu commandline arguments
         @qemu_args = [] if @qemu_args == UNSET_VALUE
+
+        # Additional qemu commandline environment variables
+        @qemu_env = {} if @qemu_env == UNSET_VALUE
```
electrofelix: I'm pretty sure you need to add something to handle `qemu_args` and `qemu_env` to the `def merge` function in this file otherwise if a user was to specify some settings in ~/.vagrant.d/Vagrantfile for the libvirt provider to always set some `qemu_args/qemu_env` any settings to these in the Vagrantfile of the directory would overwrite rather than merge as the user would expect.

ju2wheels: Added the merge update and tested locally.

ju2wheels:
@electrofelix merge update added, let me know if there's anything else that needs changing.

Issue #572, opened 2016-03-07
rework IP detection in lib/vagrant-libvirt/action/prepare_nfs_settings.rb


jim-minter:
IMO, the logic for detection of guest and host IP addresses in prepare_nfs_settings.rb is pretty kludgey.  Some issues:

1. /etc/exports gets a line per guest IP, when it should probably only get one line with the guest's known IP on the management interface (on 192.168.121.0/24 by default).  Anything else is possibly/probably spurious.
1. If the guest has multiple interfaces with identical IP addresses, the NFS server gets confused (short term patch to solve this in #571).
1. Funky logic is used to pass the host's IP to the guest to mount /vagrant.  The host's IP on the management interface should be used.

Example weirdness that this causes:
* host IPs: eth0 192.168.0.1/24, VPN 10.1.2.3/8, virbr1 192.168.121.1/24
* guest IPs: eth0 192.168.121.100/24, tun0 10.0.0.1/8

With the current logic, you get /etc/exports entries for 192.168.121.100 and 10.0.0.1, and the guest tries to mount NFS on 10.1.2.3 - which for some reason is working for me, but it's odd.

Is it possible to remove read_host_ip and read_machine_ip and pull the required values correctly without using heuristics?

infernix:
This is also related to #544 but I do agree, it needs a rework.  I hope to free up some cycles to start to address this soon

miminar:
Addressing this would be much appreciated.

Issue #544, opened 2016-01-21
Wrong nfs host ip when using docker inside Vagrant


varac:
Hi, 

i'm using NFS shared folders support for vagrant (libvirt provider). It works well, until i try to use docker inside a vagrant box.

Mounting the shared folder fails:  

```
==> jessie: Mounting NFS shared folders...
 INFO ssh: SSH is ready!
 INFO ssh: Execute:  (sudo=false)
[...]

PR #964, opened 2019-01-17
prepare_nfs_settings.rb: check for 'ssh' service to see if guest network interface is up

pxsloot:
try to open a tcp socket to 'ssh' service to determine whether guest management interface is up.

The 'ping' function tries to open a tcp socket with the 'echo' service on the guest. When it fails, it will try the next guest ip address.

The 'vagrant management interface' will have the tcp service 'ssh' available
(you wouldn't be able to vagrant ssh into your vagrant box otherwise), so it makes
sense to use the ssh tcp port for the ping.

the 'echo' service nowadays is often not available out-of-the-box, causing the 'ping' of the guest management interface to fail. When the guest has docker installed, or other network interfaces added, this results in the wrong ip addresses being used in /etc/exports, and
the vm not being able to mount the vagrant shares at 'vagrant reload', or 'vagrant halt; vagrant up' time.

This fixes the vm not being able to mount the shares over nfs after
a reload due to the wrong ip addresses being used.

related issues #544 #572

### DIFF
--- a/lib/vagrant-libvirt/action/prepare_nfs_settings.rb
+++ b/lib/vagrant-libvirt/action/prepare_nfs_settings.rb
@@ -79,7 +79,7 @@ module VagrantPlugins
         # Check if we can open a connection to the host
         def ping(host, timeout = 3)
           ::Timeout.timeout(timeout) do
-            s = TCPSocket.new(host, 'echo')
+            s = TCPSocket.new(host, 'ssh')
             s.close
           end
           true


pxsloot:
AFAIK the failed tests have nothing to do with the change in the PR.

electrofelix:
@pxsloot indeed, looks like it's caused by a rubygems -> ruby incompatibility. We'll need to fix up the testing code. However as the second job passed, I think we can merge this and sort that problem out separately.

PR #969, opened 2019-02-06
Enable QEMU Session by default

pvalena:
sets default for `@qemu_use_session` to true.

_ _ _ _

We in Fedora want to enable it on default Vagrant-Libvirt installation (and we [intend to](https://fedoraproject.org/wiki/Changes/Vagrant_2.2_with_QEMU_Session), soon).

Are there any obstacles, apart from the unability to use `private_network`?
Basic functionality works out of the box for us.

Is there plan to make it default in future vagrant-libvirt releases? If not, why?

### DIFF
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -656,7 +656,7 @@ module VagrantPlugins
         @management_network_pci_slot = nil if @management_network_pci_slot == UNSET_VALUE
         @system_uri      = 'qemu:///system' if @system_uri == UNSET_VALUE
 
-        @qemu_use_session = false if @qemu_use_session == UNSET_VALUE
+        @qemu_use_session = true if @qemu_use_session == UNSET_VALUE
 
         # generate a URI if none is supplied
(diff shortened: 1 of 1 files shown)


electrofelix:
I think this would require a major version bump because it may mean that many existing Vagrantfiles would suddenly break for other distros. Some reluctance here is probably just down to limited acceptance tests being in place to validate behaviour.

voxik:
I think the networking will be the main issue (see [[1], [2]] for example). So is there a chance to detect that networking setup is modified and somehow report this and suggest the "system" session for such Vagrant files?


[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1701755
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1697773

electrofelix:
Even if there is I think #997 is a better approach to solving this, if upstream don't think that adding support for loading a system based configuration file makes sense - see https://groups.google.com/d/msg/vagrant-up/MK77m8WNU4A/lhVV5tECCAAJ

PR #971, opened 2019-02-12
Updated Libvirt-based Vagrant box search reference and edited names for consistency

crimsun:
The Atlas reference in the root README no longer resolves, so it has been updated to the closest matching (in spirit).

Also, some very minor capitalization and consistency updates were applied.


electrofelix:
Thought that GitHub should have updated the refs/pulls/971/merge based on my update to fix the tests, but seems not. Pretty sure this is good to land though as there is no impact to tests.

PR #973, opened 2019-02-13
Update travis conf and test multiple combinations

electrofelix:
Attempt to use latest travis CI settings for installing required
packages.

Should attempt to test various combinations to try and catch problems
occurring across different versions of vagrant with different versions
of ruby when people are using distro provided versions.


electrofelix:
trigger retest

PR #974, opened 2019-02-13
Default to storage pool if snapshot pool not defined

electrofelix:
Default to using the same storage pool in case the user defines
storage_pool_name without defining snapshot_pool_name and only use a
separate pool when snapshot_pool_name is explicitly set.

This will ensure that setting storage_pool_name to a different pool does
not also require setting snapshot_pool_name to match previous behaviour.

Additionally it is reasonable for a user to assume that if a Vagrantfile
sets storage_pool_name to something other than 'default', then there
should be no need for a default pool.


electrofelix:
@pronix I believe this should ensure that the next release doesn't accidentally introduce a regression for anyone configuring `storage_pool_name` to something other than default as they might be surprised about the snapshot suddenly being placed back under the default pool with #953.

PR #976, opened 2019-02-19
Update coveralls to use travis-pro

electrofelix:
lemurheavy/coveralls-public#1119 states it has switched from travis-ci
to travis-pro.

Change-Id: Ifa4fc465eca82f33ef288ffc26c37a566b313530


electrofelix:
Shouldn't be needed it appears

Issue #986, opened 2019-03-22
storage :file fail with QEMU session


dkrm0:
### Steps to reproduce
1. Use QEMU session with `libvirt.qemu_use_session = true`
2. Create and attach an additional disks via `libvirt.storage :file`
3. Vagrant will crash, trying to create the domain

### Expected behaviour
The domain should be created with an additionnal disk.

### Actual behaviour
Vagrant crash with this backtrace:
```
/home/dkrm/vms/vagrant/gems/2.6.2/gems/vagrant-libvirt-0.0.45/lib/vagrant-libvirt/action/create_domain.rb:156:in `block in call': undefined method `+' for nil:NilClass (NoMethodError)
	from /home/dkrm/vms/vagrant/gems/2.6.2/gems/vagrant-libvirt-0.0.45/lib/vagrant-libvirt/action/create_domain.rb:147:in `each'
	from /home/dkrm/vms/vagrant/gems/2.6.2/gems/vagrant-libvirt-0.0.45/lib/vagrant-libvirt/action/create_domain.rb:147:in `call'
	from /opt/vagrant/embedded/gems/gems/vagrant-2.2.4/lib/vagrant/action/warden.rb:50:in `call'
	from /opt/vagrant/embedded/gems/gems/vagrant-2.2.4/lib/vagrant/action/builtin/before_trigger.rb:23:in `call'
	from /opt/vagrant/embedded/gems/gems/vagrant-2.2.4/lib/vagrant/action/warden.rb:50:in `call'
	from /opt/vagrant/embedded/gems/gems/vagrant-2.2.4/lib/vagrant/action/builtin/after_trigger.rb:26:in `call'
	from /opt/vagrant/embedded/gems/gems/vagrant-2.2.4/lib/vagrant/action/warden.rb:50:in `call'
	from /home/dkrm/vms/vagrant/gems/2.6.2/gems/vagrant-libvirt-0.0.45/lib/vagrant-libvirt/action/create_domain_volume.rb:83:in `call'
	from /opt/vagrant/embedded/gems/gems/vagrant-2.2.4/lib/vagrant/action/warden.rb:50:in `call'
	from /opt/vagrant/embedded/gems/gems/vagrant-2.2.4/lib/vagrant/action/builtin/before_trigger.rb:23:in `call'
	from /opt/vagrant/embedded/gems/gems/vagrant-2.2.4/lib/vagrant/action/warden.rb:50:in `call'
[...]

PR #997, opened 2019-04-10
Add environment variable for qemu_use_session

jdoss:
This PR adds in support for new environment variable `VAGRANT_LIBVIRT_QEMU_USE_SESSION` which allows a user to globally overwrite the `qemu_use_session` variable if it is not set within the Vagrantfile. 

It was inspired by this BZ comment https://bugzilla.redhat.com/show_bug.cgi?id=1697773#c3 which a user encountered a problem when the Fedora maintainers chose to set  qemu_use_session to true in by default.  Having this environment variable will allow users to not have to add `vm.qemu_use_session = false` their Vagrantfiles if distributions chose to change upstreams default setting.

Also, having this setting might allow for an easier transition for setting `vm.qemu_use_session = true` b default in https://github.com/vagrant-libvirt/vagrant-libvirt/pull/969.

I also added a bit to the error message from `network_not_available_error` to help make it more clear on what kind of issue a user might be having if they set `vm.qemu_use_session = true` or it is set to true by default by different distributions. 

Let me know if there needs to be anything changed.

### DIFF
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -659,7 +659,13 @@ module VagrantPlugins
         @management_network_domain = nil if @management_network_domain == UNSET_VALUE
         @system_uri      = 'qemu:///system' if @system_uri == UNSET_VALUE
 
-        @qemu_use_session = false if @qemu_use_session == UNSET_VALUE
+        if ENV['VAGRANT_LIBVIRT_QEMU_USE_SESSION'] == 'true'
+          @qemu_use_session = true if @qemu_use_session == UNSET_VALUE
+        elsif ENV['VAGRANT_LIBVIRT_QEMU_USE_SESSION'] == 'false'
+          @qemu_use_session = false if @qemu_use_session == UNSET_VALUE
+        else
+          @qemu_use_session = false if @qemu_use_session == UNSET_VALUE
+        end
 
         # generate a URI if none is supplied
         @uri = _generate_uri if @uri == UNSET_VALUE
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -659,12 +659,8 @@ module VagrantPlugins
         @management_network_domain = nil if @management_network_domain == UNSET_VALUE
         @system_uri      = 'qemu:///system' if @system_uri == UNSET_VALUE
 
-        if ENV['VAGRANT_LIBVIRT_QEMU_USE_SESSION'] == 'true'
-          @qemu_use_session = true if @qemu_use_session == UNSET_VALUE
-        elsif ENV['VAGRANT_LIBVIRT_QEMU_USE_SESSION'] == 'false'
-          @qemu_use_session = false if @qemu_use_session == UNSET_VALUE
-        else
-          @qemu_use_session = false if @qemu_use_session == UNSET_VALUE
+        if @qemu_use_session == UNSET_VALUE
+          @qemu_use_session = ENV['VAGRANT_LIBVIRT_QEMU_USE_SESSION'] == 'true'
         end
 
         # generate a URI if none is supplied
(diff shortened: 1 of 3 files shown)


electrofelix:
Are you aware that you can control this behaviour by defining a `${HOME}/.vagrant.d/Vagrantfile` with the following contents:
```
    libvirt.qemu_use_session = true
  end
end
```

jdoss:
I was not. I assume you feel this is the better route to go than using an environment var? If so I can close this PR and reopen another one with the error message changes if you feel those are acceptable.

voxik:
> Are you aware that you can control this behaviour by defining a `${HOME}/.vagrant.d/Vagrantfile`

I mentioned this possibility in the Fedora ticket, but the problems is, that this is just per user settings, which we cannot change globally, therefore for Fedora, we have to patch the default settings.

I am not saying that we will switch to env variable, but it might be handy.

jdoss:
I still feel that having the env var would be a handy thing for all distros to use since `qemu_use_session` seems to be pretty disruptive to end users and giving them more knobs to turn to fit their specific use case isn't a bad thing.

lib/vagrant-libvirt/config.rb
```diff
@@ -659,7 +659,13 @@ def finalize!
         @management_network_domain = nil if @management_network_domain == UNSET_VALUE
         @system_uri      = 'qemu:///system' if @system_uri == UNSET_VALUE
 
-        @qemu_use_session = false if @qemu_use_session == UNSET_VALUE
+        if ENV['VAGRANT_LIBVIRT_QEMU_USE_SESSION'] == 'true'
```
bschmeck: You could write this as:
```
if @qemu_use_session == UNSET_VALUE
  @qemu_use_session = ENV['VAGRANT_LIBVIRT_QEMU_USE_SESSION'] == 'true'
end
```

jdoss: Thanks! Made these changes in https://github.com/vagrant-libvirt/vagrant-libvirt/pull/997/commits/fc62fd27c93ff554a608f52e4d89ee602f07ec22 :man_dancing:

electrofelix:
Going to try and have a conversation with upstream before merging this and going down this route as I think this PR hints at the need for a distro supplied configuration file containing defaults to be possible to provide rather than supporting this on a per plugin per setting basis.

May well need to land this in the interim to facilitate, like to make sure that this is really the only practical way rather than considering making it possible to have an env variable tell vagrant to read from /etc/vagrant/Vagrantfile first to provide some default settings as a way of solving the more general problem and avoid the need to request this for each setting or for multiple plugins.

voxik:
> Going to try and have a conversation with upstream before merging this and going down this route as I think this PR hints at the need for a distro supplied configuration file containing defaults to be possible to provide rather than supporting this on a per plugin per setting basis.

In this context, we have a way to utilize the env variable, because in Fedora, we provide our own `vagrant` executable \[[1]\], which is used to set some of the env variables (e.g. preferred providers \[[2]\]). So it could be used to override the session settings instead of patching defaults.

[1]: https://src.fedoraproject.org/rpms/vagrant/blob/master/f/binstub

electrofelix:
@voxik that might be useful for you in this context, however I'm not liking the path of each time something needs to be overridden that we have to create another environment variable for a setting within this project.

dustymabe:
hey @jdoss @electrofelix - Another way to solve this problem is over in https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1062. It re-uses the existing libvirt tools mechanism for specifying how to connect to libvirt, which includes whether it's session or system. On my laptop i've been using `LIBVIRT_DEFAULT_URI=qemu:///system` for a long time for `virsh` and I'd like to keep using that for `vagrant`.

If we like that approach I can add similar documentation changes to that PR.

cjsin:
bug #986 seems related, currently users must specify both libvirt.uri and libvirt.qemu_use_session in a consistent manner in order to successfully create secondary hard disks

electrofelix:
Support has been merged to enable use of LIBVIRT_DEFAULT_URI and will automatically set `@qemu_use_session`, is that sufficient to solve the original problem this PR was intended to tackle?

jdoss:
Sounds good to me @electrofelix. I'll close this PR.

PR #1000, opened 2019-04-26
Add disk driver options with minor refactor

### LINKED ISSUE #998: add support for disk I/O options
Hello
please add support for other disk performance options like **I/O**, **Discard mode**, **Detect zeroes**. I see you already have option for **cache** so this would be very useful to have also.
Thank you

davidscaife:
Resolves #998 

I've had a crack at adding some of the disk driver options: `io`, `copy_on_read`, `discard` and `detect_zeroes` for both the primary vm disk and additional disks.

Since the `volume_cache` option already existed, initially I copied this approach... but that resulted in a mess of properties like `volume_io`, `volume_copy_on_write`, `volume_discard` etc which was a bit nasty. Therefore, the "minor refactor" part of this PR was to consolidate all the primary disk driver properties into a single call (`disk_driver`).

The following Vagrantfile shows all the functionality added:

```ruby
  config.vm.box = "centos/7"

    libvirt.cpus = 2 

    # deprecated:
    # libvirt.volume_cache = 'none'

    # primary disk options (all optional)
    libvirt.disk_driver \
        :cache => 'none',
        :io => 'native',
        :copy_on_read => 'on',
        :discard => 'unmap',
        :detect_zeroes => 'on'

    # additional disk
    libvirt.storage :file, :size => '20G', :type => 'raw',
        :cache => 'none',
        :io => 'threads',
        :copy_on_read => 'on',
        :discard => 'unmap',
        :detect_zeroes => 'on'
  end 
end
```

I've kept the `volume_cache` property in order to maintain backwards compatibility with existing Vagrantfiles. It is only used if `disk_driver` is not present (even if `:cache` is not set - in that case, the hypervisor default is always used).

I've marked this as a WIP even though it's functionally complete as I had a couple of "best practice"-style questions:

1. [Snip - ERB formatting - simplified in f6c93d2]

2. Having replaced/deprecated the `volume_cache` property (again noting that it still works as-is for any existing Vagrantfiles that may use it), what would be the best way to notify users that it's been replaced? Something in the `create_domain.rb` ui output? Mention it in the README?


davidscaife:
Ugh, well, I'm a doofus and having thought about it for a few seconds the .erb file can surely be trivially simplified. Give me a sec...

lib/vagrant-libvirt/templates/domain.xml.erb
```diff
@@ -108,7 +118,12 @@
 <%# additional disks -%>
 <% @disks.each do |d| -%>
     <disk type='file' device='disk'>
-      <driver name='qemu' type='<%= d[:type] %>' cache='<%= d[:cache] %>'/>
+      <driver name='qemu' type='<%= d[:type] %>' cache='<%= d[:cache] %>'<%= -%>
+<%# Extra optional disk driver atrributes -%>
+<%=      " io='#{d[:io]}'"                           unless d[:io].nil? -%>
+<%=      " copy_on_read='#{d[:copy_on_read]}'"       unless d[:copy_on_read].nil? -%>
+<%=      " discard='#{d[:discard]}'"                 unless d[:discard].nil? -%>
+<%=      " detect_zeroes='#{d[:detect_zeroes]}'"     unless d[:detect_zeroes].nil? -%>/>
```
electrofelix: Similarly is it worth just using a loop here with a defined list of params to read from?

```
<%
       keys = [:type, :cache, :io, :copy_on_read, :discard, :detect_zeros]
       @opts = d.select |key, value| { @keys.include? key && !value.nil? }.each |key, value|
            "#{key.to_str}=#{value}"
       end.join(' ')
-%>
      <driver name='qemu' <%= @opts -%> />
```

lib/vagrant-libvirt/config.rb
```diff
@@ -523,6 +525,17 @@ def smartcard(options = {})
...
+      # Disk driver options for primary disk
+      def disk_driver(options = {})
+        @disk_driver_opts = {
+          cache: options[:cache],
+          io: options[:io],
+          copy_on_read: options[:copy_on_read],
+          discard: options[:discard],
+          detect_zeroes: options[:detect_zeroes]
+        }
+      end
```
electrofelix: I think this may need some code added to the `merge` method to ensure that if someone was to define some of these settings in a `${HOME}/.vagrant.d/Vagrantfile` they would be merged with the disk_driver settings.

Though that would result in getting merged results which may cause some problems, but as you can set values to 'nil' explicitly, I think the benefits of allowing someone to provide settings in other Vagrantfiles to be used as defaults is worth it.

lib/vagrant-libvirt/config.rb
```diff
@@ -88,9 +88,10 @@ class Config < Vagrant.plugin('2', :config)
       attr_accessor :machine_virtual_size
       attr_accessor :disk_bus
       attr_accessor :disk_device
+      attr_accessor :disk_driver_opts
       attr_accessor :nic_model_type
       attr_accessor :nested
-      attr_accessor :volume_cache
+      attr_accessor :volume_cache # deprecated, kept for backwards compatibility; use disk_driver
```
electrofelix: I'm thinking there must be a standard pattern to make writing to this trigger a deprecated notice.

My guess would be to simply try defining a writer method further down
```
  def volume_cache=(value)
    # emit a deprecation message here
    @volume_cache = value
  end
```

May need to switch the `attr_accessor` to `attr_reader`, not sure whether it will work to override or get an error about having a writer method defined at the same time as `attr_accessor`.

Or is it better done via the validate method? Not sure if that supports the idea of warnings in addition to errors such as the following

https://github.com/hashicorp/vagrant/blob/master/lib/vagrant/config/v2/loader.rb#L111-L137

lib/vagrant-libvirt/templates/domain.xml.erb
```diff
@@ -118,12 +112,12 @@
...
-<%=      " io='#{d[:io]}'"                           unless d[:io].nil? -%>
-<%=      " copy_on_read='#{d[:copy_on_read]}'"       unless d[:copy_on_read].nil? -%>
-<%=      " discard='#{d[:discard]}'"                 unless d[:discard].nil? -%>
-<%=      " detect_zeroes='#{d[:detect_zeroes]}'"     unless d[:detect_zeroes].nil? -%>/>
+      <driver name='qemu' type='<%= d[:type] %>' <%=
+        d.select { |k,_| [:cache, :io, :copy_on_read, :discard, :detect_zeroes].include? k }
+         .reject { |k,v| v.nil? }
+         .map { |k,v| "#{k}='#{v}'"}
+         .join(' ')
+      -%>/>
```
electrofelix: This works for me, ignore my review about this on the previous commit

lib/vagrant-libvirt/templates/domain.xml.erb
```diff
@@ -99,17 +99,11 @@
...
-<%=      " detect_zeroes='#{@disk_driver_opts[:detect_zeroes]}'" unless @disk_driver_opts[:detect_zeroes].nil? -%>
-<% else -%>
-<%# For backwards-compatibility with deprecated 'volume_cache' property -%>
-<%=      " cache='#{@domain_volume_cache}'" -%>
-<% end -%>/>
+      <driver name='qemu' type='qcow2' <%=
+        @disk_driver_opts.empty? ? "cache='#{@domain_volume_cache}'" :
+            @disk_driver_opts.reject { |k,v| v.nil? }
+                             .map { |k,v| "#{k}='#{v}'"}
+                             .join(' ') -%>/>
```
electrofelix: This works for me, ignore my review about this on the previous commit

lib/vagrant-libvirt/config.rb
davidscaife: After a bit of tinkering, it seemed that `validate` looked like a sensible place to put it; it generally gets called once, and the vagrant environment is available to properly show a pretty colourised warning message. It looks like the `validate` return value is only for errors, but using the ui.warn method seems to work nicely.

lib/vagrant-libvirt/config.rb
davidscaife: Good call. I've added a commit to handle merging and added a note about setting the values to `nil` in the readme.

davidscaife:
^ Force-push - just made the deprecation warning use `machine.provider_config.volume_cache` instead of `@volume_cache` for consistency.

tyll:
For `discard='unmap'` to work the disk needs to be on a SCSI bus that uses VirtIO. Also I wonder if it makes sense to enable this by default. When backed by SSDs, it should not have a significant negative perfornance impact but improve disk usage.

lib/vagrant-libvirt/config.rb
electrofelix: That looks like it'll work nicely enough

davidscaife:
I was just about to add some notes in the README about `:discard` requiring the SCSI bus given the comments above, but thought I'd just tinker with it myself so I know what's going on, but...

@tyll Perhaps I've misunderstood, or maybe there've been improvements in the libvirt/qemu/kvm space - I've just tried this on some disks both with `bus="virtio"` and even `bus="ide"`, with qcow and raw images, not a SCSI controller in sight... and each configuration I tried behaved well with the unmap options, shrinking the disk files as I ran `fstrim` in the VM. Perhaps I've missed something but I wasn't expecting that based on your comments?

*...does some searching...*

...and now that I've written that, I've just had a quick look, and yes it seems that discard support is a rather recent addition for virtio-blk, added in the guest side circa Linux 5.0 and implemented in qemu in version 4.0. Good news, but probably still worthy of a note in the README, since obviously not everybody runs that bleeding-edge.

I'll triple-check those version numbers/changelogs and add some words tomorrow.

joso9001:
I'm very interested in this one!

Is there anything I could do to help it along?

JockeTF:
I'm very interested in this one!

Is there anything I could do to help it along?

PR #1005, opened 2019-05-09
Use parallel gzip in box creation if available

vpodzime:
This can speed up box creation significantly on modern machines.

### DIFF
--- a/tools/create_box.sh
+++ b/tools/create_box.sh
@@ -118,7 +118,12 @@ EOF
 
 echo "==> Creating box, tarring and gzipping"
 
-tar cvzf "$BOX" -S --totals ./metadata.json ./Vagrantfile ./box.img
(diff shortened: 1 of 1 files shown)


tools/create_box.sh
```diff
@@ -118,7 +118,12 @@ EOF
 
 echo "==> Creating box, tarring and gzipping"
 
-tar cvzf "$BOX" -S --totals ./metadata.json ./Vagrantfile ./box.img
+if type pigz >/dev/null 2>/dev/null; then
+  GZ="pigz"
+else
+  GZ="gzip"
+fi
+tar cv -S --totals ./metadata.json ./Vagrantfile ./box.img | $GZ -c > "$BOX"
```
electrofelix: Could probably have used `-I $GZ` arg to tar but this is good enough.

Issue #1011, opened 2019-06-11
Why not forward ssh-port?


Blaimi:
the port-forwarding for the ssh-port is skipped on purpose
https://github.com/vagrant-libvirt/vagrant-libvirt/blame/0f037bab9b8dba3e8c260986c1858348c65cb161/lib/vagrant-libvirt/action/forward_ports.rb#L75

we are heavily using this port-forwarding in our configurations of IDEs to auto-upload, browse files, tunneling to databases, …

at the moment we are using this workaround:

```ruby
	config.vm.provider "virtualbox" do |vb, override|
		vb.memory = "2048"
		vb.cpus = 2
		override.vm.network "forwarded_port", guest: 22, host: 2202, id: "ssh"
	end
	config.vm.provider :libvirt do |libvirt, override|
		libvirt.cpus = 2
		libvirt.memory = "2048"
		libvirt.volume_cache = "unsafe"
		# this is necessary because libvirt-provider skips any forwarding with id:'ssh'
		override.vm.network "forwarded_port", guest: 22, host: 2202, id: "ssh_lh"
	end
```

it would be better if there is a documented option to enable or disable the forwarding. To prevent backwards-compatibility-breaks, this option can default to "not forward"

```ruby
	config.vm.network "forwarded_port", guest: 22, host: 2202, id: "ssh"
	config.vm.provider :libvirt do |libvirt|
		libvirt.forward_ssh_port = true
	end
```

At least a hint in the documentation woud be nice to prevent searching for a configuration-bug.

### Steps to reproduce
1. create a vagrant-box with libvirt-provider
2. configure tools to connect with ssh with persistent config but independend of vagrant or static ip like an IDE

### Expected behaviour
ssh-connection to localhost:2222 or any other configured port should work

### Actual behaviour
ssh-port is not forwarded to localhost

### System configuration
**OS/Distro version:**: debian “buster” testing

**Libvirt version:** 5.0.0

**Output of `vagrant version; vagrant plugin list`:** Installed Version: 2.2.3 vagrant-libvirt (0.0.45, system)

axmetishe:
I'm just curious due to some circumstances you forced to define the 'id'?
`node.vm.network :forwarded_port, guest: remote, host: local`
this construction will work out of the box for libvirt

Blaimi:
[...]

PR #1012, opened 2019-06-11
WIP: option to forward ssh-port to host

Blaimi:
Closes: #1011

needs testing!


lib/vagrant-libvirt/action/forward_ports.rb
```diff
@@ -72,7 +72,7 @@ def compile_forwarded_ports(config)
           config.vm.networks.each do |type, options|
             next if options[:disabled]
 
-            next unless type == :forwarded_port && options[:id] != 'ssh'
+            next unless type == :forwarded_port && !@env[:machine].provider_config.forward_ssh_port && options[:id] != 'ssh'
```
electrofelix, line 75: Looking at the very original patch for the forwarded ports I think the expectation was the skipping forwarding the default ssh port should be removed at some point
https://github.com/vagrant-libvirt/vagrant-libvirt/pull/139/commits/bb5d0817894b54f8ecabe1fb9a8496f13b4adaf8#diff-a04ee612a672226ed5bdbe2817ec5bb9de835ca845afe54aaa4730473b3250eeR63

So I think best to invert this config option and make it possible to use to disable forwarding this if it causes problems, but otherwise allow the default forwarding of the ssh port on 2022 to exist.

electrofelix:
#1281 should make this a bit easier to adjust and be confident of the resulting behaviour.

electrofelix:
As I couldn't modify this PR, pushed up a replacement in #1284

PR #1020, opened 2019-07-03
fixes #1019: vagrant package fails with qemu >= 2.10

### LINKED ISSUE #1019: vagrant package fails with qemu >= 2.10
Vagrant package fails if the disk to package has a backing image and  the installed qemu version is greater than 2.10 while the virtual machine is still running.

The command fails with the following output:
```
qemu-img: Could not open '/var/lib/libvirt/images/[...].img': Failed to get shared "write" lock
```

This openstack-nova issue lead me to the fix of the issue:
https://bugzilla.redhat.com/show_bug.cgi?id=1654804#c4

**To fix this issue you need to change the following line:**
Change
https://github.com/vagrant-libvirt/vagrant-libvirt/blob/3f30606021444ac922c398e1737e09be4fe8fcfd/lib/vagrant-libvirt/action/package_domain.rb#L30

to:
```
backing = `qemu-img info --force-share "#{root_disk.path}" | grep 'backing file:' | cut -d ':' -f2`.chomp
```

Referring to the man page of `qemu-img` the `--force-share` option does the following:
```
If specified, qemu-img will open the image in shared mode, allowing other QEMU processes to open it in write mode. For example, this can be used to get the image information (with 'info' subcommand) when the image is used by a running guest.  Note that this could produce inconsistent results because of concurrent metadata changes, etc. This option is only allowed when opening images in read-only mode.
```

### Steps to reproduce
1. Install vagrant-libvirt on qemu >= 2.10 environment
2. Run and customize any vagrant box via libvirt provider
[Do not shutdown the virtual machine!]
3. Package the customized box via `vagrant package --output [...].box`

### Expected behaviour
The box should be packaged including copying and rebasing backing image

### Actual behaviour
The package command fails with:
```
qemu-img: Could not open '/var/lib/libvirt/images/[...].img': Failed to get shared "write" lock
```

### System configuration
**OS/Distro version:**: RHEL 8

**Libvirt version:** 4.5.0

**qemu-img version:** 2.12.0 (qemu-kvm-2.12.0-63.module+el8+2833+c7d6d092)

**Output of `vagrant version; vagrant plugin list`:** 
Installed Version: 2.2.4
vagrant-libvirt (0.0.45, global)

**Are you using upstream vagrant package or your distros package?**
Upstream

jonasjoest:
TODO: do a version check of qemu. So only provide this command line option if qemu >= 2.10

### DIFF
--- a/lib/vagrant-libvirt/action/package_domain.rb
+++ b/lib/vagrant-libvirt/action/package_domain.rb
@@ -27,7 +27,7 @@ module VagrantPlugins
           @tmp_img = @tmp_dir + '/box.img'
           Dir.mkdir(@tmp_dir)
           if File.readable?(root_disk.path)
-            backing = `qemu-img info "#{root_disk.path}" | grep 'backing file:' | cut -d ':' -f2`.chomp
+            backing = `qemu-img info --force-share "#{root_disk.path}" | grep 'backing file:' | cut -d ':' -f2`.chomp
           else
             env[:ui].error("Require set read access to #{root_disk.path}. sudo chmod a+r #{root_disk.path}")
             FileUtils.rm_rf(@tmp_dir)


electrofelix:
I'm guessing the switch to using download the image via the libvirt API means this option is no longer required. If it still is, will need to switch to the option passed via the API call so that it can support remote machines as well.

Issue #914, opened 2018-08-20
impossible to create private_network static IP with only an IPv6


lelutin:
Hi,

vagrant-libvirt offers IPv6 support for private addresses, as shown in the README. However, I find that it's impossible to setup a network interface of type `:private_network` with only an IPv6 address.

I've tried to give an IPv6 address to `:ip` as shown in the [vagrant private_network examples](https://www.vagrantup.com/docs/networking/private_network.html#ipv6), but this shows a ruby stack trace that says the ipaddress is invalid.

dropping `:ip` and using only the `:libvirt__ipv6*` parameters errors out in another way since not having the `:ip` parameter means requesting dhcp.

is the provider not accepting an IPv6 address in the `:ip` parameter for `config.vm.network` because of a constraint in how libvirt operates? or do you think that it would be possible to make some changes to setup an IPv6-only private_network address?

mykaul:
I've had a similar problem with mixed IPv4 and IPv6 - I suspect it has to do with:
1. What libvirt does and what Vagrant does. I suspect static IP allocation is done with IPv4 via Vagrant - unsure what does it with IPv6?
2. It seems to recreate the network, so the last 'creation' wins.

Here's a snippet of my code:
```
    (1..node_count).reverse_each do |num|
      config.vm.define "node-#{num}" do |node|
        vm_ip = "192.168.250.#{num+10}"
        vm_ip6 = "fddd::#{num+10}"

        node.vm.box = "centos/7"
        node.vm.network :private_network,
            :ip => vm_ip,
            :libvirt__driver_queues => "#{cpus}",
            :libvirt__guest_ipv6 => "yes",
            :libvirt__ipv6_address => vm_ip6,
            :libvirt__ipv6_prefix => "64"
```

The resulting network, in libvirt:
```
<network connections='3' ipv6='yes'>
  <name>vg0</name>
  <uuid>d6ee89af-4df3-46ee-b265-0ee8b3e596db</uuid>
  <forward mode='nat'>
    <nat>
      <port start='1024' end='65535'/>
    </nat>
  </forward>
  <bridge name='virbr2' stp='on' delay='0'/>
  <mac address='52:54:00:62:ef:2d'/>
  <ip address='192.168.250.1' netmask='255.255.255.0'>
    <dhcp>
      <range start='192.168.250.1' end='192.168.250.254'/>
    </dhcp>
  </ip>
  <ip family='ipv6' address='fddd::13' prefix='64'>
[...]

PR #1021, opened 2019-07-04
Add ipv6 only network support

agowa338:
Add ipv6 only network support

Fixes: #914

### DIFF
--- a/lib/vagrant-libvirt/action/create_network_interfaces.rb
+++ b/lib/vagrant-libvirt/action/create_network_interfaces.rb
@@ -216,6 +216,9 @@ module VagrantPlugins
               # do not run configure_networks for tcp tunnel interfaces
               next if options.fetch(:tunnel_type, nil)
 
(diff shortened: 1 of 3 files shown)


lib/vagrant-libvirt/action/create_network_interfaces.rb
```diff
@@ -298,6 +301,16 @@ def interface_network(libvirt_client, options)
             end
           end
 
+          if options[:ipv6_address]
+            address = options[:ipv6_address]
+            available_networks.each do |network|
+              if options[:ipv6_address] == network[:ip_address]
```
ekohl: This looks odd. It means the ipv6_address as specified should be the one of the host where the IPv4 address should be the one of the VM. I think it would be better to look if the `ipv6_address` is within the IPv6 network as is done in IPv4.

rubenk:
This should be fixed now by #1436

electrofelix:
Going to close this as understood to be fixed by #1436.

PR #1029, opened 2019-07-17
package_domain.rb: Fix getting virt-sysprep operations envvar

zakame:
Follow up to #955.

Use `ENV.[]` for getting the environment variable instead of the
non-standard `ENV.get`.


zakame:
Obsoleted by #1035

PR #1036, opened 2019-08-18
Add public_address provider capability

zakame:
Used by [vagrant-share plugin](https://www.vagrantup.com/docs/share/) to determine local guest address for forwarding via [ngrok](https://ngrok.com).

Here is an example that exposes HTTP in the guest:

```
~/d9 $ cat Vagrantfile
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|
  config.vm.box = "bento/debian-9"

  config.vm.network "forwarded_port", guest: 80, host: 8080

  config.vm.provision "shell", inline: <<-SHELL
    apt-get update
    apt-get install -qq apache2
  SHELL
end
~/d9 $ vagrant up
Bringing machine 'default' up with 'libvirt' provider...
==> default: Checking if box 'bento/debian-9' version '201906.17.0' is up to date...
==> default: Creating image (snapshot of base box volume).
==> default: Creating domain with the following settings...
==> default:  -- Name:              d9_default
==> default:  -- Domain type:       kvm
==> default:  -- Cpus:              1
==> default:  -- Feature:           acpi
==> default:  -- Feature:           apic
==> default:  -- Feature:           pae
==> default:  -- Memory:            512M
==> default:  -- Management MAC:    
==> default:  -- Loader:            
==> default:  -- Nvram:             
==> default:  -- Base box:          bento/debian-9
==> default:  -- Storage pool:      default
==> default:  -- Image:             /home/libvirt-images/d9_default.img (64G)
==> default:  -- Volume Cache:      default
==> default:  -- Kernel:            
==> default:  -- Initrd:            
==> default:  -- Graphics Type:     vnc
==> default:  -- Graphics Port:     -1
==> default:  -- Graphics IP:       127.0.0.1
==> default:  -- Graphics Password: Not defined
==> default:  -- Video Type:        cirrus
==> default:  -- Video VRAM:        9216
==> default:  -- Sound Type:	
==> default:  -- Keymap:            en-us
==> default:  -- TPM Path:          
==> default:  -- INPUT:             type=mouse, bus=ps2
==> default: Creating shared folders metadata...
==> default: Starting domain.
==> default: Waiting for domain to get an IP address...
==> default: Waiting for SSH to become available...
    default: 
    default: Vagrant insecure key detected. Vagrant will automatically replace
    default: this with a newly generated keypair for better security.
    default: 
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if it's present...
[...]

PR #1037, opened 2019-08-18
action/forward_ports.rb: Fix SSH tunnel spawning and PID tracking

zakame:
By-product while developing for #1036:

- Explicitly disable any SSH multiplexing here so PID tracking of
  tunnels works correctly.

- Using `exec ssh` in the spawn forces Ruby to use a subshell (as exec
  is a shell builtin) instead of spawning the ssh process directly, which
  results in getting the wrong (and dead, as the exec replaces the
  subshell) PID to track and clean up later.

- Run the ssh tunnel command on its own process group, essentially
  daemonizing it and keeping its PID intact even when `vagrant up` command
  is run on shell/consoles without an explicit TTY, such as Emacs Eshell.

### DIFF
--- a/lib/vagrant-libvirt/action/forward_ports.rb
+++ b/lib/vagrant-libvirt/action/forward_ports.rb
@@ -98,6 +98,7 @@ module VagrantPlugins
             Port=#{ssh_info[:port]}
             UserKnownHostsFile=/dev/null
             ExitOnForwardFailure=yes
+            ControlMaster=no
             StrictHostKeyChecking=no
             PasswordAuthentication=no
             ForwardX11=#{ssh_info[:forward_x11] ? 'yes' : 'no'}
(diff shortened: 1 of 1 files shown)


lib/vagrant-libvirt/action/forward_ports.rb
```diff
@@ -109,7 +110,7 @@ def redirect_port(machine, host_ip, host_port, guest_ip, guest_port,
           options += " -o ProxyCommand=\"#{ssh_info[:proxy_command]}\"" if machine.provider_config.connect_via_ssh
 
           # TODO: instead of this, try and lock and get the stdin from spawn...
-          ssh_cmd = 'exec '
```
electrofelix: This was originally added to ensure a `vagrant destroy` quickly followed by `vagrant up` would work as expected as previously the ssh process was not getting removed by a destroy. However your commit suggests that this shouldn't have worked as it should still have been getting the wrong PID back?

zakame: Exactly.  Moreover, the SSH process used might even be from one with multiplexing enabled (`ControlMaster=auto` in ssh config) so there might not even be a new process at all (e.g. from a previous destroy/up cycle, or from other/externally managed SSH processes.)

zakame:
Thanks!  I'm happy to add tests for destroy/up; let me check on how to do that here. :muscle:

Issue #626, opened 2016-07-05
Cannot boot instance with private network


stpierre:
The provider does not appear to create the `vagrant-private-dhcp` network for instances with a `private_network` interface.

### Steps to reproduce

1. `vagrant up` with a private network specified in the `Vagrantfile`

### Expected behaviour

The instance should come up with a private network interface.

### Actual behaviour

Vagrant errors out with: `Call to virDomainCreateWithFlags failed: Network not found: no network with matching name 'vagrant-private-dhcp'`

`virsh` agrees with its assessment:

```
% sudo virsh net-list
 Name                 State      Autostart     Persistent
----------------------------------------------------------
 default              active     yes           yes
 infra-ng0            active     no            yes
 vagrant-libvirt      active     no            yes
```

### System configuration
**OS/Distro version:**:

Fedora 23

**Libvirt version:**

libvirt-1.2.18.3-1.fc23.x86_64

**Output of `vagrant version; vagrant plugin list`:**

```
Installed Version: 1.8.1
Latest Version: 1.8.4
 
To upgrade to the latest version, visit the downloads page and
download and install the latest version of Vagrant from the URL
below:

  http://www.vagrantup.com/downloads.html

If you're curious what changed in the latest release, view the
CHANGELOG below:

  https://github.com/mitchellh/vagrant/blob/v1.8.4/CHANGELOG.md
landrush (1.0.0)
vagrant-libvirt (0.0.33, system)
```

**Output of `VAGRANT_LOG=debug vagrant up --provider=libvirt`**

[vagrant-up-log.txt](https://github.com/vagrant-libvirt/vagrant-libvirt/files/348459/vagrant-up-log.txt)

**A Vagrantfile to reproduce the issue:**
```
Vagrant.configure(2) do |config|
  config.vm.box = 'fedora/24-cloud-base'
  config.vm.network "private_network", type: "dhcp"
end
```

**Are you using upstream vagrant package or your distros package?**

Distro

infernix:
I could not reproduce this issue on a clean install of Fedora 23.

Please try upstream vagrant.

stpierre:
The same result obtains. Log: [vagrant-up-log.txt](https://github.com/vagrant-libvirt/vagrant-libvirt/files/348640/vagrant-up-log.txt)

```
% vagrant version; vagrant plugin list
Installed Version: 1.8.4
Latest Version: 1.8.4
 
You're running an up-to-date version of Vagrant!
[...]

PR #1038, opened 2019-08-28
Fix network creation

axmetishe:
Add ability to define multiple private networks
Enable auto configuration for tunnel interfaces
Add ability to manage bridge STP and delay parameters
Fix interaction with reload plugin

Related issues:
https://github.com/vagrant-libvirt/vagrant-libvirt/issues/626
https://github.com/vagrant-libvirt/vagrant-libvirt/issues/931​ (partial fix )
https://github.com/vagrant-libvirt/vagrant-libvirt/issues/877​


electrofelix:
Please don't submit multiple fixes in the same PR, send them up separately so that any questions/discussions can be focused

axmetishe:
@electrofelix Thank you for your reply.

Clearly understand your concerns regarding multiple fixes but it was like a series of issues during environment configuration for test execution against dpdk-based software router and some fixes were related to another one therefore I just compact them into single PR.

P.S. that was almost a year ago and I've forgot about this PR already.

h0st:
OK, I will do it. I have already started work on splitting it into several PRs.

electrofelix:
@axmetishe As I'm reading back given the recent update here, I've noticed I was rather short, apologies for that.

The networking code is somewhat of a bear to touch because it's not well tested, was written by others in past and consequently that means changes there are risky because I don't have the same understanding as to the *why* in a number of cases. This can make for difficult reviews, because *I* may not be using the behaviour that is being modified.

Recently I've merged some PRs that improve testing, especially one that does some integration testing that will exercise networking code will real examples.

Hopefully that should make the commits from this PR easier to merge in the near future, and where there are questions, it should be possible to answer them by simply adding another Vagrantfile with a minimal example to exercise. So I'm grateful that it's being picked up.

Recent release was focused a lot on just testing around configuration and extracting/populating the connection uri, as well as related configuration settings, so I was avoiding touching the networking code until the config was in better shape. I am hoping to add some networking tests to exercise the utils and create networks actions, though if @h0st gets there before me, would help immensely. Especially as I often only get a couple of hours every odd week to plug away.

lib/vagrant-libvirt/action/create_network_interfaces.rb
```diff
@@ -193,7 +193,7 @@ def call(env)
               # It's used for provisioning and it has to be available during provisioning,
               # ifdown command is not acceptable here.
               next if slot_number.zero?
-              next if options[:auto_config] === false
+              next if !options[:auto_config]
```
electrofelix, line 196: With the work to push this up with separate changes, it's been discovered that this causes a failure for the network configuration because it previously was been used to only skip if `auto_config` was explicitly set to false, whereas this changes it to skip unless it's set to `true` or a `truthy` value.

Consequently, this piece has been removed from what was merged.

Was this just an accidental change or was there something this fixed that needs to be looked at closer?

axmetishe: The main point of changes in that module was to allow tcp tunnels auto_config, so I think it's ok to set to explicit false value check in case it fails on some cases. The more important change is below this line where direct check on tunnels was removed.
Regards.

electrofelix:
On the good news side, it looks to me like all the functionality has been merged.

@h0st as you've been splitting it up into smaller pieces, can you confirm, or is there something still outstanding?

lib/vagrant-libvirt/action/create_network_interfaces.rb
```diff
@@ -213,9 +213,6 @@ def call(env)
                 network[:type] = :dhcp
               end
 
-              # do not run configure_networks for tcp tunnel interfaces
-              next if options.fetch(:tunnel_type, nil)
```
h0st, line 217: @axmetishe Do you mean this line in your last sentence? I can return a value for auto_config to false.

h0st:
@electrofelix yes I confirm this changes. There is nothing outstanding from my side.
I can setup a false value for auto_config and we check test again.

electrofelix:
@h0st thanks. I've already landed a fix to switch back to compare against `false` as one of the integration tests was failing on it. So should be good to close this one now I believe.

lib/vagrant-libvirt/action/create_network_interfaces.rb
axmetishe: No, I talked about  `next if options[:auto_config] === false` part.
Part regarding tunnels auto config `next if options.fetch(:tunnel_type, nil)` is crucial for project you are working on(if I'm not missed something=)

axmetishe: @h0st sorry for late response

PR #1041, opened 2019-08-30
Added more details on how to use vagrant-libvirt with QEMU user sessions

kittydoor:
While trying to setup QEMU Session with vagrant-libvirt, I had a lot of trouble trying to figure out what to do via the documentation. Reading through issues helped me understand what was exactly meant. Thus, I've decided to make this pull request to add a few more lines to clarify which kinds of networks are meant, and what files need to be edited, and how root access can be made unnecessary past the initial setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1241,7 +1241,14 @@ public network interface and you must connect to Libvirt via ssh.
 
 ## QEMU Session Support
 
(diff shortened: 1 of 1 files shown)


pvalena:
Great!

Issue #640, opened 2016-08-10
Feature: Option to set clock offset


joerg:
Currently the clock offset is hardcoded to UTC (https://github.com/vagrant-libvirt/vagrant-libvirt/blob/c86a6f452771d885b16227836b7354e984815731/lib/vagrant-libvirt/templates/domain.xml.erb#L59) which makes NTP a requirement for each VM. It would be nice to have an option to set this to localtime (localtime is what the clock offset is called by virt-manager, sorry, I don't know if this is also what needs to be set in the XML) so NTP is not required.

infernix:
Making clock a configurable variable is fine with me, but I can't think of any scenario where localtime is needed other than for Windows. Are you experiencing clock drift?

joerg:
Yes, we are. Our servers run on UTC+1 and all Vagrant boxes thus have the wrong time. I have no problem with using NTP, but why run another daemon if my host already has the correct time?
Maybe this also is something specific with our quite old RedHat 6, but anyways, replacing the hardcoded utc with something variable would be a nice idea in my eyes.

infernix:
But linux generally expects system/bios time to be UTC. Are you sure this is not a timezone issue in the guest OS that can be fixed by setting appropriate timezone? What guest OS are you running?

Furthermore clock drift is the time going out of sync, not an exact X hour offset. So is this clock drift or is this a timezone offset issue?

joerg:
Sorry, to be clear, it is not a drift but an offset. The timezone however is, to all my knowledge, correctly set in /etc/sysconfig/clock and /etc/localtime, which is a copy of /usr/share/zoneinfo/Europe/Vienna. We are using RedHat 6.8 as host and guest OS.
The reason I suggested making the clock offset configurable is because I tried setting "Clock Offset" in virt-manager to "localtime" and stopped and restarted the machine and then the time was correct.

infernix:
Can you try:

1. qemu should be presenting bios to VM with UTC time, check with `hwclock --show` - is this time valid and UTC?
2. I thought RHEL 6 would assume UTC, but if not try adding 'UTC=true` to `/etc/sysconfig/clock` and poweroff/on. Time should then be correct without ntp

Some more pointers here https://bugzilla.redhat.com/show_bug.cgi?id=859868

ansiwen:
[...]

PR #1047, opened 2019-09-24
Add support for clock setup

zzambers:
This adds support for setting clock offset and timers. See [1].

[1] https://libvirt.org/formatdomain.html#elementsTime


zzambers:
Referencing some related issues:
https://github.com/vagrant-libvirt/vagrant-libvirt/issues/640
https://github.com/vagrant-libvirt/vagrant-libvirt/issues/776

dragon788:
This would be really useful for cases like nested virtualization where there might otherwise be issues on wake-up if the VM is trying to use NTP or localtime rather than a virtual clock.

electrofelix:
I need to add a couple of unit tests around the domain template and possibly to cover starting a domain if the clock settings have been changed since it was originally created. @dragon788 hopefully I'll add them in the next couple of days to be able to merged this then.

PR #1062, opened 2019-11-16
config: allow for setting URI from environment

dustymabe:
Users are allowed to set a LIBVIRT_DEFAULT_URI environment variable that
controls tools (i.e. virsh, virt-install, etc) that communicate with
libvirt. Let's allow for that mechanism to be used here.

### DIFF
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -586,14 +586,21 @@ module VagrantPlugins
         @qemu_args << options if options[:value]
       end
 
-      # code to generate URI from a config moved out of the connect action
-      def _generate_uri
+      # code to generate URI from from either the LIBVIRT_URI environment
+      # variable or a config moved out of the connect action
+      def _generate_uri(qemu_use_session)
+
+        # If the LIBVIRT_DEFAULT_URI var is set, we'll use that
+        if ENV.fetch('LIBVIRT_DEFAULT_URI', '') != ""
+          return ENV['LIBVIRT_DEFAULT_URI']
+          
+
         # builds the Libvirt connection URI from the given driver config
         # Setup connection uri.
         uri = @driver.dup
         virt_path = case uri
                     when 'qemu', 'kvm'
-                      @qemu_use_session ? '/session' : '/system'
+                      qemu_use_session ? '/session' : '/system'
                     when 'openvz', 'uml', 'phyp', 'parallels'
                       '/system'
                     when '@en', 'esx'
@@ -659,10 +666,20 @@ module VagrantPlugins
         @management_network_domain = nil if @management_network_domain == UNSET_VALUE
         @system_uri      = 'qemu:///system' if @system_uri == UNSET_VALUE
 
-        @qemu_use_session = false if @qemu_use_session == UNSET_VALUE
+        # If uri isn't set then let's build one from various sources.
+        # Default to passing false for qemu_use_session if it's not set.
+        if @uri == UNSET_VALUE
+            @uri = _generate_uri(
+                    @qemu_use_session == UNSET_VALUE ? false : @qemu_use_session)
 
-        # generate a URI if none is supplied
-        @uri = _generate_uri if @uri == UNSET_VALUE
+        # Set qemu_use_session based on the URI if it wasn't set by the user
+        if @qemu_use_session == UNSET_VALUE
+          if (@uri.include? "qemu") && (@uri.include? "session")
+            @qemu_use_session = true
+          else
+            @qemu_use_session = false
+          end
+        end
 
         # Domain specific settings.
         @uuid = '' if @uuid == UNSET_VALUE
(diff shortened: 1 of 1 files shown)


dustymabe:
hello all - anybody able to give a review for something like this?

pvalena:
Well, it's a lot of hidden logic TBH, for `LIBVIRT_DEFAULT_URI`. Does it mean that it will not be used If I set uri in my `~/.vagrant.d/Vagrantfile`?

Also there's this PR: https://github.com/vagrant-libvirt/vagrant-libvirt/pull/997
Doesn't that cover the use case you have?

dustymabe:
> Well, it's a lot of hidden logic TBH, for `LIBVIRT_DEFAULT_URI`. Does it mean that it will not be used If I set uri in my `~/.vagrant.d/Vagrantfile`?

We can make the code behave in whatever way we like if we decide on the best precedent for where the value should be taken from.

I think the bigger thing here is: if you have a URI, then why do you need a user exposed `qemu_use_session` variable ? Why not always derive an internal `qemu_use_session` value from the URI and default the URI to session?

That way the user only needs to set one variable. What happens today if the user sets both `uri` and `qemu_use_session` to conflicting values?

```
    libvirt.uri = "qemu:///system"
    libvirt.qemu_use_session = true
  end
end
```

> 
> Also there's this PR: #997
> Doesn't that cover the use case you have?

Yes. #997 is a different way to cover the use case I have.

jdoss:
I didn't know virsh, virt-install, etc used `LIBVIRT_DEFAULT_URI=qemu:///system` that is awesome. Having an env var to set globally is more ideal IMO than having to add lines to Vagrantfiles to get Libvirt to use system over the user session, so either `LIBVIRT_DEFAULT_URI` or `VAGRANT_LIBVIRT_QEMU_USE_SESSION` works for my use cases.

pvalena:
Well, what you're saying is that you want to use ENVs for configuring vagrant, which is a logic, that should be in `Vagrantfile` IMO (but that may be just based on philosophy).

As a workaround you can input in your code in Vagrantfile (__untested__):
```ruby
    if ENV['LIBVIRT_DEFAULT_URI']
       libvirt.uri = ENV['LIBVIRT_DEFAULT_URI']
       if (libvirt.uri.include? "qemu") && (libvirt.uri.include? "session")
          @qemu_use_session = true
       else
          @qemu_use_session = false
       end
    end
end
```

pvalena:
As a sidenote, uri can point to any qemu-compatible interface, e.g. one tunelled via ssh: https://libvirt.org/drvqemu.html
That's why I consider this autodetection logic potentionally dangerous.

pvalena:
```
That way the user only needs to set one variable. What happens today if the user sets both uri and qemu_use_session to conflicting values?
```
It [just hangs](https://paste.centos.org/view/84f702df), waiting for IP. (I guess you have to provide valid configuration...)

That reminds me of the errors that occur when libvirtd is not running, or KVM is not available. Those could be better.

lib/vagrant-libvirt/config.rb
```diff
@@ -659,10 +659,26 @@ def finalize!
...
+        # Default to false for qemu_use_session
         @qemu_use_session = false if @qemu_use_session == UNSET_VALUE
 
-        # generate a URI if none is supplied
-        @uri = _generate_uri if @uri == UNSET_VALUE
+        # If uri is not set then:
+        #   - set based on LIBVIRT_DEFAULT_URI in env if exists
+        #   - OR generate it from scratch
+        if @uri == UNSET_VALUE
+          if ENV['LIBVIRT_DEFAULT_URI']
```
electrofelix: Probably going to have to do a little more testing here, as `LIBVIRT_DEFAULT_URI= ...` is sufficent not to return nil resulting in entering this block but setting the uri to empty which would be undesirable.

lib/vagrant-libvirt/config.rb
```diff
@@ -659,10 +659,26 @@ def finalize!
...
+            @uri = ENV['LIBVIRT_DEFAULT_URI']
+            # If we're setting the URI from the env var then let's
+            # appropriately update @qemu_use_session based on the @uri
+            if (@uri.include? "qemu") && (@uri.include? "session")
+              @qemu_use_session = true
+            else
+              @qemu_use_session = false
+            end
```
electrofelix: Should only do this if `@qemu_use_session` has not been explicitly set by configuration, Probably easier to to split this if/else block to just an if and leave the original statement to set `@uri` as it was below. Then you can move the `@qemu_use_session` setting line from above this block to below just before resolving `@uri` if not explicitly set. That way can test at this point whether `@qemu_use_session == UNSET_VALUE` before trying to infer the value from the current contents.

Of course possibly it might be a good idea to move all of this into a separate method and that way we could easily add some tests just for that method to check that the different scenarios are correct covered to avoid surprises.

dustymabe: ok I think I've addressed your concerns by moving the inspecting of the environment var into the `_generate_uri` function. I don't have a lot of time so figuring out the tests would be something I'd appreciate some help on if you don't mind.

dustymabe:
OK. I moved the inspecting the LIBVIRT_URI environment variable to inside the `_generate_uri()` function. It's not yet tested, but let me know what you think and I'll run some tests.

dustymabe:
also added some error checking in the validate function. very much untested for now

dustymabe:
@electrofelix mind giving this another look?

electrofelix:
@dustymabe I've very behind on email, are you aware that you have syntax errors in the PR?

https://travis-ci.com/github/vagrant-libvirt/vagrant-libvirt/jobs/265668570#L555

electrofelix:
as I didn't have write permissions to the original branch, I've opened #1141 in place of this PR

dustymabe:
sorry about not replying sooner. Thanks for carrying this forward @electrofelix !

PR #1079, opened 2020-02-23
Use fog-libvirt filters when querying libvirt host

ccrebolder:
This commit replaces the pattern where the metadata for all volumes,
servers, or pools was downloaded and then searched. Instead a filter
argument is passed to the connection and only the metadata for the named
resource is returned, if it exists. This results in significant speed
increases for libvirt hosts that are offsite or have many volumes.

Example timings for `vagrant up` to an offsite libvirt host with lots of volumes/images over ssh, assuming the box image is already present:
* master:
```
$ time vagrant up
Bringing machine 'default' up with 'libvirt' provider...
real    3m32.099s
user    0m5.531s
sys     0m2.176s
```
* this PR:
```
$ time vagrant up
Bringing machine 'default' up with 'libvirt' provider...
<snip>
real    1m25.364s
user    0m3.290s
sys     0m0.603s
```

All tests pass and I manually checked some of the failure/error states to make sure they still work properly:
* domain name is already in use
* domain volume exists already
* upload box image if it doesn't exist/use it if it does exist
* error if storage_pool_name doesn't exist
* box_volume doesn't exist when removing stale volumes
* pool no longer exists when removing stale volumes

### DIFF
--- a/lib/vagrant-libvirt/action/create_domain_volume.rb
+++ b/lib/vagrant-libvirt/action/create_domain_volume.rb
@@ -25,15 +25,15 @@ module VagrantPlugins
           @name = "#{env[:domain_name]}.img"
 
           # Verify the volume doesn't exist already.
-          domain_volume = ProviderLibvirt::Util::Collection.find_matching(
-            env[:machine].provider.driver.connection.volumes.all, @name
-          )
-          raise Errors::DomainVolumeExists if domain_volume
+          domain_volume = env[:machine].provider.driver.connection.volumes.all(
+            name: @name
+          ).first
+          raise Errors::DomainVolumeExists if domain_volume.id
 
           # Get path to backing image - box volume.
-          box_volume = ProviderLibvirt::Util::Collection.find_matching(
-            env[:machine].provider.driver.connection.volumes.all, env[:box_volume_name]
-          )
+          box_volume = env[:machine].provider.driver.connection.volumes.all(
+            name: env[:box_volume_name]
+          ).first
           @backing_file = box_volume.path
 
           # Virtual size of image. Take value worked out by HandleBoxImage
--- a/lib/vagrant-libvirt/action/create_domain_volume.rb
+++ b/lib/vagrant-libvirt/action/create_domain_volume.rb
@@ -28,7 +28,7 @@ module VagrantPlugins
           domain_volume = env[:machine].provider.driver.connection.volumes.all(
             name: @name
           ).first
-          raise Errors::DomainVolumeExists if domain_volume.id
+          raise Errors::DomainVolumeExists if domain_volume && domain_volume.id
 
           # Get path to backing image - box volume.
           box_volume = env[:machine].provider.driver.connection.volumes.all(
--- a/lib/vagrant-libvirt/action/handle_box_image.rb
+++ b/lib/vagrant-libvirt/action/handle_box_image.rb
@@ -63,9 +63,10 @@ module VagrantPlugins
           # locking all subsequent actions as well.
           @@lock.synchronize do
             # Don't continue if image already exists in storage pool.
-            break if env[:machine].provider.driver.connection.volumes.all(
+            box_volume = env[:machine].provider.driver.connection.volumes.all(
               name: env[:box_volume_name]
-            ).first.id
+            ).first
+            break if box_volume && box_volume.id
 
             # Box is not available as a storage pool volume. Create and upload
             # it as a copy of local box image.
--- a/lib/vagrant-libvirt/action/handle_box_image.rb
+++ b/lib/vagrant-libvirt/action/handle_box_image.rb
@@ -63,9 +63,9 @@ module VagrantPlugins
           # locking all subsequent actions as well.
           @@lock.synchronize do
             # Don't continue if image already exists in storage pool.
-            break if ProviderLibvirt::Util::Collection.find_matching(
-              env[:machine].provider.driver.connection.volumes.all, env[:box_volume_name]
-            )
+            break if env[:machine].provider.driver.connection.volumes.all(
+              name: env[:box_volume_name]
+            ).first.id
 
             # Box is not available as a storage pool volume. Create and upload
             # it as a copy of local box image.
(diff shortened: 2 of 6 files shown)


lib/vagrant-libvirt/action/handle_box_image.rb
```diff
@@ -64,9 +64,9 @@ def call(env)
           # locking all subsequent actions as well.
           @@lock.synchronize do
             # Don't continue if image already exists in storage pool.
-            break if ProviderLibvirt::Util::Collection.find_matching(
-              env[:machine].provider.driver.connection.volumes.all, env[:box_volume_name]
-            )
+            break if env[:machine].provider.driver.connection.volumes.all(
+              name: env[:box_volume_name]
+            ).first.id
```
electrofelix, line 69: does the volumes.all always return something, otherwise I'd imagine this could result in a nil access error if the list is empty?

lib/vagrant-libvirt/action/create_domain_volume.rb
```diff
@@ -25,15 +25,15 @@ def call(env)
...
 
           # Verify the volume doesn't exist already.
-          domain_volume = ProviderLibvirt::Util::Collection.find_matching(
-            env[:machine].provider.driver.connection.volumes.all, @name
-          )
-          raise Errors::DomainVolumeExists if domain_volume
+          domain_volume = env[:machine].provider.driver.connection.volumes.all(
+            name: @name
+          ).first
+          raise Errors::DomainVolumeExists if domain_volume.id
```
electrofelix, line 31: does the volumes.all always return something, otherwise I'd imagine this could result in a nil access error if the list is empty?

lib/vagrant-libvirt/action/remove_stale_volume.rb
```diff
@@ -20,20 +20,20 @@ def call(env)
...
           config = env[:machine].provider_config
           # Check for storage pool, where box image should be created
-          fog_pool = ProviderLibvirt::Util::Collection.find_matching(
-            env[:machine].provider.driver.connection.pools.all, config.storage_pool_name
-          )
-          @logger.debug("**** Pool #{fog_pool.name}")
+          fog_pool = env[:machine].provider.driver.connection.pools.all(
+            name: config.storage_pool_name
+          ).first
+          @logger.debug("**** Pool #{fog_pool.nil? ? 'not found' : fog_pool.name}")
```
electrofelix, line 26: Suggested to someone in another review, it might be easier if this just calls `@app.call(env)` with `env[:result] = nil` if fog_pool is nil at this point. Return early rather than continuing to process.

lib/vagrant-libvirt/action/create_domain_volume.rb
electrofelix: I've a feeling I've simply forgotten how stuff works here over time, because I can't see how a vagrant up would work on a new instance if this did an attribute access and domain_volume was nil

electrofelix: @ccrebolder looks like this is benefiting from a quirk in that use of a filter will always return an empty volume if no match is found rather than no volume. When I use a similar call with pools and a non-existing name, it returns an empty list, so it's likely this will change in the future and I think it would be worth checking `domain_volume` is not nil before testing `domain_volume.id` to guard against when this gets fixed upstream

lib/vagrant-libvirt/action/handle_storage_pool.rb
```diff
@@ -24,9 +24,9 @@ def call(env)
           # locking all subsequent actions as well.
           @@lock.synchronize do
             # Check for storage pool, where box image should be created
-            break if ProviderLibvirt::Util::Collection.find_matching(
-              env[:machine].provider.driver.connection.pools.all, config.storage_pool_name
-            )
+            break unless env[:machine].provider.driver.connection.pools.all(
+              name: config.storage_pool_name
+            ).empty?
```
electrofelix, line 29: you can see the correct behaviour here regarding no matching pool found.

lib/vagrant-libvirt/action/remove_stale_volume.rb
electrofelix: though worth printing out the debug message first anyway, so just an early call to @app.call(env) here, might see if it can be added via a different PR and you can then fix this one up

lib/vagrant-libvirt/action/create_domain_volume.rb
ccrebolder: @electrofelix that's a good call, it does stand out as something fog will want to normalize one day. Will get reacquainted and work on this today, thanks for taking the time to look at it.

electrofelix: I've put up fog/fog-libvirt#75 to trigger a conversation on it as well

lib/vagrant-libvirt/action/remove_stale_volume.rb
ccrebolder: let me know if my solution is too longwinded and I can remove it and wait for the other PR

lib/vagrant-libvirt/action/create_domain_volume.rb
ccrebolder: thanks, I'm checking for both `domain_volume` (if empty list is returned) and `domain_volume.id` (current behaviour of returning a single empty volume) being `nil` now in 669e5d9dfb34b82c93832397bf1518f4f4cb6482

electrofelix:
@ccrebolder I've consolidated some of the changes with another PR I was also helping get merged around lib/vagrant-libvirt/action/remove_stale_volume.rb, can you check to make sure that I haven't botched anything up with this. Assuming you're happy with the result, I'll merge it after the weekend and then try and release 0.1.0

lib/vagrant-libvirt/action/remove_stale_volume.rb
electrofelix: I updated to rebase on top of another merged version, think I got the same logic in place

ccrebolder:
Thanks @electrofelix I took a look at the remove_stale_volume.rb in the new commit and it looks just fine :+1:

PR #1086, opened 2020-04-01
Add option to configure static routes for libvirt guest network interfaces

thomasDOTwtf:
Template rending for guest systems use the configured route property to render static routes.
To make it possible to use static routes for network interfaces this value should be passed to the guest plugins. 

The modification is tested against the following environment:
- Vagrant 2.2.3
- vagrant-libvirt (0.0.45, system)

Works with and without a defined value for the route property.


thomasDOTwtf:
Hey @abbbi might you be able to review and merge?

electrofelix:
@thomasDOTwtf thanks, sorry I didn't grok what this was doing for a while, essentially it's a no-op for vagrant-libvirt and just used by other plugins.

PR #1087, opened 2020-04-04
Support discard for disk

jeffrey4l:
For better performance should allow enabling the virtio-scsi
model on the scsi controller in VM's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,9 @@ end
   _target_](http://libvirt.org/formatdomain.html#elementsDisks). NOTE: this
   option applies only to disks associated with a box image. To set the bus type
   on additional disks, see the [Additional Disks](#additional-disks) section.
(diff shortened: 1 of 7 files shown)


jeffrey4l:
this pr is come from pr #692, which is not updated for a long time. and this patch try to fix #486

davidscaife:
Seems pertinent to note that PR #1000 is also relevant to this.

electrofelix:
Catching up on some older PRs. I believe #1000 covers the discard and the virtio-scsi needs to be reworked to support multiple disks before it can be landed. Always needed to be reworked to ensure it calculated the correct number of controllers anyway depending on the allocated disks.

Going to close this PR and either will try at some point to bring #692 up to date or if you want to open a new PR based on that one to take over from it.

PR #1091, opened 2020-04-19
Use Github actions for integration tests against Vagrant-libvirt

abbbi:
Allows to spin up vagrant boxes via vagrant-libvirt in CI replacing the 
need for vagrant-libvirt-qa to be executed separately before/after 
releases.

* Adds a github integration tests workflow
* Installs the needed toolset in separate groups for easier tracking
* Add Vagrantfile definitions for the first simple test cases in tests/ 
  using tinycore linux, using the installed master version these boxes 
  are spun up and destroyed via a test matrix.

Should help reduce effort for new PRs and maintainers.


abbbi:
hi,
idk, to be honest its the first time i worked with travis :)
But we have to keep in mind that the free instance provided by travis has only limited resources in regards to memory and cpu, so spinning up multiple virtual machines at the same time my prone to run out of resources.

AFAIK BATS supports running tests in parallel using the parallel feature, but then the we have to switch to using bats-core, instead of the bats provided by the ubuntu package.

abbbi:
hi again,

i have made an attempt to include a matrix with this kind of QA scenario, an example build can be seen here:

 https://travis-ci.org/github/abbbi/vagrant-libvirt/builds/677798662

it attempts to run multiple jobs, one with master version and one with vagrant version 2.2.7
depending on the latest vagrant-libvirt checkout.Currently still fails because of the stale volume issue, but should give an impression whats possible

electrofelix:
@abbbi if it's proving difficult to support both types of tests, I can look at what we can run for free with CircleCI to handle the unit test side and have TravisCI focus on the more integration level?

abbbi:
> @abbbi if it's proving difficult to support both types of tests, I can look at what we can run for free with CircleCI to handle the unit test side and have TravisCI focus on the more integration level?

its fine with travis, no problem, see here for a latest build i tested:

https://travis-ci.org/github/abbbi/vagrant-libvirt/builds/679381498

we just have to adopt .travis.yml for the ruby versions and to run rspec. I think it makes more sense to use bionic as testing base as it comes with a newer libvirt version.
Im going to rebase and force push this branch so we dont have as much commits

electrofelix:
@abbbi I've switched this over to github actions as travis is no longer straight forward to get test minutes, just stuck on any issue with the shared folders.

electrofelix (approved):
I think finally good to go with this, thanks @abbbi for putting in the hard yards to have something that could exercise vagrant-libvirt a bit more.

PR #1098, opened 2020-04-22
Allow dev with latest vagrant to use newer vagrant-spec

electrofelix:
Latest vagrant depends on a vagrant-spec release containing a more
recent dependency on childprocess than is supported for earlier releases
of vagrant.

Adjust dependencies to pin to the specific release of vagrant-spec if
the request version of vagrant is 2.2.7 or older.

Does not attempt to handle the situation where a version is specified
that cannot be parsed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ env:
   matrix:
     - VAGRANT_VERSION=v2.0.4
     - VAGRANT_VERSION=v2.1.5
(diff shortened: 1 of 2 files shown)


electrofelix:
@abbbi does this take care of the issue you ran into for dev?

abbbi:
> @abbbi does this take care of the issue you ran into for dev?

yes, looks good. Now installs nicely with all defined versions.

PR #1107, opened 2020-05-09
With additional notes for Fedora

buckaroogeek:
Attached is an update with additional troubleshooting steps for Fedora users.


README.md
```diff
@@ -180,6 +180,43 @@ If you have issues building ruby-libvirt, try the following:
...
+then the following steps have been found to resolve the problem. After the steps below are complete, then reinstall the vagrant-libvirt plugin. Thanks to Marco Bevc (see https://github.com/hashicorp/vagrant/issues/11020#issuecomment-625801983):
+
+```shell
+dnf download --source krb5-libs
+rpm2cpio krb5-1.18-1.fc32.src.rpm | cpio -imdV
+tar xf krb5-1.18.tar.gz
+cd krb5-1.18/src
+./configure
+make
+sudo cp -a lib/crypto/libk5crypto.* /opt/vagrant/embedded/lib64/
```
electrofelix, line 218: Is there any reason you couldn't just copy the file from the current system into the directory without needing to compile again from source? The previous item it make sense because it is linking against some libraries provided by the vagrant upstream package, but for this one?

buckaroogeek: The libssh build explicitly uses the openssl library packaged with upstream vagrant to resolve the missing symbol problem. For the kerberos package, it would seem feasible to just copy the libk5crypto.so* files from /usr/lib64 to the /opt/vagrant/embedded/lib64 directory and then doing the plugin installation. I tried that at one point and it did not work (for reasons that elude me still).  Marco Bevc provided the build steps that I copied - I would have to defer to input from him. I should probably create a clean vm with fedora 32 workstation and try a few alternatives to try and answer the question.

electrofelix: It should be possible to confirm locally, if you copy across the file from /usr/lib64 to /opt/vagrant/embedded/lib64/ (ensuring it replaces the version currently there), either you can vagrant up successfully which means that you don't need to compile, or you'll get an error that indicates the system library has additional requirements and by compiling from source you skip pulling those in which is why the instructions above are needed. Shouldn't need a pristine fedora 32 system to verify

dcermak:
@buckaroogeek any particular reason why Fedora users should not use the system package? After all it is just one `dnf install vagrant-libvirt` away and @voxik and @pvalena have been taking good care of it.

buckaroogeek:
@dcermak - I have encountered two scenarios where using upstream can be helpful. First, the version of vagrant available in the stable repository can sometimes lag weeks if not longer behind the upstream version. The updated version from upstream can resolve problems or add a useful feature that is still missing for the system package. I, for one, fully recognize and value the hard work that package maintainers accomplish so this is not to disparage their contributions. I just want to point out that sometimes it can be beneficial to use upstream.

Second, the system package uses the system ruby package unlike upstream which bundles its own version of ruby. This can cause a problem for some plug-ins. I think (but would need to research) that this is due to dependency conflicts in ruby packages.

I myself would greatly prefer to use the system package and sometimes switch back to it. I started using upstream because of problems getting vagrant-libvirt and another package (vagrant-bolt?) to install and coexist. The system package for vagrant-libvirt will sometimes lag also. Since upstream provides it's own instance of ruby - this should always work for plugins (in theory :) ).

Brad

dcermak:
> I think (but would need to research) that this is due to dependency conflicts in ruby packages.

No, this is because vagrant does something very weird when resolving dependencies during `vagrant plugin install` which often breaks. There's a few issues about that upstream but they haven't fixed it yet.

buckaroogeek:
@dcermak  - thanks. Appreciate the insights.

voxik:
Just FTR, there might be dependency issues, because we need sometimes more recent versions (e.g. we had fog-core 2.2+, but other parts of fog depended on fog-core 1.45). But the main issue is that Vagrant downloads the dependency information from rubygems.org and does not respect the local changes. That is very inconvenient. So far, I had not cycles to nail it down (not sure if I will have any time soon)

README.md
electrofelix: I think I know why just using the system libraries doesn't work. The recompile here skips linking with certain other libraries, which avoids the issue of the missing symbols. There are issues trying to get this to work for CentOS however I'll try to verify on Fedora and merge

electrofelix: Confirmed, this works as needed on Fedora as the complied krb5 libraries aren't linked against openssl, the spec file for the packge contains the following done during the prep stage:
```
    --with-crypto-impl=openssl \
    --with-tls-impl=openssl \
```

That would also explain why can't use `rpmbuild -bp <specfile>` to set up a build tree for it.

PR #1108, opened 2020-05-10
Tidy up README and TOC

electrofelix:
Bring the TOC in line with recent additions and enable the ability to
use a vim plugin to quickly regenerate in lieu of a better alternative.

Using https://github.com/mzlogin/vim-markdown-toc

Clean up any trailing whitespace so that subsequent file updates
by other PRs will be able to simply regen using appropriate tooling
without pulling in additional changes unrelated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ can help a lot :-)
 
 ## Index
 
+<!-- note in vim set "let g:vmt_list_item_char='-'" to generate the correct output -->
+<!-- vim-markdown-toc GFM -->
 
 - [Features](#features)
 - [Future work](#future-work)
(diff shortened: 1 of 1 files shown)


electrofelix:
There are other options for generating the TOC automatically:
- https://github.com/atheiman/gfm-toc
- https://ecotrust-canada.github.io/markdown-toc/
- https://github.com/vaidik/markdown-toc-generator
- https://github.com/jonschlinkert/markdown-toc
- https://github.com/ekalinin/github-markdown-toc or https://github.com/ekalinin/github-markdown-toc.go

This just seemed the most expedient after looking through a few of them where there doesn't seem to be an easy option to ignore the preceding headers.

Issue #1106, opened 2020-05-09
Correctly support the new --no-tty option


mrc0mmand:
Since v2.2.7 (https://github.com/hashicorp/vagrant/pull/11465) Vagrant now supports `--no-tty` option, which should suppress progress bars and other interactive elements when running in scripted environments, to not clutter logs (which especially the progress bar does with a passion). 

However, to fully support this, a slight change in the codebase is needed as well, as currently `vagrant up --provider=libvirt` spams the output with newlines when importing the VM image for the first time.

From the PR I mentioned before, it looks like this part:

https://github.com/vagrant-libvirt/vagrant-libvirt/blob/60ef4b03d12cc6f40ca30cf8af8bcb4db2331802/lib/vagrant-libvirt/action/handle_box_image.rb#L102-L107

needs to be changed to something like this:

```ruby
  ret = upload_image(box_image_file, config.storage_pool_name, 
                    env[:box_volume_name], env) do |progress| 
    ui.rewriting do |ui|
      ui.clear_line
      ui.report_progress(progress, box_image_size, false)
    end
  end
```

I use the term *something like this* quite literally, as I have basically zero knowledge of Ruby, so please take it with a grain of salt.

PR #1120, opened 2020-06-05
Support the newly-introduced --no-tty Vagrant's option

mrc0mmand:
hashicorp/vagrant#11465 introduced a new option --no-tty which silences
progress bars so they don't spam logs when Vagrant is used
non-interactively. First Vagrant version with this change is v2.2.8.

However, existing code needs to be slightly updated to cooperate with
the new option correctly, otherwise the progress bars spam the logs with
empty lines/warnings instead of remaining silent.

Fixes: #1106


lib/vagrant-libvirt/action/handle_box_image.rb
```diff
@@ -102,13 +102,15 @@ def call(env)
             # Upload box image to storage pool
             ret = upload_image(box_image_file, config.storage_pool_name,
                                env[:box_volume_name], env) do |progress|
-              env[:ui].clear_line
-              env[:ui].report_progress(progress, box_image_size, false)
+              env[:ui].rewriting do |ui|
```
electrofelix, line 105: It appears this is quite a new API, so if we add this here, the plugin will break for all previous versions of vagrant. While this isn't an issue for some users, there have been enough issues with trying to use the upstream release of Vagrant on newer fedora releases due to mismatches in library versions that it would be better to try and avoid breaking compatibility

Maybe add a utility function that can yield based on whether `env[:ui]` responds to the rewriting method or not.

Just knocking this out from memory rather than checking that it works:
```
def rewriting(ui) do
    if ui.responds_to?(:rewriting)
        yield ui.rewriting
    else
        yield ui
    end
end
```

Then if we call this utility with `env[ui]` we should get the new behaviour for newer versions of vagrant while continuing to work with older releases, this code here would become:

```
    util.rewriting(env[:ui]) do |ui|
        ui.clear_line
        ui.report_progress(progress, box_image_size, false)
    end
```

Should be able to add the helper method under the util directory https://github.com/vagrant-libvirt/vagrant-libvirt/tree/master/lib/vagrant-libvirt/util, and simply include the module as needed.

mrc0mmand: Thank you, that definitely makes sense. I had to tweak the `rewriting()` method a bit, otherwise it would throw exceptions as the `Vagrant::UI::Inteface::rewriting()` method yields `self` or nothing [depending on the object](https://github.com/hashicorp/vagrant/blob/master/lib/vagrant/ui.rb).

However, as this is my first experience with Ruby, I'm not sure if mine approach is the best, so please feel free to kick me in the right direction.

electrofelix (approved):
lgtm, sorry it took so long to get back to this.

PR #1126, opened 2020-06-28
added ability to use emulated TPM

### LINKED ISSUE #965: Add support for tpm_version
When using the TPM `emulator` backend (available since libvirt v4.5.0 (2018-07-02)) we should be able to set the tpm version, e.g. `tpm_version = "2.0"`.

Plase see https://libvirt.org/formatdomain.html#elementsTpm and https://libvirt.org/news.html

JTarasovic:
closes #965

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,7 +27,7 @@ group :development do
   if vagrant_gem_version <= Gem::Version.new('2.2.7')
     gem 'vagrant-spec', :github => 'hashicorp/vagrant-spec', :ref => '161128f2216cee8edb7bcd30da18bd4dea86f98a'
   else
-    gem 'vagrant-spec', :github => 'hashicorp/vagrant-spec'
+    gem 'vagrant-spec', :github => 'hashicorp/vagrant-spec', :branch => "main"
   end
 
   gem 'pry'


JTarasovic:
I'm definitely not a Ruby dev. Let me know if there's anything I could / should be doing differently. Largely, though, it just extends on the previous TPM work.

rgl:
I've tried this patch as:

```bash
cd ~/.vagrant.d/gems/2.6.6/gems/vagrant-libvirt-0.1.2
tar czf ../vagrant-libvirt-0.1.2.tgz .
wget -O- https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1126.patch | patch --strip 1
```

Then with a Vagrantfile as:

```ruby
Vagrant.configure(2) do |config|
  config.vm.box = 'ubuntu-20.04-amd64'
  config.vm.provider 'libvirt' do |lv, config|
    lv.tpm_model = 'tpm-tis'
    lv.tpm_type = 'emulator'
    lv.tpm_version = '2.0'
...
```

Then did a `vagrant up`.

Everything worked except this apparent harmless error that shows up every-time a `vagrant up` (either at creating a new domain, or just starting up one again after a vagrant halt) is executed:

```
==> default: Starting domain.
==> default: Error when updating domain settings: undefined method `[]' for nil:NilClass
``` 

It doesn't seem to prevent the correct creation of the swtpm device (and the usage of `/dev/tpm0` inside the VM).

But its quite annoying to have that error pop up in there. Unfortunately it does not have a stack trace.

Have you notice it before?

lib/vagrant-libvirt/action/start_domain.rb
```diff
@@ -178,16 +178,21 @@ def call(env)
               end
 
               # TPM
-              if config.tpm_path
+              if config.tpm_path || config.tpm_version
                 raise Errors::FogCreateServerError, 'The TPM Path must be fully qualified' unless config.tpm_path[0].chr == '/'
```
rgl, line 182: This needs to be changed to:

```suggestion
                if config.tpm_path
                  raise Errors::FogCreateServerError, 'The TPM Path must be fully qualified' unless config.tpm_path[0].chr == '/'
                end
```

Otherwise this always errors with:

```
Error when updating domain settings: undefined method `[]' for nil:NilClass
```

When I use an emulated TPM:

```ruby
Vagrant.configure(2) do |config|
  config.vm.box = 'ubuntu-20.04-amd64'
  config.vm.provider 'libvirt' do |lv, config|
    lv.tpm_model = 'tpm-tis'
    lv.tpm_type = 'emulator'
    lv.tpm_version = '2.0'
...
```

lib/vagrant-libvirt/action/start_domain.rb
```diff
@@ -178,16 +178,21 @@ def call(env)
...
 
+                if config.tpm_version
+                  raise Errors::FogCreateServerError, 'The TPM Version only works with the `emulator` backend type' unless config.tpm_type == 'emulator'
+                end
+
                 tpm = REXML::XPath.first(xml_descr, '/domain/devices/tpm')
                 if tpm.nil?
```
rgl, line 189: Do you known why `tpm` is always `nil`?

Every-time I start an stopped domain the tpm configuration is not inside `xml_descr`. Looking at the code, I would expect this code path to only be executed when there is a change in the tpm settings. So I'm not sure what is going on. Thou is that in the end it ends up working inside the VM, but something odd is happening here.

rgl (commented):
I found the origin of the error, please see my comments.

lib/vagrant-libvirt/action/start_domain.rb
JTarasovic: I'll have to dig back in and take a look. It's long since been paged out. :joy:

I think I was initially hoping to be able to use qemu:///session and you'd specify the path that you wanted swtpm to create the socket but it didn't work for some reason that I can't recall. Looks like the ability to run as any user was just added in a release from 4 days ago?

JTarasovic: Either way, my logic is wrong there. :facepalm:

rgl: I didn't try using `tpm_path` because my machine does not have a TPM. And even if it had one, it seems we can only have a single VM using the real TPM in 'passthru' mode, which make that mode of operation kinda useless for me.

I really just want to play with the emulated TPM. And this patch make that a possibility! :-)

JTarasovic:
crap. i totally biffed this. :facepalm:

Issue #354, opened 2015-04-30
Support separate storage pools for additional disks


pioro:
Hello,
I have two disk pools and I would like to configure one as a default pool for OS system disks and other for data disks. According to documentation if ":path" is defined a non-default pool should be used but this is not a case.

Ex.
:path => 'pool_d2'

==> si:  -- Disks:         vdb(qcow2,40G)
==> si:  -- Disk(vdb):     /VM/d1/pool/pool_d2

I was trying to set :absolute_path but it doesn't work either.

Libvirt Provider:
- absolute volume paths like '/VM/d2/pool/si_u01.qcow2' not yet supported

regards,
Marcin

pronix:
can you show your vagrantfile ?

pioro:
Hello,

This is my vagrant file with path

Vagrant.configure(2) do |config|
  # The most common configuration options are documented and commented below.
  # For a complete reference, please see the online documentation at
  # https://docs.vagrantup.com.

  # Every Vagrant development environment requires a box. You can search for
  # boxes at https://atlas.hashicorp.com/search.
  config.vm.box = "homeora65"

  config.vm.provider :libvirt do |libvirt|
	libvirt.storage_pool_name = "pool_d1"
        libvirt.memory=4096
        libvirt.cpus=2
        libvirt.storage :file, :size => '20G', :format => 'qcow2'
        libvirt.storage :file, :size => '5G', :format => 'qcow2', :path => "pool_d2"
        libvirt.storage :file, :size => '20G', :format => 'qcow2'
        libvirt.storage :file, :size => '20G', :format => 'qcow2'
  end

  config.vm.define :host do |host|
         host.vm.hostname = "targetkvm"
         host.vm.network :public_network, :dev => "br0", :mode => "bridge", :type => "bridge", :ip => "192.168.1.152"
         host.vm.network :private_network, :libvirt__network_name => "private", :ip => "10.10.10.152"
  end

end

purpleidea:
@pioro I'm not sure I understand your question.
The system disks come as part of the vagrant base box. So using the Vagrantfile can't add a "system" disk... That already comes with the box.

pioro:
Hi,

My image has only one system disk (vda) and that is OK. Wnat I need it to add additional disks like in my Vagrantfile (vdb, vdc, and so on). But all these new disk are added in same pool where system disk - pool_d1 in my case. I would like to use path to add disks in other disks pools.

Is it clear now ?

[...]

PR #1130, opened 2020-07-13
Support separate storage pools for additional disks

abbbi:
hi,

as discussed in #354, it would be very neat to be able to define a separate storage pool for additional disks. Pull request is a POC which does introduce pool setting for additional disks, example:

```
  config.vm.box = "generic/centos8"
        domain.storage :file, :size => '20G', :pool=>'default'
        domain.snapshot_pool_name='cache'
    end
end
```

that allows to place the virtual machines snapshot in the "cache" pool, but additional disks are beeing created in the "default" storage pool. Its not doing any checks so might not be merged yet, and i did not test any reload specific operations.


lib/vagrant-libvirt/action/create_domain.rb
```diff
@@ -167,6 +167,19 @@ def call(env)
 
             disk[:absolute_path] = storage_prefix + disk[:path]
 
+            if not disk[:pool].nil?
+              disk_pool_name = disk[:pool]
```
electrofelix, line 171: This is a local variable while below you use an instance variable, might be best to use local variable here only for the disk_pool_name. Mixing may result in some weird behaviour on subsequent loops:

the following prints what is expected
```
ruby -e 'def test
  value = "a different value"
  @value = "some value"
  puts @value
end

test()'
some value
```

This does not though and it could occur for a second or third disk where the pool isn't set.
```
ruby -e 'def test
  @value = "a different value"
  value = "some value"
  puts @value
end

test()'
a different value
```

lib/vagrant-libvirt/action/create_domain.rb
```diff
@@ -167,6 +167,19 @@ def call(env)
...
+              @logger.debug "Overriding pool name with: #{disk_pool_name}"
+              disk_storage_pool = env[:machine].provider.driver.connection.client.lookup_storage_pool_by_name(disk_pool_name)
+              raise Errors::NoStoragePool if disk_storage_pool.nil?
+              xml = Nokogiri::XML(disk_storage_pool.xml_desc)
+              disk_storage_prefix = xml.xpath('/pool/target/path').inner_text.to_s + '/'
```
electrofelix, line 176: Is it worth trying to make this a function to reuse above as well for retrieving the storage pool? Though I am wondering what happens in the the case of `config.qemu_use_session` being set and why it's guarding the storage pool retrieval above. The commit making the change to guard that just mentions qemu_use_session bug fixes, so I'm not sure if there is a potential issue

abbbi: yes, that would make sense, for POC i was copy/pasting the needed parts. What file would you place that function in?

electrofelix:
@abbbi I've made one small fix and I'm adding some spec tests in https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1158 that will help exercise this area a little, once that lands I'll fix up the conflict here and I think we're good to merge.

PR #1134, opened 2020-07-22
remove redundant setting of default volume/backingStore mode

eighthave:
According to the docs, for volumes and backingStores: "The mode defaults to 0600 when not provided.  Removing this setting here allows the pool configuration to set the default mode.

* https://gitlab.com/libvirt/libvirt/-/blob/v6.5.0/docs/formatstorage.html.in#L801
* https://libvirt.org/formatstorage.html


electrofelix:
Looks like RHEL/CentOS 6 would both be too old to have a sufficiently new enough libvirt that has this setting, however I'm thinking that anyone needing to run stuff in VMs on those releases would want various other pieces to be sufficiently newer anyway so unlikely to impact.

If someone is impacted, we can revert.

PR #1141, opened 2020-08-16
config: allow for setting URI from environment

electrofelix:
Users are allowed to set a LIBVIRT_DEFAULT_URI environment variable that
controls tools (i.e. virsh, virt-install, etc) that communicate with
libvirt. Let's allow for that mechanism to be used here.

### DIFF
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -622,14 +622,21 @@ module VagrantPlugins
         @qemu_env.merge!(options)
       end
 
-      # code to generate URI from a config moved out of the connect action
-      def _generate_uri
+      # code to generate URI from from either the LIBVIRT_URI environment
+      # variable or a config moved out of the connect action
+      def _generate_uri(qemu_use_session)
+
+        # If the LIBVIRT_DEFAULT_URI var is set, we'll use that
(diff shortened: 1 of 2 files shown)


voxik:
I welcome integration of this env variable and I think it should have no impact at all. The only thing we change is the [default](https://src.fedoraproject.org/rpms/vagrant-libvirt/blob/master/f/vagrant-libvirt-0.0.45-enable-qemu-session-by-default.patch), i.e. we enable the user session in contrast to upstream.

Nevertheless, with this env variable, we could set it somewhere in our Vagrant [binary](https://src.fedoraproject.org/rpms/vagrant/blob/master/f/binstub) instead of patching vagrant-libvirt. OTOH, the change would be in different package, which is not nice. For the moment, I think we'll keep patching, unless somebody has some convincing arguments ;)

electrofelix:
@voxik would dropping a file into /etc/profile.d to set `LIBVIRT_DEFAULT_URI` help ensure that someone starting virt-manager or virsh would see the same set of machines as available to vagrant?

voxik:
That seems to me to be too big hammer. Its fine to set some defaults for vagrant-libvirt somewhere, but dropping something into /etc/profile.d would change the whole system settings and I guess that libvirt maintainers would be against if not anyone else.

Issue #602, opened 2016-05-27
Multiple disks in Vagrant box / metadata.json


ruzickap:
Hello.

Is it possible to add the "Multiple disks in Vagrant box" into the vagrant-libvirt? For example the official Virtualbox Vagrant box for Ubuntu 16.04 contains two disks:

```
$ wget -q https://atlas.hashicorp.com/ubuntu/boxes/xenial64/versions/20160521.0.0/providers/virtualbox.box -O - | tar tvf -
-rw-r--r-- root/root     11013 2016-05-21 16:36 box.ovf
-rw-r--r-- root/root       652 2016-05-21 16:36 Vagrantfile
-rw-r--r-- root/root        31 2016-05-21 16:36 metadata.json
-rw-r--r-- root/root       314 2016-05-21 16:36 ubuntu-xenial-16.04-cloudimg.mf
-rw-r--r-- root/root 290942976 2016-05-21 16:36 ubuntu-xenial-16.04-cloudimg.vmdk
-rw-r--r-- root/root     72192 2016-05-21 16:36 ubuntu-xenial-16.04-cloudimg-configdrive.vmdk
```

Is it possible to extend the vagrant-libvirt plugin to be able to support more than one disk in box file? (probably some metadata.json enhancement will be needed)
### Steps to reproduce
1. See the output of

```
$ wget -q https://atlas.hashicorp.com/ubuntu/boxes/xenial64/versions/20160521.0.0/providers/virtualbox.box -O - | tar tvf -
-rw-r--r-- root/root     11013 2016-05-21 16:36 box.ovf
-rw-r--r-- root/root       652 2016-05-21 16:36 Vagrantfile
-rw-r--r-- root/root        31 2016-05-21 16:36 metadata.json
-rw-r--r-- root/root       314 2016-05-21 16:36 ubuntu-xenial-16.04-cloudimg.mf
-rw-r--r-- root/root 290942976 2016-05-21 16:36 ubuntu-xenial-16.04-cloudimg.vmdk
-rw-r--r-- root/root     72192 2016-05-21 16:36 ubuntu-xenial-16.04-cloudimg-configdrive.vmdk
```
### Expected behaviour

Vagrant libvirt boxes (metadata.json) should support more than one disk.
### Actual behaviour

The Vagrant libvirt boxes has always one disk.
### System configuration

**OS/Distro version:**:

```
$ head -1 /etc/fedora-release
Fedora release 23 (Twenty Three)
```

[...]

PR #1143, opened 2020-08-18
Allow to use many disks in vagrant box for libvirt provider (#602)

yamatoRT:
Hello,

I offer in this pull request for implement the issue #602.

With my modification, the `metadata.json` have new attributes :
  - `disk_number` whose is the number of disk in the box
  - `virtual_size_X` for each disk in the box (the disk 0 keep attribute `virtual_size`)

The name of disk file in the box must respect the format `box_X.img` expect for the first disk, it keep the name `box.img`.

Too I append unit tests.

I think my code may be improved, and I need your comments for it.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -18,6 +18,7 @@ jobs:
         - simple vm provision via shell
         - bring up with custom default prefix
         - bring up with second disk
+        - bring up with two disks
         - bring up with adjusted memory settings
         - bring up with adjusted cpu settings
         - ip is reachable with private network
@@ -95,6 +96,7 @@ jobs:
           end
         end
         EOF
+        bash ./tools/create_box_with_two_disks.sh $HOME/.vagrant.d/ ./bin/vagrant
         # run under libvirt group
         sg libvirt -c "bats -f '${{ matrix.test_name }}' ./tests/runtests.bats"
       env:
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -96,7 +96,6 @@ jobs:
           end
         end
         EOF
-        bash ./tools/create_box_with_two_disks.sh $HOME/.vagrant.d/ ./bin/vagrant
         # run under libvirt group
         sg libvirt -c "bats -f '${{ matrix.test_name }}' ./tests/runtests.bats"
       env:
--- a/lib/vagrant-libvirt/action/handle_box_image.rb
+++ b/lib/vagrant-libvirt/action/handle_box_image.rb
@@ -15,12 +15,62 @@ module VagrantPlugins
           @app = app
         end
 
-        def call(env)
-          # Verify box metadata for mandatory values.
-          #
+        def self.get_volume_name(env, index)
+          name = env[:machine].box.name.to_s.dup.gsub('/', '-VAGRANTSLASH-')
+          name << "_vagrant_box_image_#{
+          begin
+            env[:machine].box.version.to_s
+          rescue
+            ''
+          end}_#{index}.img"
+          return name
+        end
+
+        def self.get_virtual_size(env)
           # Virtual size has to be set for allocating space in storage pool.
           box_virtual_size = env[:machine].box.metadata['virtual_size']
           raise Errors::NoBoxVirtualSizeSet if box_virtual_size.nil?
+          return box_virtual_size
+        end
+
+        def self.get_default_box_image_path(index)
+          return index <= 0 ? 'box.img' : "box_#{index}.img"
+        end
+
+        def self.get_box_image_path(env, box_name)
+          return env[:machine].box.directory.join(box_name).to_s
+        end
+
+        def self.verify_virtual_size_in_disks(disks)
+          disks.each_with_index do |disk, index|
+            raise Errors::NoDiskVirtualSizeSet, disk_index:index if disk['virtual_size'].nil?
+          end
+        end
+
+        def call(env)
+          # Verify box metadata for mandatory values.
+          #
+          # Verify disk number
+          disks = env[:machine].box.metadata.fetch('disks', [])
+          if disks.empty?
+            disks.push({
+              'path' => HandleBoxImage.get_default_box_image_path(0),
+              'name' => HandleBoxImage.get_volume_name(env, 0),
+              'virtual_size' => HandleBoxImage.get_virtual_size(env),
+            })
+          end
+          HandleBoxImage.verify_virtual_size_in_disks(disks)
+
+          env[:box_volume_number] = disks.length()
+          env[:box_volumes] = Array.new(env[:box_volume_number]) {|i| {
+              :path => HandleBoxImage.get_box_image_path(
+                env,
+                disks[i].fetch('path', HandleBoxImage.get_default_box_image_path(i))
+              ),
+              :name => disks[i].fetch('name', HandleBoxImage.get_volume_name(env, i)),
+              :virtual_size => disks[i]['virtual_size']
+            }
+          }
 
           # Support qcow2 format only for now, but other formats with backing
           # store capability should be usable.
@@ -33,16 +83,13 @@ module VagrantPlugins
 
           # Get config options
           config = env[:machine].provider_config
-          box_image_file = env[:machine].box.directory.join('box.img').to_s
-          env[:box_volume_name] = env[:machine].box.name.to_s.dup.gsub('/', '-VAGRANTSLASH-')
-          env[:box_volume_name] << "_vagrant_box_image_#{
-          begin
-            env[:machine].box.version.to_s
-          rescue
-            ''
-          end}.img"
+          box_image_files = []
+          env[:box_volumes].each do |d|
+            box_image_files.push(d[:path])
+          end
 
           # Override box_virtual_size
+          box_virtual_size = env[:box_volumes][0][:virtual_size]
           if config.machine_virtual_size
             if config.machine_virtual_size < box_virtual_size
               # Warn that a virtual size less than the box metadata size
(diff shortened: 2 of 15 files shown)


electrofelix:
@yamatoRT I think the code looks mostly fine, need to build a box that has multiple disks and test. Only concern might be that in modifying the box format to use indices instead of separate elements per volume that supporting additional settings for individual volumes within the box format requires adding complexity each time instead to manage.

electrofelix:
I've had a look around and there doesn't seem to be any restrictions on formatting of the box metadata, as long as it's valid json and has the required keys, everything else is up to the provider.

Looking at one of the use cases that would benefit from multiple disks was the ability to run an ESXi and VCSA instances using vagrant-libvirt, the latter would required bundling 11 disks of different sizes.

I've also tried considering what happens around backwards compatibility vs being transparent about what is required for a box to work. If the format is transparent, then it also means that boxes built for newer versions of vagrant-libvirt will kind of start on older versions but then fail in weird ways because the additional disks aren't imported.

Considering dealing with that number along with the possibility of needing to set bus options differently for different disks, as well as making it easy to enable control of naming of the disk images, it seems like it will be easier to support:
```
{
    'disks' => [
        {
            'name' => 'disk1.img',
            'virtual_size' => 10,
            'format' => 'qcow2'
        },
        {
            'name' => 'disk2.img',
            'virtual_size' => 15,
            'format' => 'raw'
        },
        {
            'name' => 'disk3.img',
        }
    ]
```

I can certainly understand how it might be easier to have disks set to 3 and then modify the virtual size of only 2 and allow extending the existing format, but I wonder if it might just be better to support a new format if you want more than one disk? That way there is no confusion about whether the parameters without the index apply as defaults or just to the first disk, and it should be nice a clear what options are being applied to each disk?

The goal should be that we can translate from the old format to the new format when handling the box, while older versions of vagrant-libvirt should then fail with a more obvious manner when trying to load the box that has the "disks" parameter because the parameters expected will be missing?

Thoughts?

yamatoRT:
@electrofelix Hello, tanks for your comment,
I haven't problem with your proposal.

I propose this behavior :
- When `disk` is missing or empty, we have only one disk, named `box.img` and the virtual size must specified with `virtual_size`.
- When `disk` is used, we have many disks, and their names and virtual sizes must be specified for each disk. The old `virtual_size` parameter is ignored.

Do you agree with this behavior?

electrofelix:
@yamatoRT sorry, haven't been able to work on this project much for the last month.

I think that makes sense, essentially a new format and an old format.

We should try and ping some of the other projects to alert them and ask if the suggested new format makes sense to them

electrofelix:
@yamatoRT this looks nice, I'll probably need to read through it a couple more times, in the mean time if you get a chance to rebase onto the latest master which now includes some tests that will run some integration level tests in addition to the unit tests, that would be great.

I have 2-3 more minor fixes that I need to put together to release what has currently landed since the last release and then I'll send that out and look to land this one a couple of days later.

yamatoRT:
@electrofelix I rebase my branch onto the laster master.

yamatoRT:
@electrofelix I append new integration test with box with two disks.

electrofelix:
@yamatoRT I'm hoping to spend time on this next week, an integration test involving a box with two disks was the main thing that occured to me when thinking about what would be needed to be able to release this once merged.

The really good news is that the recent release doesn't appear to have resulted in any new issues being added (couple of existing ones still present, but they were there before), so fingers crossed this can land during the week and we can plan to release it a bit quicker that the gap since the previous release

.github/workflows/integration-tests.yml
```diff
@@ -95,6 +96,7 @@ jobs:
           end
         end
         EOF
+        bash ./tools/create_box_with_two_disks.sh $HOME/.vagrant.d/ ./bin/vagrant
```
electrofelix, line 99: Can this be placed into the test that requires it itself rather than as part of running for all tests? Or is there something makes it difficult to run there?

Spotted that it's also referenced there so I'm wondering if this can just be removed?

tests/runtests.bats
```diff
@@ -77,6 +80,21 @@ cleanup() {
   cleanup
 }
 
+@test "bring up with two disks" {
+  export VAGRANT_CWD=tests/two_disks
+  cleanup
+  run create_box_with_two_disks
```
electrofelix, line 86: It seems like you've already got the code being called here for creating the box with two disks, so can the other entries be removed?

tests/runtests.bats
```diff
@@ -15,6 +16,8 @@ setup_file() {
 
     export VAGRANT_HOME=${TEMPDIR}/.vagrant.d
     echo "# Using ${VAGRANT_HOME} for VAGRANT_HOME" >&3
+
+    bash "${SCRIPT_DIR}/../tools/create_box_with_two_disks.sh" "${VAGRANT_HOME}" "${VAGRANT_CMD}"
```
electrofelix, line 20: Do we need this? It's also getting called as part of the github workflow script in addition to being called below, the one below being just for the test that requires it which would be the preferred option until there is a published box that has two disks, though likely always useful to be able to prepare them in various forms.

lib/vagrant-libvirt/action/handle_box_image.rb
```diff
@@ -57,69 +104,73 @@ def call(env)
...
-                format_type: box_format,
-                owner: @storage_volume_uid,
-                group: @storage_volume_gid,
-                pool_name: config.storage_pool_name
-              )
-            rescue Fog::Errors::Error => e
-              raise Errors::FogCreateVolumeError,
-                    error_message: e.message
-            end
+            env[:box_volumes].each_index do |i|
```
electrofelix, line 114: I've been trying to decide if it might be better to move this entire block of code to a separate function, for the most part, given the starting point I don't think it makes much difference either way, just want to see first if you think it'll make any unit tests easier to write against this part or if it's unnecessary effort. I'm mostly leaning towards unnecessary at the moment.

lib/vagrant-libvirt/action/handle_box_image.rb
```diff
@@ -15,12 +15,62 @@ def initialize(app, _env)
...
+        def call(env)
+          # Verify box metadata for mandatory values.
+          #
+          # Verify disk number
+          disks = env[:machine].box.metadata.fetch('disks', [])
+          if disks.empty?
+            disks.push({
+              'path' => HandleBoxImage.get_default_box_image_path(0),
+              'name' => HandleBoxImage.get_volume_name(env, 0),
+              'virtual_size' => HandleBoxImage.get_virtual_size(env),
```
electrofelix, line 59: I can't comment on it properly, but looking further down I see a `format` entry in the metadata that I think needs to be added here as it should be associated with individual disks.

Thoughts on whether there should be a top level value that is a default or should assume a default value of `qcow2`? Either way, might be better to decided whether it is handled now to be settable for each individual disk and whether it is therefore required to be specified for each disk or has a default.

electrofelix (commented):
Some minor nits, a question on `format` and how best to handle it, I'm thinking it should be allowed to be set on a per disk basis as well, but perhaps we should consider qcow2 to be the default format going forward with the assumption that can be used to override if we support other formats. Finally just a small question on whether you think it might be beneficial to move some of the code being moved to a loop into a separate function, or if that isn't really worth it?

Otherwise I think it looks like just about ready

.github/workflows/integration-tests.yml
yamatoRT: I had a problem when I want create box. If I am calling script from the runtest.sh, I have error. I don't know why i have this error, I think it's due to `sg libvirt`. It is alone solution I found.
PS: What is `sg`?

tests/runtests.bats
yamatoRT: Yes i can remove it.

tests/runtests.bats
yamatoRT: When we run in local the runtest.sh, we use a new temp directory and the box doesn't exist. 
I think the most solution, it is send box into box repository.

.github/workflows/integration-tests.yml
electrofelix: `sg` allows executing a command as a different group id. Since in this case libvirt is being installed as part of the CI run, and as I want to ensure the tests run as a normal user to be sure they work, and therefore not run this command as root, it's necessary to ensure the user is in the correct group. However at this point I don't believe the shell used by the CI agent will have updated the user groups as that typically requires a login/logout, so to avoid that, I can just switch to explicitly using the group using sg and since the user is a member of the group, it behaves the same as though I've logged in.

Not entirely sure why it would fail, as any lack of permissions around creating file/directories as the user should cause it to fail.

yamatoRT: I upload box in vagrant cloud for the test : yamatoRT/tinycore-two-disks
It's simplify the code. I let the script for create box if need.

tests/runtests.bats
yamatoRT: I remove it.

lib/vagrant-libvirt/action/handle_box_image.rb
yamatoRT: I create a function what send the box image.

lib/vagrant-libvirt/action/handle_box_image.rb
yamatoRT: With the new function `send_box_image`, the `box_format` is a default value for each disks. If you want I may append the box format options for each disks.

electrofelix: No need, if we're using that as a default for all disk images it should be fine.

lib/vagrant-libvirt/action/handle_box_image.rb
```diff
@@ -31,18 +126,27 @@ def call(env)
...
 
+          env[:box_volume_number] = disks.length()
+          env[:box_volumes] = Array.new(env[:box_volume_number]) {|i| {
+              :path => HandleBoxImage.get_box_image_path(
+                env,
+                disks[i].fetch('path', HandleBoxImage.get_default_box_image_path(i))
+              ),
+              :name => disks[i].fetch('name', HandleBoxImage.get_volume_name(env, i)),
+              :virtual_size => disks[i]['virtual_size'],
+              :box_format => box_format
```
electrofelix, line 137: I think keep the attribute name as `format` e.g. `:format => box_format`. That way when adding support for different formats on individual disks, more likely to remember to use `:format => disks[i].fetch('format', box_format)`. I do see that will require adjusting the send_box_image and the tests, but I think it'll fit in better with the naming of all the other attributes.

electrofelix (commented):
One last change I think, along with a rebase. Not sure if you want to amend the existing commits or add this a change on top. Hoping to merge rather than squash the entire PR.

hswong3i:
@yamatoRT I had apply the PR to master but having some question.

I had create a Vagrant box as below:
```
  config.vm.box = "alvistack/ubuntu-20.04"
    libvirt.storage :file, :size => '20G'
  end
end
```

Once running with `vagrant package`, I am expecting that the result `package.box` should contain both root disk and the additional 20G disk, but only root disk exists.

Shall this PR support create box with both `vagrant package` and `tools/create_box.sh`?

hswong3i:
@electrofelix @yamatoRT I would like to suggest for a more generic handling for this PR, looking for your comment on #1256 ;-)

lib/vagrant-libvirt/action/handle_box_image.rb
yamatoRT: I rename `box_format` to `format`.

yamatoRT:
> @yamatoRT I had apply the PR to master but having some question.
> 
> I had create a Vagrant box as below:
> 
> ```
> Vagrant.configure("2") do |config|
>   config.vm.box = "alvistack/ubuntu-20.04"
>   config.vm.provider :libvirt do |libvirt|
>     libvirt.storage :file, :size => '20G'
>   end
> end
> ```
> 
> Once running with `vagrant package`, I am expecting that the result `package.box` should contain both root disk and the additional 20G disk, but only root disk exists.
> 
> Shall this PR support create box with both `vagrant package` and `tools/create_box.sh`?

I didn't know `vagrant package`. In this PR, we need only use `tools/create_box.sh` for create box with two disks. For `vagrant package` with multi disks, I would prefer create new issue for it.

Do you agree with that?

electrofelix:
@yamatoRT I've made some tidy ups and pushed this up as #1271 as I didn't have permission to update your fork.

electrofelix:
@yamatoRT I agree with the expectation that adding support to the package action should come after this, mostly because I believe we need to support being able to package in both the old and new format.

The work on using the libvirt XML may well be the long term future, but it too will need to through a few cycles and any PRs that add support will need to make it trivial to continue to use the older mechanisms as well as package up using the previous formats. Otherwise anyone using `vagrant package` will immediately be prevented from publishing boxes that can be consumed by anyone on an older version of vagrant-libvirt. And not all of the distros move along that quickly.

For that reason I'm not expecting that we'll be able to jump directly to that approach and will need to retain the ability to package up using the different formats until confident that enough of those using vagrant-libvirt will be on a new enough version to make the default packaging format one of the newer approaches.

electrofelix:
Based on some experimentation, and improvements to fetch the virtual size and format via qemu-img, going to make the disk path a required option so that there is always 1 field specified for the disks instead of the behaviour added here of inferring the disk path name based on the index number of the disk.

I suspect there will be a few more things to fix, such as the additional storage on top of a multi disk VM might not yet work, so I'll look to fix that, as well as implement the packaging part if you haven't gotten to it.

PR #1144, opened 2020-08-18
Use Vagrant::Util::TemplateRenderer instead of Erubis.

### LINKED ISSUE #1142: Erubis dependency
I want to get rid of Erubis from Fedora, because this project is dead upstream and on various places replaced by Erubi. Therefore I submitted hashicorp/vagrant/pull/11829 PR just to realize that vagrant-libvirt is also using Erubis, but without explicitly stating this. So there are several things to consider/do:

1. The easiest would be to specify the dependency. This should be done anyway, because vagrant-libvirt should keep working no matter what is Vagrant using (and what version of Vagrant is used actually). But that does not remove the reason why I am digging into this matter.
2. Replace the Erubi by Erubis. This should be possible and reasonably easy. If the dependency was specified, this can be done independently on Vagrant.
3. The best option would be to use the `Vagrant::Util::TemplateRenderer` and this idea is hanging in the air [for some while already](7ae9e03c5dee45ab194da4bca4802613878d5850). Unfortunately, this is not straight forward. The proof of concept, which passes test suite is down bellow, but unfortunately it fails in reality with errors such as ```undefined method `split_size_unit' for #<Vagrant::Util::TemplateRenderer:0x00007f9e70945690>```.

Then there are also other options such as:

1. Change the templates to not use instance variables and methods. It seems Vagrant does not use instance variables, which makes the matter a bit easier.
2. Use the [`Nokogiri::XML::Builder`](https://github.com/vagrant-libvirt/vagrant-libvirt/blob/d7debb68c634c80fbdae6e8bd4e3187581e80f8f/lib/vagrant-libvirt/util/erb_template.rb#L7)
3. Or of course the best option to improve on fog-libvirt (unless vagrant-libvirt wanted to get rid of the fog-libvirt dependency, the current mix is unfortunate).
4. Improve the `Vagrant::Util::TemplateRenderer` to (not)use binding.

Any thoughts?

---

Attempt to utilize `Vagrant::Util::TemplateRenderer`

~~~
$ git diff
diff --git a/lib/vagrant-libvirt/util/erb_template.rb b/lib/vagrant-libvirt/util/erb_template.rb
index 3bee45c..1d42a29 100644
--- a/lib/vagrant-libvirt/util/erb_template.rb
+++ b/lib/vagrant-libvirt/util/erb_template.rb
@@ -1,20 +1,17 @@
-require 'erubis'
-
 module VagrantPlugins
   module ProviderLibvirt
     module Util
       module ErbTemplate
         # TODO: remove and use nokogiri builder
-        # TODO: might be a chance to use vagrant template system according to https://github.com/mitchellh/vagrant/issues/3231
         def to_xml(template_name = nil, data = binding)
           erb = template_name || self.class.to_s.split('::').last.downcase
-          path = File.join(File.dirname(__FILE__), '..', 'templates',
-                           "#{erb}.xml.erb")
-          template = File.read(path)
+          path = File.join(File.dirname(__FILE__), '..', 'templates')
+          template = "#{erb}.xml"
 
-          # TODO: according to erubis documentation, we should rather use evaluate and forget about
-          # binding since the template may then change variables values
-          Erubis::Eruby.new(template, trim: true).result(data)
+          Vagrant::Util::TemplateRenderer.render_with(:render, template, template_root: path) do |renderer|
+            iv = data.eval ("instance_variables.collect {|i| [i, instance_variable_get(i.to_sym)]}")
+            iv.each {|k, v| renderer.instance_variable_set(k, v)}
+          end
         end
       end
     end
~~~

voxik:
Erubis is dead upstream, therefore it is good idea to avoid its usage.

Also, this allows to drop explicit dependency on Erubis, which is not specified anywhere and may cause issues when Vagrant changes its renderer.

Fixes #1142.

### DIFF
--- a/lib/vagrant-libvirt/util/erb_template.rb
+++ b/lib/vagrant-libvirt/util/erb_template.rb
@@ -1,20 +1,19 @@
-require 'erubis'
-
 module VagrantPlugins
   module ProviderLibvirt
     module Util
       module ErbTemplate
         # TODO: remove and use nokogiri builder
-        # TODO: might be a chance to use vagrant template system according to https://github.com/mitchellh/vagrant/issues/3231
         def to_xml(template_name = nil, data = binding)
           erb = template_name || self.class.to_s.split('::').last.downcase
-          path = File.join(File.dirname(__FILE__), '..', 'templates',
-                           "#{erb}.xml.erb")
-          template = File.read(path)
+          path = File.join(File.dirname(__FILE__), '..', 'templates')
+          template = "#{erb}.xml"
 
           # TODO: according to erubis documentation, we should rather use evaluate and forget about
           # binding since the template may then change variables values
-          Erubis::Eruby.new(template, trim: true).result(data)
+          Vagrant::Util::TemplateRenderer.render_with(:render, template, template_root: path) do |renderer|
+            iv = data.eval ("instance_variables.collect {|i| [i, instance_variable_get(i.to_sym)]}")
+            iv.each {|k, v| renderer.instance_variable_set(k, v)}
+          end
         end
       end
     end


voxik:
The only problem is, that not all templates are apparently covered by the test suite :thinking: I have tested the 'domain' and 'public_interface' templates.

electrofelix:
@voxik I have been expanding the unit tests and I've had a look that suggests there isn't anything in the other templates that would be impacted (such as a local function being used) so I think it should be safe. If this can wait until next week I can prioritize adding the missing unit tests this Friday, otherwise I think it should be safe enough to land.

voxik:
@electrofelix I'm going to add this patch into Fedora. It will land just in Rawhide, so if there is breakage, it won't hurt that much and it might get some additional testing. IOW take your time. For me is important to have link to the upstream PR with all the context and have a discussion then to have upstream release.

Issue #1155, opened 2020-10-23
Support parameters for openvswitch network configuration


abbbi:
Currently vagrant-libvirt only supports the option 

ovs: true

which adds the virtualport type='openvswitch'> flag to the created interface. With recent libvirt versions, and especially with
OVN, there may be more parameters required, one in particular is the option "interfaceid", which is used by libvirt to match the 
ports that are configured via libvirt into an OVN Network.

Lets have the following example:

1) System A acts as OVN controller node and has setup a virtual switch with DHCP enabled (northbound)
2) System B acts as libvirt node for vagrant boxes and is also an node in the OVN network.

Now, there are a few things to consider if attaching a virtual machine to this network:

1) The configuration must look as follows:

```
> <interface type='bridge'>
>       <mac address='c0:ff:ee:00:00:12'/>
>       <source bridge='br-int'/>
>       <virtualport type='openvswitch'>
>         <parameters interfaceid='00000000-0000-0000-0000-000000000001'/>
>       </virtualport>
>       <target dev='vnet1'/>
>       <model type='virtio'/>
>       <alias name='net1'/>
>       <address type='pci' domain='0x0000' bus='0x00' slot='0x06' function='0x0'/>
>     </interface>
> 
```

Where the mac address and the interfaceid must match the configured port in the OVN controllers northbound database.

During VM spinnup, libvirt takes care to attach the virtual port to the openvswitch bridge, it is listed in ovs-vsctl show output.

Beeing able to set the interfaceid parameter via vagrant config would be very neat, this way vagrant-libvirt could spin up boxes connected to an OVN network.

PR #1156, opened 2020-10-23
#1155 support setting ovs interface id

abbbi:
totally untested ..


electrofelix:
@abbbi I'm thinking might be worth merging, doing a release and dealing with any additional work on it in the future?

abbbi:
currently i dont have found any more parameters for the openvswitch configuration other than the interface id, might be worth adding more if libvirt adds more parameters, for now it should be enough..

electrofelix (approved):
lgtm

PR #1166, opened 2020-11-14
Add ability to use emulated tpm

### LINKED ISSUE #965: Add support for tpm_version

JTarasovic:
closes #965

qemu has supported tpm 2 and the ability to start
swtpm for a while. this change just allows passing
through the necessary config.

```ruby
    libvirt.tpm_model = "tpm-crb"
    libvirt.tpm_type = "emulator"
    libvirt.tpm_version = "2.0"
  end
end
```


JTarasovic:
@rgl 

Sorry, I totally biffed the rebase and closed the previous PR (#1126). Removing the check on version that I added did fix that noisy warning and is working for me both with passthrough as well as using the emulator.

lib/vagrant-libvirt/action/start_domain.rb
```diff
@@ -181,13 +181,18 @@ def call(env)
               if config.tpm_path
                 raise Errors::FogCreateServerError, 'The TPM Path must be fully qualified' unless config.tpm_path[0].chr == '/'
 
+                if config.tpm_version
+                  raise Errors::FogCreateServerError, 'The TPM Version only works with the `emulator` backend type' unless config.tpm_type == 'emulator'
+                end
```
electrofelix, line 186: Reading the code I believe the intent is the following Vagrantfile would produce an error at this point:
```
    libvirt.tpm_model = "tpm-tls"
    libvirt.tpm_type = "passthrough"
    libvirt.tpm_version = "2.0"
  end
end
```
However I don't think that will happen with the current code?

I've added a rudimentary spec test for this action in https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1167, be worth seeing if you can test for this error being returned. Some of the settings below being added would be nice, but not too worried as that's no worse that the current scenario as none of the other settings around are being checked and we can build up in the future.

JTarasovic: I went back to the [libvirt documentation](https://libvirt.org/formatdomain.html#tpm-device) and created a lil script to iterate through the different permutations. The script, libvirt domain XML and results can be seen [here](https://gist.github.com/JTarasovic/1d1411008674ef62d479110cfad68719).

It doesn't really make sense to "pass through" a specific version - the device is either one version or the other - but libvirt starts the VM despite a version being specified so I'm not even sure that it makes sense to have a check. The only check that seems like it could be useful would be `model = 'tpm-crb' && version != '2.0'` which does fail to start. The error for that is pretty obvious though:

`Error while creating domain: Error saving the server: Call to virDomainDefineXML failed: unsupported configuration: Unsupported interface tpm-crb for TPM 1.2`

Does this seem reasonable?

- [ ] remove version check
- [ ] update README

electrofelix: I think the problem here is the existing code only supports realigning config changes if `tpm_path` was set, however with the planned code changes the domain should be created with tpm settings if either `tpm_path` or `tpm_version` are set and thus likely reasonable for users to expect if either of those are set, changing settings on the tpm in the Vagrantfile and halting => starting should result in modifications being picked up similar to if the path was changed. However I don't think that will happen here, with the version check just being one symptom of the problem.

I added a very rudimentary test for the start domain to make it easier to add some for the tpm part here as it should be possible to exercise the behaviour here for the tpm_version being set and some attributes being modified.

electrofelix: @JTarasovic some more spec tests have been added around the start domain as well some now emitting some debug messages when the domain is getting modified, so it might make it a little easier if you enable debug output and compare what happens if you enable the tpm_version config setting bring the VM up and then try halt, modify the tpm config, followed by an up and you'll see there are no updates to the domain made unless `config.tpm_path` is set.

Happy enough if the version check is removed if you can add some tests for the start domain to validate when the tpm settings should be updated based on config changes getting applied since the VM was originally created.

JTarasovic: yeah, I didn't fully have my head wrapped create vs start. You're 100% right that it previously wouldn't have changed the domain from passthrough -> emulated or 1.2 -> 2.0.

I've added tests for a number of scenarios. Not sure if the actual implementation is what you're looking for but it'll certainly be easier to refactor with the behavior defined a bit better.

JTarasovic: @electrofelix do those tests and implementation seem cromulent?

electrofelix (approved):
lgtm, thanks for taking the time to rework

rgl:
So nice! Thank you!

JTarasovic:
Thank you both, @rgl and @electrofelix!

PR #1171, opened 2020-11-20
Add travis jobs on ppc64le

manish364824:
Hi,
I had added ppc64le(Linux on Power) support on travis-ci in the branch and looks like its been successfully added. I believe it is ready for the final review and merge. The travis ci build logs can be verified from the link below.

https://travis-ci.com/github/manish364824/vagrant-libvirt/builds/202997840

Please have a look.

Regards,
Manish Kumar

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,5 +30,24 @@ jobs:
       env: VAGRANT_VERSION=v2.2.4
     - rvm: 2.6.6
       env: VAGRANT_VERSION=master
+      
+      
+#Power jobs added
+    - rvm: 2.2.10
+      arch: ppc64le
+      env: VAGRANT_VERSION=v2.0.4
+    - rvm: 2.3.5
+      arch: ppc64le
+      env: VAGRANT_VERSION=v2.1.5
+    - rvm: 2.4.10
+      arch: ppc64le
+      env: VAGRANT_VERSION=v2.2.4
+    - rvm: 2.6.6
+      arch: ppc64le
+      env: VAGRANT_VERSION=v2.2.4
+    - rvm: 2.6.6
+      arch: ppc64le
+      env: VAGRANT_VERSION=master
+
   allow_failures:
     - env: VAGRANT_VERSION=master


electrofelix:
@manish364824 thanks, could you elaborate on the benefits of testing the ppc64le architecture here?

I would have assumed that for the current unit tests there wouldn't be any differences. If the tests were to try standing up VMs as a more functional test it would pick up any incorrect assumptions as to features available on other architectures as well as mistakes around handling box images cross architecture (can a x86_64 box be run on ppc64le?).

manish364824:
@gerrith3, could you please look into this?

gerrith3:
@electrofelix so x86_64/amd64 is a set of architectural instructions which are distinct from ppc64le - the Power architecture now uses little endian processing the same way that x86_64/amd64 does but the instruction sets/machine code are quite different.  While you can sort of use qemu to run across architecture sets it isn't guaranteed to work, e.g. some differences in page sizes between ppc64le and x86_64 often cause issues.  Emulation has its place but there is nothing better than testing on actual hardware to find and fix any corner cases.  This kind of testing can find corner cases and issues on actual hardware that you simply can't catch with emulation.

electrofelix:
@gerrith3 are you aware that none of the unit tests exercise qemu at all? The calls never get as far as even libvirt, that's what I meant that there isn't much benefit with the current units. I guess some of calls that use nokogiri might hit native machine code as nokogiri does compile native extensions.

It would be good to sort out some functional tests that would exercise some of the functionality via libvirt/qemu, as that would catch a number of issues that get missed. There is another PR that looks to bring up some VMs as tests, and I'd hope maybe to move the existing unit tests to GitHub actions at which point that pr could land, then this would have an impact.

https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1091 is the pr if your interested in helping land that one first.

electrofelix:
@gerrith3 @manish364824 based on the above, I'm going to close this based on the assumption that it was thought this was exercising QEMU on ppc64le when at the moment the tests are really just exercising ruby code.

Issue #851, opened 2018-01-15
vagrant package  won't fork/clone/duplicate/generate a similar vagrant-up-able box/image.


ribamar-santarosa:
### Description

`vagrant package`  won't fork/clone/duplicate/generate a similar vagrant-up-able box/image.

### Vagrant version
````
vagrant -v
Vagrant 2.0.1
````

### Host operating system

````
lsb_release  -a
Distributor ID: Debian
Description:    Debian GNU/Linux 9.1 (stretch)
Release:        9.1
Codename:       stretch
````

### Guest operating system

````
vagrant ssh -c 'cat /etc/debian_version'
9.1
````

### Vagrantfile

fill vagrant image here:
````
Vagrant.configure("2") do |config|
  config.vm.box = "mynewbox"
end
````

### Debug output

````
vagrant package --output mynew.box
# Total bytes written: 1042032640 (994MiB, 26MiB/s)
# ==> default: Box created
# ==> default: You can now add the box:
# ==> default: vagrant box add mynew.box --name any_comfortable_name
````

Full output of `vagrant up` at:
https://gist.github.com/ribamar-santarosa/7b98edacaada8fc1892140a2aabca0e1 

Full output of `ssh` to the virtual machine at: 
https://gist.github.com/ribamar-santarosa/964a8feda145c3ab9602a6dad0aa4888

### Expected behavior

After running:
````
vagrant package --output mynew.box
````
I would expect the command bellow to finish someday: 
````
vagrant init mynewbox  && vagrant up --provider=libvirt 
````

### Actual behavior
````
vagrant init mynewbox  && vagrant up --provider=libvirt 
````
Hangs forever (on `up` part).

### Steps to reproduce
````
# Install vagrant. Check https://github.com/hashicorp/vagrant/issues/9285 / https://github.com/ribamar-santarosa/champselysees/tree/master/vagrant-debian-stretch

# run any image:

vagrant init debian/stretch64 
vagrant up --provider=libvirt
vagrant ssh -c pwd
vagrant ssh -c 'cat /etc/debian_version'
# 9.1

# clone/fork/duplicate that image:

vagrant package --output mynew.box
vagrant box add mynewbox mynew.box
# ==> default: Box created
# ==> default: You can now add the box:
# ==> default: vagrant box add mynew.box --name any_comfortable_name
vagrant box list
# ...
# mynewbox         (libvirt, 0)

vagrant destroy default
[...]

PR #1184, opened 2020-12-13
Fix Broken `/etc/machine-id` by `vagrant package` due to virt-sysprep's `customize` operation

hswong3i:
virt-sysprep by default enable the `customize` operation, which will regenerate the `/etc/machine-id` by the end of the operation. This cause all newly created VMs by `vagrant up` coming with idential `/etc/machine-id` and so receive conflicted IP from DHCP.

~~Moreover, virt-sysprep also by default enable the `ssh-hostkeys` operation. On Debian/Ubuntu `openssh-server` package will NOT regenerate the SSH host keys if they are missing BEFORE `systemctl start ssh`, therefore all newly created VMs by `vagrant up` will not able to complete the initial SSH connection phase.~~

This PR disable ~~both~~ `customize` ~~and `ssh-hostkeys`~~ as per https://github.com/vagrant-libvirt/vagrant-libvirt/issues/851#issuecomment-520799720 recommendation.

Fixes https://github.com/vagrant-libvirt/vagrant-libvirt/issues/851

### DIFF
--- a/lib/vagrant-libvirt/action/package_domain.rb
+++ b/lib/vagrant-libvirt/action/package_domain.rb
@@ -50,7 +50,7 @@ module VagrantPlugins
           # remove hw association with interface
           # working for centos with lvs default disks
           options = ENV.fetch('VAGRANT_LIBVIRT_VIRT_SYSPREP_OPTIONS', '')
-          operations = ENV.fetch('VAGRANT_LIBVIRT_VIRT_SYSPREP_OPERATIONS', 'defaults,-ssh-userdir')
+          operations = ENV.fetch('VAGRANT_LIBVIRT_VIRT_SYSPREP_OPERATIONS', 'defaults,-ssh-userdir,-customize')
           `virt-sysprep --no-logfile --operations #{operations} -a #{@tmp_img} #{options}`
           # add any user provided file
           extra = ''


electrofelix:
I see https://bugzilla.redhat.com/show_bug.cgi?id=1557042 is noted as a bug in virt-sysprep, so yes it makes sense to disable it by default until that is fixed.

However the ssh-hostkeys should be regenerated for new VMs rather than defaulting to those in the original image built. This works as expected for redhat/fedora based distributions (see https://www.redhat.com/archives/libguestfs/2016-July/msg00090.html) and disabling there is likely to result in unexpected behaviour. 

Suggestion at https://www.cyberciti.biz/faq/reset-a-kvm-clone-virtual-machines-with-virt-sysprep-on-linux/ is that it should be possible to run sysprep at package time to trigger the openssh-server to regenerate the host keys on the next boot.

Possibly if we use `--firstboot-command 'if command -v dpkg-reconfigure; then dpkg-reconfigure openssh-server; else :; fi'` by default that should be sufficient to ensure it applies to any debian derivative while ignored on others?

Will try to get to testing it locally to confirm

electrofelix:
Didn't quite work as expected, however I've located the solution used for the generic/debian* and generic/ubuntu* boxes that I think should include here https://github.com/lavabit/robox/blob/a0d6f21eb4cb971f8f2560d1ac59aad7baaf49c8/scripts/common/keys.sh#L22

Wrapping the printf call with a check if dpkg-reconfigure is on the system should ensure it only runs for the expected targets. That'll ensure the behaviour matches boxes packaged via other tooling as well.

electrofelix:
Running some tests locally suggests might be better creating a script to be bundled and then passing that to virt-sysprep via the `--run` option.

script contents so far
```
#!/bin/bash

if [[ ! -e /usr/sbin/dpkg-reconfigure ]]
then
    exit 0
fi

printf "@reboot root command bash -c 'export PATH=$PATH:/usr/sbin ; export DEBIAN_FRONTEND=noninteractive ; export DEBCONF_NONINTERACTIVE_SEEN=true ; /usr/sbin/dpkg-reconfigure openssh-server &>/dev/null ; /bin/systemctl restart ssh.service ; rm --force /etc/cron.d/keys'\n" > /etc/cron.d/keys
```

hswong3i:
Let me share some of my founding:
- [**CRITICAL**] `+customize` broken on EVERY OS IS a bug for upstream `virt-sysprep` logic
  - BTW before upstream get fixed we could override it with `-customize` manually
- [*MINOR*] `+ssh-hostkeys` broken on Debian/Ubuntu/etc `openssh-server` IS a bug for upstream distro package, but not related to `virt-sysprep` nor `vagrant-libvirt` logic (e.g. Fedora/openSUSE/etc both check and recreate SSH host keys before service start up)
  - BTW if we could keep it disable ALL OS could working happily (at least during `vagrant up`)

Solutions coming with:
- Patch upstream Debian/Ubuntu or `virt-sysprep`
  - Well it always works but really take time and out of our control...
- Patch our downstream logic specifically for Debian/Ubuntu
  - IMHO, this should be packager's responsibility but not ours (see below)
- With `-customize`
  - Working for all OS now, and could control by us (except Debian/Ubuntu/etc which are still broken due to missing SSH host keys)
- With `-customize` AND `-ssh-hostkeys`
  - Working for all OS now, and could control by us
  - If OS support (e.g. Fedora/openSUSE/etc), packager could remove the SSH host keys before running `vagrant packer`, which new keys will be generated automatically during next reboot
  - If OS NOT support (e.g. Debian/Ubuntu/etc), packer could keep the existing SSH host keys before running `vagrant packer`, so at least next `vagrant up` will not being blocked out

So back to this PR:
- Let's fix the CRITICAL with `-customize` so our default could always functioning?
- Let's fix the MINOR with `-ssh-hostkeys` so let the packager to choose how to handle it as per case by case?

hswong3i:
@electrofelix In case of fixing from Debian/Ubuntu logic, now I clone the handling from Fedora with some tweaking for my Ansible Role for SSHD:
  - https://github.com/alvistack/ansible-role-sshd/commit/8b896624f0c6cbee8fffbe7b9a9fa2bab190c2c7

I will try to submit a PR to Debian if possible; BTW, fixing in this way is much complicated than setting `-ssh-hostkeys`, and shouldn't be the duty for `vagrant-libvirt`, too :-S

electrofelix:
@hswong3i I don't expect the ssh hostkeys issue to be fixed anytime soon, the fact that it's been around since 2010 makes it unlikely that it's going to be properly fixed soon, and even if it is, it'll take time for it to filter through.

Setting `-ssh-hostkeys` makes the defaults less secure, and doesn't provide a decision point for people to be aware of it, hence I'm reluctant to bring that in because it impacts all libvirt boxes built via `vagrant package` not just those that suffer from this issue. As there is already existing examples that provide a solution that will fix the problem without impacting the security behaviour, it seems better to try and pick them up and utilize.

Using the following fixes the issue and results in ssh keys being regenerated on first boot for resulting machines:
```
cat <<EOF > sysprep.sh
#!/bin/sh

# repeat what machine-ids does in sysprep as this script needs to run via customize
# which has a bug resulting in the machine-ids being regenerated

if [ -f /etc/machine-id ]
then
    truncate --size=0 /etc/machine-id
fi

if [ -f /var/lib/dbus/machine-id ]
then
    truncate --size=0 /run/machine-id
fi

# for debian based systems ensure host keys regenerated on boot
if [ -e /usr/sbin/dpkg-reconfigure ]
then
    printf "@reboot root command bash -c 'export PATH=$PATH:/usr/sbin ; export DEBIAN_FRONTEND=noninteractive ; export DEBCONF_NONINTERACTIVE_SEEN=true ; /usr/sbin/dpkg-reconfigure openssh-server &>/dev/null ; /bin/systemctl restart ssh.service ; rm --force /etc/cron.d/keys'\n" > /etc/cron.d/keys
fi
EOF
export VAGRANT_LIBVIRT_VIRT_SYSPREP_OPTIONS="--run $(pwd)/sysprep.sh"
```

It appears that in order to run a script it is necessary to allow customize to run, so to allow the host keys to regenerate on first boot, need to leave the customize operation in place.

Making this the default behaviour by bundling the script and having `VAGRANT_LIBVIRT_VIRT_SYSPREP_OPTIONS` resolve to `--run <path-to-script>` as the default value will allow easy replacing as needed if there is an issue with the script for some distro while still providing a useful baseline for anyone needing to override.

hswong3i:
Take my Packer + Vagrant Builder + VirtualBox [example](https://github.com/alvistack/vagrant-gitlab-runner/blob/857742cb6128d9d204bcd361c4f532a8c12139c3/packer/virtualbox-13.6/packer.json#L13-L31):
```
  "post-processors": [
    {
      "inline": [
        "set -eu",
        "export _VMDK=\"$(sudo bash -c \"ls -1d /root/VirtualBox*/output-vagrant_source*/*.vmdk\")\"",
        "export _QCOW2=\"$(echo \"$_VMDK\" | sed \"s/.vmdk$/.qcow2/g\")\"",
        "export _UUID=\"$(sudo vboxmanage showhdinfo \"$_VMDK\" | egrep -e \"^UUID:\" | sed \"s/^UUID:\\s*//g\")\"",
        "sudo qemu-img convert -f vmdk -O qcow2 \"$_VMDK\" \"$_QCOW2\"",
        "sudo virt-sysprep --operations defaults,-ssh-userdir,-ssh-hostkeys,-customize -a \"$_QCOW2\"",
        "sudo virt-sparsify --in-place \"$_QCOW2\"",
        "sudo qemu-img convert -f qcow2 -O vmdk \"$_QCOW2\" \"$_VMDK\"",
        "sudo vboxmanage internalcommands sethduuid \"$_VMDK\" \"$_UUID\"",
        "cd output-vagrant",
        "sudo vagrant package source",
        "sudo vagrant destroy --force source"
      ],
      "type": "shell-local"
    }
  ],
```

So when compare with that for Packer + Vagrant Builder + Libvirt [example](https://github.com/alvistack/vagrant-gitlab-runner/blob/857742cb6128d9d204bcd361c4f532a8c12139c3/packer/libvirt-13.6/packer.json#L13-L26):
```
  "post-processors": [
    {
      "inline": [
        "set -eu",
        "export _QCOW2=$(sudo bash -c \"ls -1d /var/lib/libvirt/images/output-vagrant_source*.img\")",
        "sudo virt-sysprep --operations defaults,-ssh-userdir,-ssh-hostkeys,-customize -a \"$_QCOW2\"",
        "sudo virt-sparsify --in-place \"$_QCOW2\"",
        "cd output-vagrant",
        "sudo VAGRANT_LIBVIRT_VIRT_SYSPREP_OPERATIONS=\"defaults,-ssh-userdir,-ssh-hostkeys,-customize\" vagrant package source",
        "sudo vagrant destroy --force source"
      ],
      "type": "shell-local"
    }
  ],
```

You could found that by default `vagrant package` for VirtualBox is NOT expected to silently preprocess with `virt-sysprep` nor `virt-sparsify` so I will need to manually trigger them.

IMHO, vagrant-libvirt's `vagrant package` support is now too much then originally expectation (BTW, we just handle with `virt-sysprep` so I still need to manually trigger `virt-sparsify` for a smaller image size).

Rather than a complicated implementation, I would like to suggest [KISS](https://en.wikipedia.org/wiki/KISS_principle) as original.

electrofelix:
At least some of the common boxes for ubuntu/debian will ensure that new host keys are generated on first boot. I'm assuming the same behaviour happens for all fedora/suse/centos distributions and derivatives. This tends to be the more secure default.

Using the scripted approach solves the problem and avoids the risk of creating an additional problem for others using `vagrant package` with vagrant-libvirt.

Using the suggested approach changes the behaviour for all distros even those without the bug. It doesn't actually fix the issue, and if debian were to accept a fix to ensure the host keys were created in this case in the future, we would have to change behaviour again to adopt. There is no graceful switch over, where at some point we can just stop applying the change due to the fixes in debian because it's now another behaviour change.

This might appear KISS for some outcomes, but is not KISS for all cases nor do I think in the long term, because it is more likely to result in behaviour changes and difficult to ensure consistency for consumers.

hswong3i:
@electrofelix Shall we handle in this way:
- Keep this PR focus on original `-customize` related issue, so could make everyone happy ASAP (PR updated with force push)
- Temporarily I had a workaround for Debian/Ubuntu by my [Ansible Role for SSHD](https://github.com/alvistack/ansible-role-sshd/commit/8b896624f0c6cbee8fffbe7b9a9fa2bab190c2c7), so running `VAGRANT_LIBVIRT_VIRT_SYSPREP_OPERATIONS="defaults,-ssh-userdir,-customize" vagrant package` is already good enough for me
- I am planning to fine tune above patch and submit it to Debian upstream, so someday (well yes, "someday"...) that could be get fixed for every Debian related distro

Would you mind to get this PR fixed first, therefore you could focus on your proposal about fixing `-ssh-hostkeys` related with another PR?

electrofelix:
Agreed, it's a shame it's not just a simple setting that would work and be consistent for everything, so even if we have to remove this change when landing a subsequent fix for the ssh host keys, the end effect for those consuming will still be moving in the correct direction

hswong3i:
Ok patch for Debian/Ubuntu now goes to https://salsa.debian.org/ssh-team/openssh/-/merge_requests/11, let's see if we could fix it from upstream ;-)

cjwatson:
https://factorable.net/weakkeys12.extended.pdf explains why doing this during the boot process is risky; this is why Debian's openssh packaging specifically avoids generating host keys at boot.  Can vagrant generate host keys externally and provision them into the new box?  That sort of approach is generally much safer, since systems that have been running for a while tend to have much better entropy available.

hswong3i:
@cjwatson please kindly correct me: shall the uniqueness of SSH host keys based on the uniqueness of `/var/lib/systemd/random-seed` and so [systemd-random-seed.service](https://www.freedesktop.org/software/systemd/man/systemd-random-seed.service.html)?

Some reference:
  - https://manpages.debian.org/jessie/libguestfs-tools/virt-sysprep.1.en.html#RANDOM_SEED
  - https://pagure.io/fedora-kickstarts/blob/a7f91775/f/fedora-cloud-base.ks#_95-99
  - https://github.com/lavabit/robox/blob/693b8fe/scripts/ubuntu2004/cleanup.sh#L22-L23
  - https://github.com/rgl/ubuntu-vagrant/blob/c5213b0/provision.sh#L78-L94
  
IMHO the responsibility of reset `/var/lib/systemd/random-seed` in order to secure the uniqueness of SSH host keys, should belongs to the image packager, but not the `ssh.service` itself:
  - **Remove SSH host keys before packaging, but NOT regenerate it during next reboot**: This is current Debian/Ubuntu default behavior, which result as a broken `vagrant up` unless some additional manual hack (e.g. using [`virt-sysprep --firstboot`](https://manpages.debian.org/jessie/libguestfs-tools/virt-sysprep.1.en.html#FIRSTBOOT_VS_SCRIPT) and some [hacking script](https://github.com/DanHam/packer-virt-sysprep/blob/055c069/sysprep-op-ssh-hostkeys.sh#L32-L103)
  - **Remove both SSH host keys AND `/var/lib/systemd/random-seed` before packaging, AND regenerate it during next reboot**: This make sure the uniqueness of SSH host keys for each newly provisioned machine, where sshd also functioning as normal without additional hacking

>  Can vagrant generate host keys externally and provision them into the new box?

`vagrant up` could having some [provisioning logic](https://www.vagrantup.com/docs/provisioning), but this all comes AFTER initial successful ssh connection. If the sshd is down by default (due to the missing SSH host keys), we have no way to alter it.

> That sort of approach is generally much safer, since systems that have been running for a while tend to have much better entropy available.

This works ONLY IF we have a VM box in shutdown status, therefore `virt-sysprep` could alter its FS directly. BTW, this is not how `vagrant up` works: the above provision logic is targeting on a live system, but not for shutdown status...

Issue #921, opened 2018-09-20
vagrant up hangs on reconnecting with new ssh key (remote libvirt server)


binaryronin:
### Steps to reproduce
1. set up remote SSH connection to libvirt server
2. vagrant up

### Expected behaviour
Vagrant should reconnect to the machine using the new SSH key and then carry on.

### Actual behaviour
Vagrant hangs on this step and never reconnects to the machine.

### System configuration
**OS/Distro version:**: Arch Linux on both workstation and server

**Libvirt version:** libvirt-4.6.0-3

**Output of `vagrant version; vagrant plugin list`:**
```
$ vagrant version              
Installed Version: 2.1.5
Latest Version: 2.1.5
 
You're running an up-to-date version of Vagrant!

$ vagrant plugin list  
pkg-config (1.3.1, global)
vagrant-libvirt (0.0.43, global)
```

**Output of `VAGRANT_LOG=debug vagrant ... --provider=libvirt`**
```
DEBUG ssh: == Net-SSH connection debug-level log START ==
DEBUG ssh: D, [2018-09-20T08:59:19.027464 #6638] DEBUG -- net.ssh.transport.session[3fd9fc3f8c78]: establishing connection to 192.168.121.90:22 through proxy
D, [2018-09-20T08:59:19.261438 #6638] DEBUG -- net.ssh.transport.session[3fd9fc3f8c78]: connection established
I, [2018-09-20T08:59:19.261734 #6638]  INFO -- net.ssh.transport.server_version[3fd9fc401ef4]: negotiating protocol version
D, [2018-09-20T08:59:19.261835 #6638] DEBUG -- net.ssh.transport.server_version[3fd9fc401ef4]: local is `SSH-2.0-Ruby/Net::SSH_5.0.2 x86_64-linux'
D, [2018-09-20T08:59:19.262048 #6638] DEBUG -- net.ssh.transport.server_version[3fd9fc401ef4]: remote is `SSH-2.0-OpenSSH_7.4'
I, [2018-09-20T08:59:19.262570 #6638]  INFO -- net.ssh.transport.algorithms[3fd9fc4011e8]: sending KEXINIT
D, [2018-09-20T08:59:19.263264 #6638] DEBUG -- io[3fd9fc3f83e0]: queueing packet nr 0 type 20 len 1156
D, [2018-09-20T08:59:19.263445 #6638] DEBUG -- io[3fd9fc3f83e0]: sent 1160 bytes
[...]

PR #1222, opened 2021-03-17
Use OpenSSH functionality instead of netcat for proxy_command

electrofelix:
Solves vagrant not detecting end of ssh connection when a proxy is 
used, described in #921

Allows ssh argument population to append arguments only when 
respective values are not nil, avoiding creation of invalid 
proxy_command when ssh-agent or ssh config based connections are used


electrofelix:
crap, I thought squash merge would keep the commit author, but it uses the PR author

PR #1225, opened 2021-03-17
auto enable ssh connection

electrofelix:
- Enable ssh on keyfile set and extract other settings
- Set connect via ssh and keyfile automatically

### DIFF
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -711,21 +711,26 @@ module VagrantPlugins
           uri = 'qemu' # use QEMU uri for KVM domain type
         end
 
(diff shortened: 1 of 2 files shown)


ElArtista:
Just built this (along with the little bugfix from latest master) and tested it locally, everything seems good!

electrofelix:
@ElArtista Split this up slightly to separate the extraction call from being included here, and instead just focus on enabling the connect_via_ssh if the ssh key file is explicitly provided. I think this becomes just a tidy up now and can be considered separately in the future whether it's a good idea, instead of bundling other fixes in as well.

#1230 should have solved most of the issues around the keyfile being set when not valid.

PR #1236, opened 2021-03-22
Fix interaction with reload plugin

h0st:
This is a first pull request for this PR (https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1038), which was created by another developer but was not divided into several PRs.

Goal: Fix interaction with reload plugin

Original PR -- https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1038

### DIFF
--- a/lib/vagrant-libvirt/action/halt_domain.rb
+++ b/lib/vagrant-libvirt/action/halt_domain.rb
@@ -13,22 +13,30 @@ module VagrantPlugins
         def call(env)
           env[:ui].info(I18n.t('vagrant_libvirt.halt_domain'))
 
+          timeout = env[:machine].config.vm.graceful_halt_timeout
           domain = env[:machine].provider.driver.connection.servers.get(env[:machine].id.to_s)
           raise Errors::NoDomainError if domain.nil?
 
           begin
-            env[:machine].guest.capability(:halt)
-          rescue
-            @logger.info('Trying Libvirt graceful shutdown.')
-            domain.shutdown
-          end
+            Timeout.timeout(timeout) do
+              begin
+                env[:machine].guest.capability(:halt)
+              rescue Timeout::Error
+                raise
+              rescue
+                @logger.info('Trying Libvirt graceful shutdown.')
+                # Read domain object again
+                dom = env[:machine].provider.driver.connection.servers.get(env[:machine].id.to_s)
+                if dom.state.to_s == 'running'
+                  dom.shutdown
+                end
+              end
 
-
-          begin
-            domain.wait_for(30) do
-              !ready?
+              domain.wait_for(timeout) do
+                !ready?
+              end
             end
-          rescue Fog::Errors::TimeoutError
+          rescue Timeout::Error
             @logger.info('VM is still running. Calling force poweroff.')
             domain.poweroff
           end


lib/vagrant-libvirt/action/halt_domain.rb
```diff
@@ -13,19 +13,28 @@ def initialize(app, _env)
         def call(env)
           env[:ui].info(I18n.t('vagrant_libvirt.halt_domain'))
 
+          timeout = env[:machine].config.vm.graceful_halt_timeout
```
electrofelix, line 16: I would assume looking at this PR, that part of the problem for the reload plugin is taking too long for the halt to be performed?

Looking at the below, the max timeout allowed could be as long as double the specified because it's used twice.

In the first case if halt fails to succeed in the give timeout, it will attempt to access the domain and perform a shutdown via libvirt, it will then wait the same entire timeout, which exceeded will then perform a poweroff.

My guess would be that the intention of the rescue wasn't to catch a timeout to the halt capability, but rather in case where there was an error attempting to use the guest capability to perform a halt (console not available for some reason or ssh disabled), attempt the shutdown using libvirt to send a signal into the VM to initiate a graceful power off.

Given that, I'm thinking that the graceful halt timeout should be around both blocks and be calling the `domain.poweroff` in case of rescue.

I'll have to look at the `domain.wait_for(timeout)` below as whether that needs to be provided the remaining time to wait for the machine to be shutdown according to libvirt assuming the initial call to `env[:machine].guest.capability(:halt)` or `dom.shutdown` took some time before they returned.

electrofelix: As an example I think the following may be more correct:
```
          begin
            Timeout.timeout(timeout) do
              begin
                env[:machine].guest.capability(:halt)
              rescue Timeout::Error
                raise
              rescue
                @logger.info('Trying Libvirt graceful shutdown.')
                # domain.shutdown
                # Read domain object again
                dom = env[:machine].provider.driver.connection.servers.get(env[:machine].id.to_s)
                if dom.state.to_s == 'running'
                  dom.shutdown
                end
              end

              begin
                # this will always wait a little longer than the outer timeout so
                # failing to complete should always trigger a Timeout::Error instead
                # of a Fog::Errors::TimeoutError
                domain.wait_for(timeout) do
                  !ready?
                end
              end
            end
          rescue Timeout::Error
            @logger.info('VM is still running. Calling force poweroff.')
            domain.poweroff
          end
```

Can always add a rescue for the `Fog::Errors::TimeoutError` and re-raise a `Timeout::Error` in the block calling `domain.wait_for(timeout)` to cover all posibilities.

electrofelix:
@h0st have you had a chance to read my comment? Just wondering if it makes sense or if I'm misunderstanding something?

h0st:
@electrofelix yes sure, i had a chance but i don't have time to check your suggestion and test it.

Regarding your first questions, I guess it is the inability to allow some machines that need longer a longer timeout. 
I observed that plugin waits some period of time while the domain shutdown (halt) and then it fails with a timeout error. I can provide you logs if necessary.

lib/vagrant-libvirt/action/halt_domain.rb
h0st: Thank you, yes, i think it will be more correct! I will apply this some later.

Issue #1238, opened 2021-03-29
Please consider removing the contextual_proc gem


dcermak:
The latest release of vagrant-libvirt added the contextual_proc gem as a test dependency. Unfortunately this gem cannot be included in a Linux distribution, because it lacks a license (that makes it effectively proprietary and legally unfit for inclusion).

Could it please be removed? I see that it is only used in two places in the tests and the code of the gem is so simple, that I think it could be re-implemented in vagrant-libvirt.

I can try to send a PR, if you would accept it?

electrofelix:
@dcermak thanks, I had completely miss that when I was pulling it in for some tests.

Please do, I'll ask the upstream project if they'll consider adding an appropriate license for others to consume.

dcermak:
> @dcermak thanks, I had completely miss that when I was pulling it in for some tests.

I'll give it a try. Could you assign this issue to me please?

> Please do, I'll ask the upstream project if they'll consider adding an appropriate license for others to consume.

You can try, but upstream hasn't had any activity since 2016, so I wouldn't get my hopes up :-(

electrofelix:
I've emailed the two authors directly, as they appear to have some activity. Given the original code was put together to answer a question on stackoverflow, I'm guessing it just is likely an oversight, as the original code from there would be covered by https://creativecommons.org/licenses/by-sa/3.0/. They may just not have realized that publishing the code via github would require explicitly stating the license.

PR #1240, opened 2021-04-01
Add a simple Proc extension to support bindings

dcermak:
This removes the dependency on the contextual_proc gem

This fixes https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1238


electrofelix (approved):
lgtm, thanks

Issue #1198, opened 2021-01-05
Unable to install another plugin into  Docker based Installation


pgillich:
### Steps to reproduce
1. Define alias for Docker based Installation, described on README.md
2. Try to install another plugin locally (not into the container), for example:
`vagrant plugin install --local vagrant-mutate`
3. The result is an error, including (see the whole error message below):
`ERROR: Failed to build gem native extension.`

### Expected behaviour
```
$ vagrant plugin install --local vagrant-mutate
Starting with UID: 1000, GID: 1001
Installing the 'vagrant-mutate' plugin. This can take a few minutes...
Fetching racc-1.5.2.gem
Building native extensions. This could take a while...
Fetching nokogiri-1.11.0-x86_64-linux.gem
Fetching excon-0.78.1.gem
Fetching formatador-0.2.5.gem
Fetching fog-core-2.2.3.gem
Fetching fog-json-1.2.0.gem
Fetching fog-xml-0.1.3.gem
Fetching ruby-libvirt-0.7.1.gem
Building native extensions. This could take a while...
Fetching fog-libvirt-0.7.0.gem
Fetching vagrant-mutate-1.2.0.gem
Installed the plugin 'vagrant-mutate (1.2.0)'!

$ vagrant plugin install --local vagrant-hostmanager
Starting with UID: 1000, GID: 1001
Installing the 'vagrant-hostmanager' plugin. This can take a few minutes...
Building native extensions. This could take a while...
Building native extensions. This could take a while...
Fetching vagrant-hostmanager-1.8.9.gem
Installed the plugin 'vagrant-hostmanager (1.8.9)'!

$ ls .vagrant/plugins/gems/2.6.6/gems/ | grep vagrant
vagrant-hostmanager-1.8.9
vagrant-mutate-1.2.0
```

### Actual behaviour
```sh
$ vagrant plugin install --local vagrant-mutate
Starting with UID: 1000, GID: 1001
Installing the 'vagrant-mutate' plugin. This can take a few minutes...
Fetching excon-0.78.1.gem
Fetching formatador-0.2.5.gem
Fetching fog-core-2.2.3.gem
Fetching fog-json-1.2.0.gem
Fetching racc-1.5.2.gem
Building native extensions. This could take a while...
Vagrant failed to properly resolve required dependencies. These
errors can commonly be caused by misconfigured plugin installations                                                                                                                                                                   
or transient network issues. The reported error is:                                                                                                                                                                                   
[...]

PR #1246, opened 2021-04-07
Default to build and publish extendible image

electrofelix:
Switch to publishing the image containing all of the libraries and build
dependencies needed to support the plugin being reinstalled as that is
the default behaviour when vagrant installs a new plugin.

This should allow the default image to be extended with additional
plugins and if needed, the authors of the new image may follow and use
the slim version as a base to provide a reduced size docker image.

Fixes: #1198


electrofelix:
Looks like I need to rework the entrypoint before this is complete, but this should be the first step.

PR #1247, opened 2021-04-08
Add ability to manage bridge STP and delay parameters

h0st:
Second part of PR #1038 - ability to manage Bridge STP and Delay parameters

Follows defaults defined at https://libvirt.org/formatnetwork.html#elementsConnect

### DIFF
--- a/lib/vagrant-libvirt/action/create_networks.rb
+++ b/lib/vagrant-libvirt/action/create_networks.rb
@@ -297,6 +297,9 @@ module VagrantPlugins
 
           @network_ipv6_address = @interface_network[:ipv6_address]
           @network_ipv6_prefix = @interface_network[:ipv6_prefix]
+          
+          @network_bridge_stp = @options[:bridge_stp].nil? ? 'on' : @options[:bridge_stp] ? 'on' : 'off'
+          @network_bridge_delay = @options[:bridge_delay] ? @options[:bridge_delay] : 0
 
           @network_forward_mode = @options[:forward_mode]
           if @options[:forward_device]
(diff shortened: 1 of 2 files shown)


lib/vagrant-libvirt/action/create_networks.rb
```diff
@@ -297,6 +297,9 @@ def create_private_network(env)
 
           @network_ipv6_address = @interface_network[:ipv6_address]
           @network_ipv6_prefix = @interface_network[:ipv6_prefix]
+          
+          @network_bridge_stp = @options[:bridge_stp].nil? ? 'on' : @options[:bridge_stp] ? 'on' : 'off'
```
electrofelix, line 301: Small improvement, I think you can adjust this to perform the second check as part of the initial expression:
```
@network_bridge_stp = options[:bridge_stp].nil? || options[:bridge_stp] ? 'on' : 'off'
```

I was wondering for a second if use of `@option` would impact setting this on a per device and then I realised the existing code is setting this instance var on each loop through, that must be confusing for people trying to fix this.. really must get to the TODO at some point

electrofelix (commented):
If you get a chance to tweak the ternary expression can merge afterwards, otherwise will look to adjust it and merge once I have a couple of minutes to adjust it.

lib/vagrant-libvirt/action/create_networks.rb
h0st: Thanks for improvement, i've fixes it. 
I did not fully understand about TODO statement, could you please clarify? Do i need to add here some comment?

electrofelix: There isn't anything for you to do, it's more a talk to myself about it so that I'll hopefully remember there is a note buried in this code that all the functions should be updated at some point to avoid the need for acting on an instance variable

electrofelix (approved):
lgtm

lib/vagrant-libvirt/action/create_networks.rb
h0st: OK, thanks for explanation! Yes, it is really helpful to have comments about current changes.

PR #1248, opened 2021-04-08
Add ability to define multiple private networks

h0st:
Ability to define multiple private networks

Originally proposed as part of #1038.

May be some overlap with #1249.

### DIFF
--- a/lib/vagrant-libvirt/action/create_networks.rb
+++ b/lib/vagrant-libvirt/action/create_networks.rb
@@ -54,6 +54,8 @@ module VagrantPlugins
                 env[:machine].provider.driver.connection.client
               )
 
+              current_network = @available_networks.detect { |network| network[:name] == @options[:network_name] }
+
               # Prepare a hash describing network for this specific interface.
               @interface_network = {
                 name:             nil,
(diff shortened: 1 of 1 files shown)


lib/vagrant-libvirt/action/create_networks.rb
```diff
@@ -255,7 +257,9 @@ def handle_dhcp_private_network(env)
 
           # Do we need to create new network?
           unless @interface_network[:created]
-            @interface_network[:name] = 'vagrant-private-dhcp'
+            @interface_network[:name] = @options[:libvirt__network_name] ?
+                                        @options[:libvirt__network_name] :
```
electrofelix, line 261: I think these should be using `@options[:network_name]` because there is a function helper from vagrant called `scoped_hash_override` called at https://github.com/vagrant-libvirt/vagrant-libvirt/blob/master/lib/vagrant-libvirt/util/network_util.rb#L101 where it helps populate the options to override any settings such as `:network_name` with `:<provider>__network_name` for all options that are defined using `:<provider>__<option>`. So there shouldn't be any need in this file to refer to an option using `libvirt__` as it should already be converted.

I've mentioned it elsewhere, but the networking code isn't well tested, so I can't be 100% sure about this as I do need to add some tests that exercise this code and the networking util helper being used to be fully sure. So I'm basing this purely on my current understanding of the code.

h0st: Ah ok sure, i got it. Fixed! 
Yes, you told about this in a PR #1038.

PR #1249, opened 2021-04-08
Auto configuration for tunnel interfaces

h0st:
Forth part of PR #1038 -- Enable auto configuration for tunnel interfaces

### DIFF
--- a/lib/vagrant-libvirt/action/create_network_interfaces.rb
+++ b/lib/vagrant-libvirt/action/create_network_interfaces.rb
@@ -194,7 +194,7 @@ module VagrantPlugins
               # It's used for provisioning and it has to be available during provisioning,
               # ifdown command is not acceptable here.
               next if slot_number.zero?
-              next if options[:auto_config] === false
+              next if !options[:auto_config]
               @logger.debug "Configuring interface slot_number #{slot_number} options #{options}"
 
               network = {
@@ -213,9 +213,6 @@ module VagrantPlugins
               else
                 network[:type] = :dhcp
               end
-
-              # do not run configure_networks for tcp tunnel interfaces
-              next if options.fetch(:tunnel_type, nil)
 
               networks_to_configure << network
             end
--- a/lib/vagrant-libvirt/util/network_util.rb
+++ b/lib/vagrant-libvirt/util/network_util.rb
@@ -1,5 +1,14 @@
+require 'ipaddr'
 require 'nokogiri'
 require 'vagrant/util/network_ip'
+
+class IPAddr
+  def get_mask
+    if @addr
+      _to_string(@mask_addr)
+    end
+  end
+end
 
 module VagrantPlugins
   module ProviderLibvirt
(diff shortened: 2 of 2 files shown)


lib/vagrant-libvirt/util/network_util.rb
```diff
@@ -1,6 +1,14 @@
 require 'nokogiri'
 require 'vagrant/util/network_ip'
 
+class IPAddr
```
electrofelix, line 4: I think should require `ipaddr` for this file because we're explicitly opening the IPAddr class to modify it's behaviour here.

lib/vagrant-libvirt/util/network_util.rb
```diff
@@ -97,14 +105,18 @@ def configured_networks(env, logger)
             # store type in options
             # use default values if not already set
             options = {
-              iface_type:  type,
-              netmask:      '255.255.255.0',
+              iface_type:   type,
+              netmask:      options[:libvirt__network_address] ?
+                            IPAddr.new(options[:libvirt__network_address]).get_mask :
```
electrofelix, line 110: Looks like we can safely use `:network_address` here and similarly below `:network_name` instead of `:libvirt__network_address` and `:libvirt__network_name` as options is provided by `scoped_hash_override(original_options, :libvirt)`.

electrofelix (commented):
minor tweaks, one to help make things a little more obvious where IPAddr is coming from and the other to align with the support provided by the scoped_hash_override.

lib/vagrant-libvirt/util/network_util.rb
h0st: Oh, yes, sure. I missed this dependencie.  I'll fix it ASAP.

lib/vagrant-libvirt/action/create_network_interfaces.rb
```diff
@@ -194,7 +194,7 @@ def call(env)
               # It's used for provisioning and it has to be available during provisioning,
               # ifdown command is not acceptable here.
               next if slot_number.zero?
-              next if options[:auto_config] === false
```
electrofelix, line 197: I missed that this is actually incorrect, was picked up by the integration test for the private network luckily.

The reason is that it will skip when `auto_config` is falsey, when we only want to skip when explicitly disabled.

lib/vagrant-libvirt/action/create_network_interfaces.rb
```diff
@@ -194,7 +194,7 @@ def call(env)
...
+              next if !options[:auto_config]
```
electrofelix, line 197: I missed that this is actually incorrect, was picked up by the integration test for the private network luckily.

The reason is that it will skip when auto_config is falsey, when we only want to skip when explicitly disabled.

h0st: Ok, sure. If test fails, I could return a false value for this variable and we check it again. I've answered about this in a PR #1038

electrofelix: I've reverted this piece for master as I think the original is probably the best approach, unless there is something specific around the original behaviour that was broken

h0st: Ok good, I got it! Thanks!

Issue #1191, opened 2020-12-28
Docker based Installation does not work


pgillich:
### Steps to reproduce

Running the image by `docker run -it --rm ...`, as described on README.md, on a vanilla Linux (vagrant was never ran on the machine).

### Expected behaviour

vagrant starts without error, for example:

```
vagrant up master
Starting with UID: 1000, GID: 1001
Bringing machine 'master' up with 'libvirt' provider...
(...)
```

### Root causes

On a vanilla Linux install, the `~/.vagrant.d` does not exists, so the Docker will create it in root rights. If the `~/.vagrant.d` exists with root owner, the container cannot create `boxes`, `data`, `tmp` directories.

Another problem with the `docker run` command: the container cannot ssh to the VM, because the `--network host` is missing.

### Solutions

Extending the documentation by below comment:

Before first start of **Docker based Installation**, make sure, all needed directories exists and owned by you, for example:

```
mkdir -p ~/.vagrant.d/{boxes,data,tmp}
ls -la ~/.vagrant.d # All dirs must be owned by you
```

Adding `--network host` to the `docker run` command, for example:

```
docker run -it --rm --network host ...
```

### Improvement

The `docker run` command can be called by an alias, which may be added to `~/.bashrc`. Example:

```
alias vagrant='docker run -it --rm --network host -e LIBVIRT_DEFAULT_URI -v /var/run/libvirt/:/var/run/libvirt/ -v ~/.vagrant.d:/.vagrant.d -v $(pwd):$(pwd) -w $(pwd) vagrantlibvirt/vagrant-libvirt:latest vagrant'
```

### Actual behaviour

Instead of starting the vagrant, below output is printed out:

```
WARNING! Running as root, if this breaks, you get to keep both pieces
Starting with UID: 0, GID: 0
chown: invalid user: 'vagrant:users'
usermod: user 'vagrant' does not exist
```

After the user right issue is solved, vagrant, from the container, cannot ssh to the VM (cannot check the IP address), because it's not on the host network.

### System configuration

**OS/Distro version:** Lubuntu 20.04 (same on Lubuntu 18.04)

**Libvirt version:** 6.0.0-0ubuntu8.5

**Output of `vagrant version; vagrant plugin list`:**

```
Installed Version: 2.2.10
Latest Version: 2.2.14
```

```
vagrant-libvirt (0.3.0, global)
  - Version Constraint: 0.3.0
```

[...]

PR #1254, opened 2021-04-20
Fix missing vagrant user on docker based Installation(#1191)

ekeyme:
1.fix issue #1191
2.add symbolic link support on Docker based Installation doc

When run as root, vagrant user inside container will not be created, so we get #1191 error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,12 +133,13 @@ docker run -it --rm \
 It's possible to define an alias in `~/.bashrc`, for example:
 ```bash
 alias vagrant='
-  mkdir -p ~/.vagrant.d/{boxes,data,tmp}; \
+  VAGRANT_HOME=~/.vagrant.d
+  mkdir -p $VAGRANT_HOME/{boxes,data,tmp}; \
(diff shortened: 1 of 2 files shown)


electrofelix:
I don't think this is quite right, it's attempting to let the container continue as root, when the problem is that the correct uid/gid detection is incorrect.

The underlying issue is really https://github.com/vagrant-libvirt/vagrant-libvirt/blob/master/entrypoint.sh#L5 is looking at the wrong file to determine the correct uid/gid to allow adjustment of the local user to reflect the correct user and permissions.

Instead it should be looking at the working directory and searching for the base Vagrantfile and then use the uid/gid on that to determine the correct values. Once these are determined, if it turns out ~/.vagrant.d has the wrong user/group relative to the Vagrantfile should be able to set it to the correct values (not sure if that will take outside of the docker image or not as it's a mount point rather than contents?).

Can you look at switching to this approach instead?

electrofelix:
As I've implemented the more complete fix with #1320, going to close this.

PR #1267, opened 2021-05-07
Synchronize Snapshot Disks Storage File Permission

hswong3i:
From <https://github.com/fog/fog-libvirt/blob/39ead72/lib/fog/libvirt/models/compute/templates/volume.xml.erb> storage files are created with permission `<mode>0744</mode>`; BTW, we are missing corresponding parameters from `lib/vagrant-libvirt/action/create_domain_volume.rb`, therefore file created with permission `0600`.

Fixes #1275

### DIFF
--- a/lib/vagrant-libvirt/action/create_domain_volume.rb
+++ b/lib/vagrant-libvirt/action/create_domain_volume.rb
@@ -53,6 +53,7 @@ module VagrantPlugins
                     xml.permissions do
                       xml.owner storage_uid(env)
                       xml.group storage_gid(env)
+                      xml.mode '0744'
                       xml.label 'virt_image_t'
                     end
                   end
@@ -62,6 +63,7 @@ module VagrantPlugins
                     xml.permissions do
                       xml.owner storage_uid(env)
                       xml.group storage_gid(env)
+                      xml.mode '0744'
                       xml.label 'virt_image_t'
                     end
                   end
--- a/spec/unit/action/create_domain_volume_spec/one_disk_in_storage.xml
+++ b/spec/unit/action/create_domain_volume_spec/one_disk_in_storage.xml
@@ -6,6 +6,7 @@
     <permissions>
       <owner>0</owner>
       <group>0</group>
+      <mode>0744</mode>
       <label>virt_image_t</label>
     </permissions>
   </target>
@@ -15,6 +16,7 @@
     <permissions>
       <owner>0</owner>
       <group>0</group>
+      <mode>0744</mode>
       <label>virt_image_t</label>
     </permissions>
   </backingStore>
--- a/spec/unit/action/create_domain_volume_spec/three_disks_in_storage_disk_0.xml
+++ b/spec/unit/action/create_domain_volume_spec/three_disks_in_storage_disk_0.xml
@@ -6,6 +6,7 @@
     <permissions>
       <owner>0</owner>
       <group>0</group>
+      <mode>0744</mode>
       <label>virt_image_t</label>
     </permissions>
   </target>
@@ -15,6 +16,7 @@
     <permissions>
       <owner>0</owner>
       <group>0</group>
+      <mode>0744</mode>
       <label>virt_image_t</label>
     </permissions>
   </backingStore>
--- a/spec/unit/action/create_domain_volume_spec/three_disks_in_storage_disk_1.xml
+++ b/spec/unit/action/create_domain_volume_spec/three_disks_in_storage_disk_1.xml
@@ -6,6 +6,7 @@
     <permissions>
       <owner>0</owner>
       <group>0</group>
+      <mode>0744</mode>
       <label>virt_image_t</label>
     </permissions>
   </target>
@@ -15,6 +16,7 @@
     <permissions>
       <owner>0</owner>
       <group>0</group>
+      <mode>0744</mode>
       <label>virt_image_t</label>
     </permissions>
   </backingStore>
--- a/spec/unit/action/create_domain_volume_spec/three_disks_in_storage_disk_2.xml
+++ b/spec/unit/action/create_domain_volume_spec/three_disks_in_storage_disk_2.xml
@@ -6,6 +6,7 @@
     <permissions>
       <owner>0</owner>
       <group>0</group>
+      <mode>0744</mode>
       <label>virt_image_t</label>
     </permissions>
   </target>
@@ -15,6 +16,7 @@
     <permissions>
       <owner>0</owner>
       <group>0</group>
+      <mode>0744</mode>
       <label>virt_image_t</label>
     </permissions>
   </backingStore>


hswong3i:
@electrofelix this PR is just about incorrect storage file permission, therefore should be safe enough for a simple merge?

electrofelix:
@hswong3i I want to understand why it wasn't an issue before so I'll want to do some digging before landing the change. In one way it suggests that the virt-v2v won't work with remote systems, where vagrant-libvirt was connected to another host as it must be accessing the files directly instead of retrieving them via libvirt. But as I haven't looked into what's going on I didn't want to merge it just yet.

lib/vagrant-libvirt/action/create_domain_volume.rb
```diff
@@ -53,6 +53,7 @@ def call(env)
                     xml.permissions do
                       xml.owner storage_uid(env)
                       xml.group storage_gid(env)
+                      xml.mode '0744'
```
electrofelix, line 56: This suggests that maybe virt-v2v-copy-to-local should be used first to retrieve the files from libvirt, as making things readable by other users will likely be considered a security risk.

lib/vagrant-libvirt/action/create_domain.rb
```diff
@@ -198,7 +198,7 @@ def call(env)
                 path: disk[:absolute_path],
                 capacity: disk[:size],
                 owner: storage_uid(env),
-                group: storage_uid(env),
+                group: storage_gid(env),
```
electrofelix, line 201: This looks to be a very straight forward fix, can you look at just sending this part, because I think changing the file modes will cause concern over security.

electrofelix (commented):
I'd suggest looking at virt-v2v-copy-to-local or consider building on the existing code to retrieve the disk images (if that is quicker, which it might well be according to virt-v2v-copy-to-local) to a local path and then use virt-v2v as needed subsequently.

That should avoid the issues around the file modes and also avoid bumping into security concerns as well.

hswong3i:
> That should avoid the issues around the file modes and also avoid bumping into security concerns as well.

Since `<mode>0744</mode>` is now hard-coded from https://github.com/fog/fog-libvirt/blob/39ead72/lib/fog/libvirt/models/compute/templates/volume.xml.erb, our backing image file and additional disks are now therefore created as `<mode>0744</mode>`. This PR basically synchronize the behavior for snapshot image file that we manually created with XML, too.

From security point of view, Vagrant is primarily target for re-deployable development environment but not production, therefore:
* If running QEMU as `root:root`: `vagrant package` should always running as root or sudo (well, we are already in the risky way therefore no additional risk generated...).
* If running QEMU as `root:libvirt`: `/var/lib/libvirt/images` should also owned by `root:libvirt`, therefore non-root user within `libvirt` group should able to run `vagrant package` without sudo (i.e. target for this PR).

`virt-v2v-copy-to-local` could be a solution, unless we also hope to support remote instance packaging to local. Anyway, could this be solved by #1268, afterward?

electrofelix:
Having the backing file, which is never modified by the VM be readable by other users outside of the libvirt process isn't great, but also isn't as big an issue. Having the files containing the changes made for the VM itself, which may contain sensitive information, be readable by other users does not seem like it follows good practice. As it currently stands we could correct to limit access further without impacting, instead this is looking to increase access.

In general I'd consider that if using an API to upload a VM disk image, you should use the same API to retrieve it instead of trying to go in and access the file directly. This stepping in behind libvirt to access the files on disk directly seems like a recipe for further issues. Hence my hesitation in making a change like this where it's based on the assumption that by-passing libvirt makes sense.

I think I'd need to see more about virt-v2v to be confident that it's ultimately the right way to go. Since it's targeted at importing from other hypervisors I'm left with questions as to whether it's really used all that much for the libvirt -> libvirt use case. The fact that it expects to have direct access to the disk images is definitely something I'd want to talk to libvirt maintainer about to understand the issues in by-passing the API.

lib/vagrant-libvirt/action/create_domain.rb
hswong3i: Split into #1275

hswong3i:
With more studying on https://libguestfs.org/virt-v2v-copy-to-local.1.html, how it works:
* Copying remote VM as local libvirt instance managed VM
* Running `virt-v2v` with local libvirt instance
* (BOTTLENECK) `virt-v2v` reading the backend storage pool disks with current user, therefore required for at least group-based read access

Long story short, `virt-v2v-copy-to-local` couldn't solve the required read permission issue (because VM is imported into local libvirt instance, and we are NOT going to use `virt-v2v -i disk` since this will miss all addition box base disk information).

Again, IMHO if a user define its `/etc/libvirt/qemu.conf` as below, our storage files should gracefully respect it and created with corresponding *correct* permission and ownership. A hard-coded result (with our own security consideration) which overriding user's custom configuration shouldn't consider as a bug not a feature :-S
```
$ cat /etc/libvirt/qemu.conf | egrep -e '^(user|group) ='
user = "root"
group = "libvirt"
```

electrofelix:
Have you looked at keeping the existing disk retrieval as is, update the disk paths in the retrieved domain XML to use the local paths the disks have been copied to and then execute virt-v2v using the in-place options to normalize?

That will keep everything working as is for remote and local, and should also make supporting export to the different box formats possible when it's possible to use the alternative format. Being able to do that will make it much easier to merge in as experimental support with less concerns.

hswong3i:
> Have you looked at keeping the existing disk retrieval as is, update the disk paths in the retrieved domain XML to use the local paths the disks have been copied to and then execute virt-v2v using the in-place options to normalize?

For `virt-v2v` we could handle with:
* (OK) Single disk + `virt-v2v -i disk` + existing disk retrieval logic
* (OK) Multiple disks + `virt-v2v -i libvirt` + permission fix in this PR
* (FAILED) Multiple disks + `virt-v2v -i disk` (here only keep the root disk in generated box.xml)

Umum... Your suggestion going for (with original VM information degraded during VM -> disk -> VM conversion):
* Retrieval ALL disks with existing logic
* Only generate box.xml for root disk with `virt-v2v -i disk`
* Manually alter box.xml for injecting additional disks...

Well, it really works... But for remote libvirt why not (with original VM information preserved):
* Document `vagrant package` logic only working with local libvirt instance
* For packaging remote box, please manually import it into local libvirt instance with `virt-v2v-copy-to-local`
* Now export single/all disks once together for local libvirt instance, with simple `virt-v2v -i libvirt` + this PR fix?

For security:
* For user don't care about security, running everything with root or sudo
* For user do care about security, running everything with group libvirt, this PR fix permission required for this mode

electrofelix:
> Well, it really works... But for remote libvirt why not (with original VM information preserved):
> 
>     * Document `vagrant package` logic only working with local libvirt instance

Because it currently is possible this means we're removing support.

>     * For packaging remote box, please manually import it into local libvirt instance with `virt-v2v-copy-to-local`

Because vagrant/vagrant-libvirt will have no concept of the VM locally. Importing it to the local libvirt isn't enough for vagrant to be away and be able to package it, and if you could run it locally to package, would had done so already. Which means it's not just more steps, it becomes completely broken and requires separate tooling rendering the `vagrant package` command broken.

hswong3i:
> > Well, it really works... But for remote libvirt why not (with original VM information preserved):
> > ```
> > * Document `vagrant package` logic only working with local libvirt instance
> > ```
> 
> Because it currently is possible this means we're removing support.

Please kindly correct me, I couldn't find any code about `vagrant package` from https://github.com/vagrant-libvirt/vagrant-libvirt/commit/4830182ef8142f295d2e04215341177a01bbafb3, the multiple base disks packaging support is currently missing or broken.

> 
> > ```
> > * For packaging remote box, please manually import it into local libvirt instance with `virt-v2v-copy-to-local`
> > ```
> 
> Because vagrant/vagrant-libvirt will have no concept of the VM locally. Importing it to the local libvirt isn't enough for vagrant to be away and be able to package it, and if you could run it locally to package, would had done so already. Which means it's not just more steps, it becomes completely broken and requires separate tooling rendering the `vagrant package` command broken.

I would comment it as a trade off:
* Multiple base disks box usually more sensitive on VM configuration when comparing with legacy single base disk box
* Packaging those VM configuration with `virt-v2v` box.xml should be more suitable then keep bumping up our custom metadata.json structure with similar functionality
* According to this enhancement we trade off the remote libvirt `vagrant package` support (well, how many user would actually require it for daily operation? Vagrant itself is primary design for local development environment management...)
* User with this rare requirement could still packaging their remote libvirt instance, just need to fire a `virt-v2v-copy-to-local` manually before `vagrant package`

hswong3i:
@electrofelix after some research with `virt-v2v` + `packer` + https://app.vagrantup.com/alvistack multiple OS packaging, I could comment that `virt-v2v` itself is not stable enough across different OS, e.g. working well for Debian 10 and Ubuntu 18.04/20.04, but not Ubuntu 20.10/21.04 (asking for `/boot/efi` in partition) or Fedora 33/34 (missing `/var/lib/rpm/Name`) or CentOS 7/8 (couldn't detect bootloader)... Will detail the proposed changes at #1268.

Therefore this PR no longer primary prepare for #1268, we could just simply review and comment it with its original target, which synchronize snapshot disk storage file permission as the other else file generated by fog-libvirt ;-)

Issue #1256, opened 2021-04-26
Export/Import Box as Original VM with `virsh dumpxml` and `virsh define`?


hswong3i:
A Standard Virtualbox Vagrant box `metadata.json` coming with following simple data structure, then reading all VM definition from bundled `box.ovf`:
```
{
  "provider": "virtualbox"
}
```

Currently our box `metadata.json` coming with following custom format, e.g.
```
{
  "format": "qcow2",
  "provider": "libvirt",
  "virtual_size": 128
}
```

For #602 and #1143 we are discussing add multiple disk support in our Vagrant box format, by [adding additional key/value into `metadata.json`](https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1143/files#diff-515fec02588d6d41f3df9bce232b39a18d0352959c7b51f68fa459f55ef97a52R18-R30), e.g.
```
{
  "provider": "libvirt",
  "format": "qcow2",
  "disks": [
    {
      "virtual_size": 2
    },
    {
      "path": "disk2.qcow2",
      "virtual_size": 10
    }
  ]
}
```

With this direction, if we going to support additional NIC/USB/COM/etc device into vagrant-libvirt, we will keep bumping the custom structure into `metadata.json`, which is not scalable ;-(

What if we [export our VM with `virsh dumpxml`, then import it with `virsh define`](https://serverfault.com/a/707564/628984)? This come with a lot of benefits:

- Keeping `metadata.json` clean and tidy
- Support all original VM definition, e.g. `box.xml` (similar as Virtualbox's `box.ovf` format and functionality)

 
### Steps to reproduce
During `vagrant package`:
1. Export VM definition with `virsh dumpxml`
2. Search the exported `box.xml` for the disk image path and copy to local directory for packaging
3. Package both `metadata.json`, `Vagrantfile`, `box.xml`, and `*.qcow2` as our Vagrant Box archive

During `vagrant up`:
1. Restore original VM definition with `virsh define`
2. Search the `box.xml` for disk image filename, and copy to target directory, and update the VM definition with new disk image absolute path
3. Regenerate a new `<uuid>` for the imported VM
4. Remove some parameters that may conflict, e.g. `<cpu>`, `<interface>`, etc
4. Overwrite other else VM definition as per our existing logic

### Expected behaviour
Extracted box archive should coming with following files, e.g.:
```
box.xml
metadata.json
[...]

PR #1268, opened 2021-05-07
Simplify `metadata.json`, Fetch Image Format and Virtual Size by `qemu-img`

### LINKED ISSUE #1013: File system not fully synced on `vagrant package`
### Steps to reproduce

1. Customize a ppc64el guest vagrant-libvirt box with a provisioning script.
2. `vagrant package`
3. Import the resulting box.
4. `vagrant up` the resulting box.

### Expected behaviour

All customizations specified in the provisioning script, such as systemd modifications, appear in the exported box.

### Actual behaviour

* The upstream box is not shutdown correctly during the `vagrant package` procedure.
* Many modifications provisioned on the upstream box, fail to appear on the downstream box.

For context, my shutdown command configured originally in the root base box with Packer, is `echo 'vagrant' | sudo -S halt -p`. This command appears to work normally when triggered by `vagrant halt`. It's just the `vagrant package` command interfacing with vagrant-libvirt that somehow fails to shutdown the box and thereby sync the file system to virtual disk persistence.

I can confirm that all the customizations are present in the parent box (either by `vagrant ssh` or VNC). However, neither the ensure-host-key shell script nor the ensure-host-key service are present on the child box (`vagrant ssh` broken, hence working on these scripts; use VNC instead).

### System configuration
**OS/Distro version:** Ubuntu 19.04 Disco Dingo amd64

**Libvirt version:** 5.0.0-1ubuntu2.3

**Output of `vagrant version; vagrant plugin list`:**

```console
$ vagrant version; vagrant plugin list
2.2.4
vagrant-libvirt (0.0.45, global)
vagrant-rsync-back (0.0.1, global)
```

bootstrap.sh:

```sh
#!/bin/sh
sudo apt-get update &&
    sudo apt-get install -y clang &&
    sudo apt-get autoclean -y &&
    sudo apt-get clean -y &&
    sudo rm -rf /var/lib/apt/lists/* \
        /var/cache/apt/pkgcache.bin \
        /var/cache/apt/srcpkgcache.bin


sudo cp /tmp/ensure-host-keys /usr/sbin/ensure-host-keys &&
    sudo chmod u+x /usr/sbin/ensure-host-keys &&
    sudo cp /tmp/ensure-host-keys.service /lib/systemd/system &&
    sudo systemctl enable ensure-host-keys
```

ensure-host-keys:

```sh
#!/bin/sh
test \
    -f /etc/ssh/ssh_host_rsa_key \
    -a -f /etc/ssh/ssh_host_ecdsa_key \
    -a -f /etc/ssh/ssh_host_ed25519_key ||
    dpkg-reconfigure openssh-server
```

ensure-host-keys.service:

```
[Unit]
Description=Prepare sshd host keys
Before=ssh.service

[Service]
Type=oneshot
RemainAfterExit=true
StandardOutput=journal
ExecStart=/usr/sbin/ensure-host-keys

[Install]
WantedBy=multi-user.target
```

**A Vagrantfile to reproduce the issue:**

```ruby
    config.vm.box = "mcandre/debian-ppc64el"
    config.vm.box_version = "0.0.1"

    # Prepare for vagrant package
    config.ssh.insert_key = false
    config.vm.synced_folder ".", "/vagrant", type: "rsync", disabled: true

    config.vm.provision "file", source: "ensure-host-keys", destination: "/tmp/ensure-host-keys"
    config.vm.provision "file", source: "ensure-host-keys.service", destination: "/tmp/ensure-host-keys.service"

    config.vm.provision "shell", path: "bootstrap.sh"
end
```

**Are you using upstream vagrant package or your distros package?**

Distro

### Workaround

Forcibly write `sudo sync` at the end of my provisioning scripts.

### LINKED ISSUE #1019: vagrant package fails with qemu >= 2.10

hswong3i:
From #1264 which looking for solution to package box with `virt-v2v`,
ideally base box default hardware information should be provided by
`box.xml`, therefore no need to hardcode custom info inside
`metadata.json`.

This PR include:

  - Fetch all images virtual size and format by `qemu-img` during box
    creation, rather than read in the hardcoded value from
    `metadata.json`
  - Shutdown VM before running packaging
  - Simplify `metadata.json` with only `"provider": "libvirt"`
  - Simplify `Vagrantfile` as reference from Vagrant and Packer

Fixes #1013
Fixes #1019
Fixes #1275
Fixes #1267

### Minimal `box.xml` Generated by `virt-v2v` Always Looks Similar

Lesson learn from https://github.com/alvistack/vagrant-ubuntu/commit/029e5b86d13ca879dfd35315415ab104e5d1b30c: The generated `box.xml` for both Debian 10 and Ubuntu 18.04/20.04 are basically identical, as like as below:
```
<?xml version='1.0' encoding='utf-8'?>
<domain type='kvm'>
  <name>box</name>
  <memory unit='KiB'>2097152</memory>
  <currentMemory unit='KiB'>2097152</currentMemory>
  <vcpu>1</vcpu>
  <features>
    <acpi/>
    <apic/>
  </features>
  <os>
    <type arch='x86_64' machine='q35'>hvm</type>
  </os>
  <on_poweroff>destroy</on_poweroff>
  <on_reboot>restart</on_reboot>
  <on_crash>restart</on_crash>
  <devices>
      <driver name='qemu' type='qcow2'/>
      <source file='box.img'/>
      <target dev='vda' bus='virtio'/>
    </disk>
    <interface type='network'>
      <source network='default'/>
      <model type='virtio'/>
    </interface>
    <video>
      <model type='qxl' ram='65536' heads='1'/>
    </video>
    <graphics type='sdl' autoport='yes' port='-1'/>
    <rng model='virtio'>
      <backend model='random'>/dev/urandom</backend>
    </rng>
    <memballoon model='virtio'/>
    <panic model='isa'>
      <address type='isa' iobase='0x505'/>
    </panic>
    <viosock model='none'/>
    <input type='tablet' bus='usb'/>
    <input type='mouse' bus='ps2'/>
    <console type='pty'/>
  </devices>
</domain>
```

### `virt-v2v` Not Stable Enough Across Different OS

I could only handle `virt-v2v` via Debian 10 and Ubuntu 18.04/20.04, where the following OS are failed:
  * Ubuntu 20.10/21.04 (asking for /boot/efi in partition)
  * Fedora 33/34 (missing /var/lib/rpm/Name)
  * CentOS 7/8 (couldn't detect bootloader)

This means impossible to generate a new default `box.xml` for legacy box on-the-fly with `virt-v2v` during VM creation.

Therefore provide a default simple static `box.xml` as above would be a better idea.

### `vagrant package` and `packer` Support Include Additional Files

Besides patching `packer`, it is already support include additional files: https://www.packer.io/docs/post-processors/vagrant#include. Define our additional base image in a custom `box.xml`, then include it with above option simply solve the problem.

Similar case goes for `vagrant package`: https://www.vagrantup.com/docs/cli/package#include-x-y-z. We could first provide a sample default `box.xml`, then override it with user custom defined include files.

### Could Keep Existing Image Download Logic for both Local or Remote Libvirt Instance

Since no more `virt-v2v` is called during box packaging, we could now reuse existing image download logic, and extend its ability for downloading all images once together.

From the other point of view, since we are NOT using `virt-v2v` now, user need to provide their own custom `box.xml` for packaging a new box with multiple base disks.


hswong3i:
@electrofelix this PR show that it is not necessary to bump our existing metadata.json in order to support additional base box disks.

It could be done by just simply search all disk image files from box directory, and check its format and virtual size on-the-fly.

P.S. I am totally agree with your backward compatibility concern, especially distribution package always behind us for **many** years. That's why I also vote down for bumping `metadata.json` structure: once we kickstart with this direction, we could never roll it back to simple stupid structure, moreover generate much more effort for 3rd party project (e.g. packer) integration.

P.P.S. From https://github.com/libguestfs/virt-v2v#readme mention that *Virt-v2v and virt-p2v have been in continuous development since 2007*. Using its exported xml as foundation of base box import should be a more mature solution then our custom implementation.

electrofelix:
@hswong3i making the other tools in the ecosystem (e.g. packer) adopt XML when they don't use libvirt in the first place doesn't seem like the nicest path. Also making this new format the only output format after this PR is merged is also a path to maintenance headaches because it means there is no easy alternative available to others to use and puts pressure on fixing asap, which considering the amount of available time doesn't seem the best plan.

My view of the path to getting there is to support the different formats side by side for periods of time, and in particular to start with making the future format an option when packaging rather than a default.

* format v1: the old single box
* format v2: multi disk using json from metadata.json
* format v3: box.xml

After a couple of releases where format v2 is the default, format v1 can be removed, and we can switch to format v3 as the default, with a backup for a while of format v2.

This will allow the likes of packer to solve the immediate problem of multi disk support and then consider how to deal with moving to v3 longer term.

It may appear to be more work, but adopting the approach of side-by-side is much easier to merge and experiment with, because it means changes can be built on and problems uncovered instead of needing to solve everything in a single step.

The alternative is to take a long time carefully testing the various PRs.

hswong3i:
> The alternative is to take a long time carefully testing the various PRs.

With https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1264/commits/23d0056d3b9efc50a71a4f15ff84630d85807331 show the ability and potential of `virt-v2v`'s `box.xml`.

Legacy format v1 is supported by a one-off on-the-fly `virt-v2v`, by reading in the legacy `box.img` and generate the most suitable minimal default `box.xml` for it.

Since we didn't release the next stable release yet, my suggestion is skip the format v2, and go strict forward to format v3 with `box.xml` + multiple base box image support.

electrofelix:
With this code landed, any vagrant-libvirt machine packaged by vagrant package will not be usable by any previous version of vagrant-libvirt. It will only be usable by the latest version. This doesn't mean that only new boxes that need the newer capabilities will be only usable, it means all boxes packaged this way.

The v1 => v2 => v3 approach I've laid out is not just about compatibility with the current version can use the older boxes, it's to limit the issues that will occur with people not being able to continue to package boxes that will run with the older versions of vagrant-libvirt where there is no need for the newer capabilities provided by the box.xml.

hswong3i:
> With this code landed, any vagrant-libvirt machine packaged by vagrant package will not be usable by any previous version of vagrant-libvirt. It will only be usable by the latest version. This doesn't mean that only new boxes that need the newer capabilities will be only usable, it means all boxes packaged this way.

What if during `vagrant package`, after `virt-v2v` we simply rename the `box-sda` into `box.img` for backward compatibility?

Therefore new box will coming with both `box.xml` and `box.img` + `box-sdb` + `box-sdc` and so on. Legacy vagrant-libvirt will NEVER read those additional box base disk image information (yes, it just assumed with `box.img`) so everything will keep functioning perfectly.

For new box + new vagrant-libvirt, we read those additional box base disk image information (and even more and more comes later?) from `box.xml`, again functioning perfectly ;-)

hswong3i:
From https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1267#issuecomment-836862165

> Sorry I wasn't referring to multi disk support. Currently packaging a vagrant VM brought up on a remote libvirt works, the proposed solution removes this, so it is removing support for what is currently working. It's not just a temporary removal, I can't see how virt-v2v would ever be capable of working in a suitable manner because it always requires a local libvirt instance that can house the VM being packaged, which is not currently required with the remote libvirt support.
> 
> To restore it in the future would require restoring the existing code and applying the disk download to iterate over multiple disks. Since that code already exists, I don't see the problem in extending it to all of the box disks.

For preserving existing VM configuration together with multiple disks support, we could choose:
* `virsh dumpxml` which target for full and identical clone across host machine, which we already found that toooo much unnecessary information included, and not suitable for Vagrant use case
* `virt-v2v -i libvirt` which target for generating a minimal template for fire up another similar VM, which very close to our Vagrant use case, and also similar as like as VirtualBox/VMware .ova/.ovf export

I couldn't see how our existing manual single disk download logic + custom defined matadata.json structure could handle above requirement precisely in the future...

> You have to create the machine on the local host in order with vagrant to be aware of it. If it's not, it cannot run the package code.

Ok let's brain storming in this way:
* We download ALL disks from remote libvirt instance *manually* with improved existing logic => box-sda, box-sdb, etc
* We *manually* use `virsh dumpxml` or similar ruby call to export the FULL VM configuration => box.ml
* Now we need to trim that box.xml *manually* and carefully, as like as your concern [here](https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1256#issuecomment-826732501)
* Or even more, selectively extract those configuration, and *manually* convert it into our own custom metadata.json schema, and keep bumping it with v3/v4/v5/v6 endlessly for on going feature enhancement...
* All these *manually* additional effort are subject to *preserve remote libvirt `vagrant package` support*... I couldn't understand how high price-performance ratio coming with this solution... ?_?

Here we are just trimming that box.xml according the our limited internal knowledge management, where `virt-v2v` is actively maintenance by RedHat since 2007 for corresponding trimming duty...

P.S. personally I am using packer for https://app.vagrantup.com/alvistack daily, but not our existing `vagrant package` logic... Because it is broken and now usable from some point of view, as https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1268/files#diff-ddcc0e24706ecd3c03ccc31606a887868a71bfd3d3611f82edab05a1976094c5L164 mentioned...

electrofelix:
> Ok let's brain storming in this way:
> 
>     * We download ALL disks from remote libvirt instance _manually_ with improved existing logic => box-sda, box-sdb, etc
> 
>     * We _manually_ use `virsh dumpxml` or similar ruby call to export the FULL VM configuration => box.ml
> 
>     * Now we need to trim that box.xml _manually_ and carefully, as like as your concern [here](https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1256#issuecomment-826732501)

Not quite.

Take the xml domain definition from libvirt using the standard connection, modify the disks paths to use paths to the area that the disks have been downloaded and then write it out. Next use virt-v2v with the libvirtxml option to normalize the XML.

In such a case you can safely enable in-place options to modify the disk images as needed, it doesn't need to reach behind the libvirt API since the disks have already been retrieved, which means it will work for both local and remote, and at the same time hands off the normalization to virt-v2v.

I think it still requires talking to the libvirt developers to understand the implications because it's not necessarily the same people, and the tool looks like it may have been primarily developed and tested on RHEL and so may not necessarily be kept current with the latest libvirt releases. Or maybe it is. Either way, need to know before starting to depend on it.


Fundamentally any change to the format used for packaging needs to be done in an optional manner, removing the existing behaviour to replace it with completely new behaviour isn't going to get accepted. It must be possible to switch the packaging format used through an environment variable until such time as enough of the consumers have upgraded to a version of vagrant-libvirt that will have support for the new format to allow picking up boxes published with the new format.

hswong3i:
Lesson learn from https://github.com/alvistack/vagrant-ubuntu/commit/029e5b86d13ca879dfd35315415ab104e5d1b30c:

### Minimal `box.xml` Generated by `virt-v2v` Always Looks Similar

The generated `box.xml` for both Debian 10 and Ubuntu 18.04/20.04 are basically identical, as like as below:
```
<?xml version='1.0' encoding='utf-8'?>
<domain type='kvm'>
  <name>box</name>
  <memory unit='KiB'>2097152</memory>
  <currentMemory unit='KiB'>2097152</currentMemory>
  <vcpu>1</vcpu>
  <features>
    <acpi/>
    <apic/>
  </features>
  <os>
    <type arch='x86_64' machine='q35'>hvm</type>
  </os>
  <on_poweroff>destroy</on_poweroff>
  <on_reboot>restart</on_reboot>
  <on_crash>restart</on_crash>
  <devices>
      <driver name='qemu' type='qcow2'/>
      <source file='box.img'/>
      <target dev='vda' bus='virtio'/>
    </disk>
    <interface type='network'>
      <source network='default'/>
      <model type='virtio'/>
    </interface>
    <video>
      <model type='qxl' ram='65536' heads='1'/>
    </video>
    <graphics type='sdl' autoport='yes' port='-1'/>
    <rng model='virtio'>
      <backend model='random'>/dev/urandom</backend>
    </rng>
    <memballoon model='virtio'/>
    <panic model='isa'>
      <address type='isa' iobase='0x505'/>
    </panic>
    <viosock model='none'/>
    <input type='tablet' bus='usb'/>
    <input type='mouse' bus='ps2'/>
    <console type='pty'/>
  </devices>
</domain>
```

### `virt-v2v` Not Stable Enough Across Different OS

I could only handle `virt-v2v` via Debian 10 and Ubuntu 18.04/20.04/20.10/21.04, where the following OS are failed:
  * Ubuntu 20.10/21.04 (asking for /boot/efi in partition)
  * Fedora 33/34 (missing /var/lib/rpm/Name)
  * CentOS 7/8 (couldn't detect bootloader)

This means impossible to generate a new default `box.xml` for legacy box on-the-fly during VM creation.

### `vagrant package` and `packer` Support Include Additional Files

Besides patching `packer`, it is already support include additional files: https://www.packer.io/docs/post-processors/vagrant#include. If we hope to provide a custom `box.xml` which defining our additional box base disks, user could simply prepare that and include it manually.

Similar case goes for `vagrant package`: we could first provide a sample default `box.xml`, then override it with user custom defined include files.

### Could Keep Existing Image Download Logic for both Local or Remote Libvirt Instance

Since no more `virt-v2v` is called during box packaging, we could now reuse existing image download logic, and extend its ability for downloading all images once together.

electrofelix:
That unfortunately is the risk with adopting an entirely new approach, and is also one of the reasons to try and not have it be the default packaging format until there is some time spent with it.

Going to back to some of my previous comments, the path to getting to using the libvirt VM xml as part of the box format involves the following steps:

1. Talk to the libvirt developers to understand the portability of the XML
2. Depending on the plan around disks should talk to the packer devs about the impact
   * If the disks need to be in the XML, then packer will need to support including it, if it's just a case of bundling a box XML that specifies any additional devices that is used as the starting point and vagrant-libvirt handles adding the disks. Then they don't need to modify it and instead can just take what the user supplies, then possibly can be skipped.
3. Expose the ability to provide a callback to the user that can be defined in the Vagrantfile to mutate the XML before creating and or starting the domain, this will allow papering over most issues encountered by any box using the XML format. This would be useable right now for any existing box and avoids supporting edge case capabilities in the future.
4. Add the capabilities for fully supporting use of the XML as a starting template via an optional packaging format, not used by default, only if someone enables it for packaging.

The more earlier items done before item 4, the less risk there is including and the easier it is to merge earlier to build up additional testing subsequently compared to trying to jump in and get the format right from the start. Considering that the first approach was to use the XML directly from `virsh dumpxml <name>`, which is not portable, second was to use virt-v2v which itself is not portable, there is no reason there won't be further issues encountered. Using an env var to enable packaging to the new format would go a long way to being able to release in an experimental state.

electrofelix:
I've had a good look at the packaging code that is currently in place over this weekend, and I think before adding any additional formats I'm going to rewrite it to use the general packaging middleware from vagrant itself under https://github.com/hashicorp/vagrant/tree/main/lib/vagrant/action/general and remove some of the special handling around the extra files, some error handling, and assembling the box.

Probably will start with implementing some integration level tests to cover the following:
* Repackage infernix/core with defaults
* Repackage with additional files
* Repackage with custom Vagrantfile

Then with the switch to use Vagrant::Action::General::Package as the parent class I'll add unit tests to cover the expected behaviour, and rely on the above integration tests to ensure that the final contents are correct to help ensure I don't miss anything when constructing the unit tests.

This will break this PR further, but it should also make it much smaller and simpler to add the box.xml as a format subsequently, not to mention easier to to follow the impact by having tests cover the packaging behaviour to ensure that it behaves as expected.

PR #1275, opened 2021-05-10
Bugfix Additional Disks Storage File Ownership

hswong3i:
For additional disk image created by
`lib/vagrant-libvirt/action/create_domain.rb`, it is now incorrectly
hard-coded with `group: storage_uid(env)`, so even QEMU process is
running with `root:libvirt` but file is still owned by `root:root`


electrofelix:
@hswong3i  can you wrap the commit message to 72 characters to make it a little easier when using git log locally. I can't modify the PR because you haven't given permissions to maintainers to make changes, and if I squash + merge github will remove you as the author (which is a known issue I believe).

I've made the change to the PR description, but that won't apply to a rebase + merge

hswong3i:
Commit message updated ;-)

PR #1276, opened 2021-05-10
Corrections to two test testing tool

electrofelix:
Ensure the two disk test tool is used to ensure any tweaks to the format
made to the script are exercised.


electrofelix:
@yamatoRT looks like it was a path expansion issue that caused the problem with your script for #1143, this should fix it

Issue #1002, opened 2019-05-07
Can't boot on PXE with public network!


0xelion:
Hi!
I've used one of the network examples to make my VM boot in PXE while having a box:

```
     config.vm.define :pxeclient do |pxeclient|
      pxeclient.vm.network :private_network,ip: '10.0.0.5',
            libvirt__network_name: "foreman_managed",
            libvirt__dhcp_enabled: false,
            libvirt__host_ip: '10.0.0.1'

       pxeclient.vm.provider :libvirt do |domain|
          domain.memory = 1000
          boot_network = {'network' => 'foreman_managed'}
          domain.storage :file, :size => '100G', :type => 'qcow2'
          domain.boot boot_network
          domain.boot 'hd'
        end
      end
```

However, i wanted to connect to a local PXE server, so i tried with a bridged public network configuration.

With just `libvirt.boot "network"`, only the first network which is automatically created try to boot and it doesn't work.

And when i tried to give a network name to the public network and set the boot values as:
```
      boot_network = {'network' => 'foreman_managed'}
      libvirt.boot boot_network
      libvirt.boot 'hd'`
```
It only boot on the box, no network boot

### Steps to reproduce
1. sudo vagrant up (bridge on hotedev requires sudo)
2. the vm is deployed but it doesn't boot properly

### Expected behaviour
I should be able to boot on the public network with PXE.

### Actual behaviour
Either only the first auto network wants to boot or only the box.

### System configuration
**OS/Distro version:**:
Linux Fedora 5.0.11-300.fc30.x86_64 GNU/Linux
**Libvirt version:**
Libvirt 5.1.0

**Output of `vagrant version; vagrant plugin list`:**
Vagrant Installed Version: 2.2.3
vagrant-libvirt (0.0.45, system)

**A Vagrantfile to reproduce the issue:**
```
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|
  config.vm.define :libvirt_capture do |config|
    config.vm.box = "roboxes/fedora30"
    config.vm.network "public_network",
    :dev => "enp1s0",
    :mode => "bridge",
    :network_name => "foreman_managed"

    config.vm.provider :libvirt do |libvirt|
      libvirt.cpus = 4
      libvirt.memory = 2048
      libvirt.driver = "kvm"
      boot_network = {'network' => 'foreman_managed'}
      libvirt.boot boot_network
      libvirt.boot 'hd'
[...]

PR #1294, opened 2021-05-26
add support for boot 'network' from a vagrant public_network

rgl:
when we use a `public_network` in vagrant, this is how libvirt domain xml looks like (notice that the interface type is 'direct'):

```xml
    <interface type="direct">
      <mac address="08:00:27:00:00:01"/>
      <source dev="br-rpi" mode="bridge"/>
      <model type="virtio"/>
      <boot order="2"/> <!-- this boot xml node will be inserted with this commit -->
      <address type="pci" domain="0x0000" bus="0x00" slot="0x05" function="0x0"/>
    </interface>
```

closes https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1002

### DIFF
--- a/lib/vagrant-libvirt/action/set_boot_order.rb
+++ b/lib/vagrant-libvirt/action/set_boot_order.rb
@@ -86,9 +86,13 @@ module VagrantPlugins
 
         def search_network(nets, xml)
           str = '/domain/devices/interface'
-          str += "[(@type='network' or @type='udp' or @type='bridge')"
+          str += "[(@type='network' or @type='udp' or @type='bridge' or @type='direct')"
           unless nets.empty?
-            str += " and source[@network='#{nets.first['network']}']"
+            net = nets.first
+            network = net['network']
+            dev = net['dev']
+            str += " and source[@network='#{network}']" if network
+            str += " and source[@dev='#{dev}']" if dev
           end
           str += ']'
           @logger.debug(str)
--- a/README.md
+++ b/README.md
@@ -1389,6 +1389,24 @@ Name of network "foreman_managed" is key for define boot order
       end
 ```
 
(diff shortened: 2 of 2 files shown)


lib/vagrant-libvirt/action/set_boot_order.rb
```diff
@@ -86,9 +86,13 @@ def final_boot_order(boot_order, devices)
...
           str = '/domain/devices/interface'
-          str += "[(@type='network' or @type='udp' or @type='bridge')"
+          str += "[(@type='network' or @type='udp' or @type='bridge' or @type='direct')"
           unless nets.empty?
-            str += " and source[@network='#{nets.first['network']}']"
+            net = nets.first
+            network = net['network']
+            dev = net['dev']
+            str += " and source[@network='#{network}']" if network
+            str += " and source[@dev='#{dev}']" if dev
```
electrofelix, line 95: Think add a note to the readme to indicate that the boot_network can contain a `dev` reference as well.

For my better understanding, are these exclusive options? as in can the source contain both? or is it going to be only one or the other? It looks like they cannot both be set based on https://libvirt.org/formatdomain.html#network-interfaces

rgl: I've added an example for the `public_network` that uses the `dev` filter.

I also think they cannot be used together, but I'm not really sure (I could not find a formal libvirt xml schema that describes all the available options to confirm; the current docs are somewhat dense to read and understand).

Do you want me to raise an error when we try to use both `network` and `dev` in the filter?

electrofelix: I don't think so, not yet anyway, when we get to adding tests for more of the networking stack then it's possibly something that can be looked at when there is confidence on what XML blocks are actually acceptable.

Issue #994, opened 2019-04-05
box export broken in multiple ways


mcandre:
Hi, I am trying to add some configuration to a libvirt Vagrant box and then export that for reuse. I can do this with VMware and VirtualBox guests, but when I do this with libvirt, I get all sorts of errors!

Source:

https://github.com/mcandre/vagrant-debian-cpp/tree/ppc64el

One error strangely indicates missing read permissions on .IMG files. Even though Packer and `vagrant up` can read image files just fine.

When I grant read permissions, the next error has to do with a missing filename. The Ruby log seems to use a blank filename here?

And after enough runs, I finally get an error about some `_tmp_package` file already exists. Ideally that file would simply be overwritten. For now I am manually removing that file and trying and trying again.

I'm using base box mcandre/debian-ppc64el, on qemu 2.12 with machine "pseries", with vagrant-libvirt 0.0.45, with libvirt 4.6.0, on Vagrant 2.2.4, on Ubuntu 18.10 Cosmic Cuttlefish, on a System76 Galaga Pro.

mcandre:
Example trace:

```console
$ vagrant package --output debian-ppc64el.box
...
/home/andrew/.vagrant.d/gems/2.4.4/gems/vagrant-libvirt-0.0.45/lib/vagrant-libvirt/action/package_domain.rb:41:in ``': No such file or directory - virt-sysprep (Errno::ENOENT)
	from /home/andrew/.vagrant.d/gems/2.4.4/gems/vagrant-libvirt-0.0.45/lib/vagrant-libvirt/action/package_domain.rb:41:in `call'
	from /opt/vagrant/embedded/gems/2.2.4/gems/vagrant-2.2.4/lib/vagrant/action/warden.rb:50:in `call'
	from /opt/vagrant/embedded/gems/2.2.4/gems/vagrant-2.2.4/lib/vagrant/action/builtin/before_trigger.rb:23:in `call'
	from /opt/vagrant/embedded/gems/2.2.4/gems/vagrant-2.2.4/lib/vagrant/action/warden.rb:50:in `call'
	from /opt/vagrant/embedded/gems/2.2.4/gems/vagrant-2.2.4/lib/vagrant/action/builtin/after_trigger.rb:26:in `call'
	from /opt/vagrant/embedded/gems/2.2.4/gems/vagrant-2.2.4/lib/vagrant/action/warden.rb:50:in `call'
...
```

mcandre:
[...]

PR #1307, opened 2021-06-07
Basic packaging tests and restructure action

electrofelix:
Restructure action to remove custom handling around packaging of the
box and instead use more of the built-in provided actions instead.

Includes some packaging tests to cover both simple where the public
key is retained (can't modify the tinycore VM without more complex
provisioning steps), and a more complex one that utilizes a script
and supports triggering regenerating the hosts on subsequent boots.

The use of the upstream packaging helpers means that when the
default insecure ssh key has been replaced, the packaging process
will automatically include the generated key.

Fixes: #759
Fixes: #765
Fixes: #1013
Fixes: #994

### DIFF
--- a/lib/vagrant-libvirt/action.rb
+++ b/lib/vagrant-libvirt/action.rb
@@ -4,8 +4,9 @@ require 'log4r'
 module VagrantPlugins
   module ProviderLibvirt
     module Action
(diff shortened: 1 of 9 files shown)


electrofelix:
This also fixed #1019

Issue #1308, opened 2021-06-09
GPT PMBR size mismatch with second disk


yamatoRT:
Hi all,

when I use a second disk in box, the VM wrong detect the disk. When I use `fdisk`, I have this message :
```
GPT PMBR size mismatch (20971528 != 20971519) will be corrected by w(rite).
The backup GPT table is corrupt, but the primary appears OK, so that will be used.

Disk /dev/sdb: 10 GiB, 10737418240 bytes, 20971520 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: 3FF63104-B751-AB4A-A567-96BBCD052F77

Device     Start      End  Sectors Size Type
/dev/sdb1   2048 20971495 20969448  10G Linux filesystem
```

### Steps to reproduce
1. Create alias  :
```bash
alias vagrant='
  mkdir -p ~/.vagrant.d/{boxes,data,tmp}; \
  docker run -it --rm \
    -e LIBVIRT_DEFAULT_URI \
    -v /var/run/libvirt/:/var/run/libvirt/ \
    -v ~/.vagrant.d:/.vagrant.d \
    -v $(pwd):$(pwd) \
    -w $(pwd) \
    --network host \
    vagrantlibvirt/vagrant-libvirt:latest \
    vagrant'
```
2. Set a attached Vagrantfile in current directory
3. vagrant up
4. Into the vm, run the command `sudo fdisk -l`

### Expected behaviour
The second disk is normally mounted

### Actual behaviour
The size of disk doesn't match with the GPT PMBR size 

### System configuration
**OS/Distro version:**:
Linux archbox 4.3.3-3-ARCH

**Libvirt version:**
```
vagrant-libvirt (0.5.2, global)
  - Version Constraint: 0.5.2
```
**Output of `vagrant version; vagrant plugin list`:**
```
Installed Version: 2.2.10
Latest Version: 2.2.16
 
To upgrade to the latest version, visit the downloads page and
download and install the latest version of Vagrant from the URL
below:

  https://www.vagrantup.com/downloads.html

If you're curious what changed in the latest release, view the
CHANGELOG below:

  https://github.com/hashicorp/vagrant/blob/v2.2.16/CHANGELOG.md
Starting with UID: 1000, GID: 1000
[...]

PR #1309, opened 2021-06-10
Use qemu-img json output and compute virtual size #1308

yamatoRT:
Hi,

this merge request fix the bug #1308 

I use the json output of `qemu-img info` for read virtual size of disk.
I create new `ByteNumber` for convert the virtual size to gygabyte, byte and megabyte.
Now, the unit of virtual size is byte and not gigabyte.

### DIFF
--- a/lib/vagrant-libvirt/action/create_domain.rb
+++ b/lib/vagrant-libvirt/action/create_domain.rb
@@ -266,7 +266,7 @@ module VagrantPlugins
           end
           env[:ui].info(" -- Storage pool:      #{@storage_pool_name}")
           @domain_volumes.each do |volume|
-            env[:ui].info(" -- Image(#{volume[:device]}):     #{volume[:path]}, #{volume[:virtual_size]}G")
+            env[:ui].info(" -- Image(#{volume[:device]}):     #{volume[:path]}, #{volume[:virtual_size].to_GB}G")
           end
 
           if not @disk_driver_opts.empty?
--- a/lib/vagrant-libvirt/action/handle_box_image.rb
+++ b/lib/vagrant-libvirt/action/handle_box_image.rb
@@ -1,5 +1,8 @@
 require 'log4r'
 require 'open3'
+require 'json'
+
+require 'vagrant-libvirt/util/byte_number'
 
 module VagrantPlugins
   module ProviderLibvirt
@@ -65,7 +68,7 @@ module VagrantPlugins
               {
                 :path => image_path,
                 :name => volume_name,
-                :virtual_size => virtual_size.to_i,
+                :virtual_size => virtual_size,
                 :format => HandleBoxImage.verify_box_format(format)
               }
             }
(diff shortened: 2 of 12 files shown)


tests/runtests.bats
```diff
@@ -84,7 +84,7 @@ cleanup() {
 @test "bring up with two disks" {
   export VAGRANT_CWD=tests/two_disks
   cleanup
-  tools/create_box_with_two_disks.sh ${VAGRANT_HOME} ${VAGRANT_CMD}
+  tools/create_box_with_two_disks.sh "${VAGRANT_HOME}" "${VAGRANT_CMD}"
```
electrofelix, line 87: Did something flag these? Just wondering as it looks like something that I get reminded by shellcheck when I remember to run it, seems like we could benefit from maybe enabling it to run on the repo and fix up any issues as a separate PR?

electrofelix (commented):
I think there is a typo, suggestion to delay conversion to 'G' in one case, takes a bit of checking otherwise to realise it's safe, just less likely to be a future mistake if delay conversion until displaying.

lib/vagrant-libvirt/action/handle_box_image.rb
```diff
@@ -81,16 +84,17 @@ def call(env)
           # Override box_virtual_size
           box_virtual_size = env[:box_volumes][0][:virtual_size]
           if config.machine_virtual_size
-            if config.machine_virtual_size < box_virtual_size
+            config_machine_virtual_size = ByteNumber.from_BG(machine_virtual_size)
```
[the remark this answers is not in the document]
yamatoRT: Yes, it is a mistake. I forget a commit in my MR. It's done, and I append new unit test for this block.

tests/runtests.bats
yamatoRT: I remove this. When I run `bats ./test/runtests.sh`, I need to use the command `bundle exec vagrant` instead of `vagrant`.
Then this test fail because `bash` run the command `tools/create_box_with_two_disks.sh "${VAGRANT_HOME}" bundle exec vagrant` instead of `tools/create_box_with_two_disks.sh "${VAGRANT_HOME}" "bundle exec vagrant"`

lib/vagrant-libvirt/action/create_domain.rb
```diff
@@ -155,7 +155,7 @@ def call(env)
                 :cache => @domain_volume_cache,
                 :bus => @disk_bus,
                 :path => domain_volume.path,
-                :virtual_size => env[:box_volumes][index][:virtual_size]
+                :virtual_size => env[:box_volumes][index][:virtual_size].to_GB
```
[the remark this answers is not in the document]
yamatoRT: done, i call `to_GB` when we display information.

tests/runtests.bats
electrofelix: Ah, that would be handy to have, usually I'm running after doing `bundle --binstubs ...`, no harm if putting it up with a separate PR and we can try and put together something that runs shellcheck because I suspect it would have caught the issue you bumped into

electrofelix (approved):
lgtm, thanks

PR #1310, opened 2021-06-11
Use GracefulHalt builtin action

chrisroberts:
This PR updates the halt action behavior to utilize the
GracefulHalt builtin action and fallback to the HaltDomain
action when the GracefulHalt action is unsuccessful. This
allows for removal of the GracefulHalt-like implementation.

Related to: hashicorp/vagrant#12368


chrisroberts:
Result is behavior that looks like this:

```
workspace/hashicorp/vagrant-scratch-5 via 💎 v2.7.2 via ⍱ v2.2.17.dev on 🅰 (us-east-1)
❯ be vagrant up
Bringing machine 'default' up with 'libvirt' provider...
==> default: Checking if box 'generic/ubuntu1804' version '3.2.24' is up to date...
==> default: Creating image (snapshot of base box volume).
==> default: Creating domain with the following settings...
==> default:  -- Name:              vagrant-scratch-5_default
==> default:  -- Description:       Source: /home/spox/workspace/hashicorp/vagrant-scratch-5/Vagrantfile
==> default:  -- Domain type:       kvm
==> default:  -- Cpus:              2
==> default:  -- Feature:           acpi
==> default:  -- Feature:           apic
==> default:  -- Feature:           pae
==> default:  -- Clock offset:      utc
==> default:  -- Memory:            2048M
==> default:  -- Management MAC:
==> default:  -- Loader:
==> default:  -- Nvram:
==> default:  -- Base box:          generic/ubuntu1804
==> default:  -- Storage pool:      default
==> default:  -- Image():     /var/lib/libvirt/images/vagrant-scratch-5_default.img, 128G
==> default:  -- Disk driver opts:  cache='default'
==> default:  -- Kernel:
==> default:  -- Initrd:
==> default:  -- Graphics Type:     vnc
==> default:  -- Graphics Port:     -1
==> default:  -- Graphics IP:       127.0.0.1
==> default:  -- Graphics Password: Not defined
==> default:  -- Video Type:        cirrus
==> default:  -- Video VRAM:        256
==> default:  -- Sound Type:
==> default:  -- Keymap:            en-us
==> default:  -- TPM Backend:       passthrough
==> default:  -- TPM Path:
==> default:  -- INPUT:             type=mouse, bus=ps2
==> default: Creating shared folders metadata...
==> default: Starting domain.
==> default: Waiting for domain to get an IP address...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 192.168.121.122:22
    default: SSH username: vagrant
    default: SSH auth method: private key
    default:
    default: Vagrant insecure key detected. Vagrant will automatically replace
    default: this with a newly generated keypair for better security.
    default:
    default: Inserting generated public key within guest...
    default: Removing insecure key from the guest if it's present...
    default: Key inserted! Disconnecting and reconnecting using new SSH key...
==> default: Machine booted and ready!

workspace/hashicorp/vagrant-scratch-5 via 💎 v2.7.2 via ⍱ v2.2.17.dev on 🅰 (us-east-1) took 18s
❯ be vagrant ssh -- sudo service sshd stop

workspace/hashicorp/vagrant-scratch-5 via 💎 v2.7.2 via ⍱ v2.2.17.dev on 🅰 (us-east-1) took 2s
❯ be vagrant halt
==> default: Attempting graceful shutdown of VM...
    default: Guest communication could not be established! This is usually because
    default: SSH is not running, the authentication information was changed,
    default: or some other networking issue. Vagrant will force halt, if
    default: capable.
==> default: Halting domain...

workspace/hashicorp/vagrant-scratch-5 via 💎 v2.7.2 via ⍱ v2.2.17.dev on 🅰 (us-east-1)
❯ be vagrant status
Current machine states:

default                   shutoff (libvirt)

The Libvirt domain is not running. Run `vagrant up` to start it.

workspace/hashicorp/vagrant-scratch-5 via 💎 v2.7.2 via ⍱ v2.2.17.dev on 🅰 (us-east-1)
❯
```

electrofelix:
@chrisroberts is it alright if I keep this open, I'm probably a couple of weeks from working on it to complete the integration, but I'd like to use this as the starting point?

lib/vagrant-libvirt/action/shutdown_domain.rb
```diff
@@ -0,0 +1,49 @@
...
+        end
+
+        def call(env)
+          timeout = env[:machine].config.vm.graceful_halt_timeout
+
+          start_time = Time.now
+
+          # call nested action first under the assumption it should try to
+          # handle shutdown via client capabilities
+          @app.call(env)
```
electrofelix, line 22: @chrisroberts I've done some reworking and testing, and part of me is wondering whether instead of calling the next action in the chain, it would be better if this action either subclassed or called the GracefulHalt action directly at this point instead. As I don't think it's necessarily valid to treat them as separate actions where it's feasible to add this action elsewhere as it always needs to understand what it is wrapping. Thoughts?

chrisroberts: Yeah, the way it's structured now, the `ShutdownDomain` action is making an assumption/requirement on what is being done after it and if that's the case then just calling it directly would be better simply for ensuring the expected behavior. You can remove that assumed dependency, though, using the `EnvSet` action to just log the start time for you. Something like:

```ruby
b3.use Call, Message, "Attempting nice shutdowns..." do |_, b4|
  b4.use EnvSet, :libvirt_shutdown_start, Time.now
  b4.use GracefulHalt, :shutoff, :running
  b4.use ShutdownDomain, :shutoff, :running
end
```

Then, in this `ShutdownDomain` action, your `start_time` value can be:

```ruby
start_time = env.fetch(:libvirt_shutdown_start, Time.now)
```

so if the start time was set, you can use that, otherwise you just assume you have the full timeout to proceed.

PR #1330, opened 2021-08-03
Unrestrict rspec-* version to >= 3.5 to sync with vagrant

dcermak:
Vagrant recently bumped its minimum rspec version to 3.10. This brings the version in sync with vagrant so that we do not require two different versions of the same gem.

### DIFF
--- a/vagrant-libvirt.gemspec
+++ b/vagrant-libvirt.gemspec
@@ -17,9 +17,9 @@ Gem::Specification.new do |s|
   s.require_paths = ['lib']
   s.version       = VagrantPlugins::ProviderLibvirt.get_version
 
-  s.add_development_dependency "rspec-core", "~> 3.5"
-  s.add_development_dependency "rspec-expectations", "~> 3.5"
-  s.add_development_dependency "rspec-mocks", "~> 3.5"
+  s.add_development_dependency "rspec-core", ">= 3.5"
+  s.add_development_dependency "rspec-expectations", ">= 3.5"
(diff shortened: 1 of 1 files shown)


dcermak:
Hm, this will unfortunately require dropping support for older ruby & vagrant versions. Is that something that we can do, or is that a nogo?

vagrant-libvirt.gemspec
```diff
@@ -17,9 +17,9 @@ Gem::Specification.new do |s|
   s.require_paths = ['lib']
   s.version       = VagrantPlugins::ProviderLibvirt.get_version
 
-  s.add_development_dependency "rspec-core", "~> 3.5"
-  s.add_development_dependency "rspec-expectations", "~> 3.5"
-  s.add_development_dependency "rspec-mocks", "~> 3.5"
+  s.add_development_dependency "rspec-core", "~> 3.10"
+  s.add_development_dependency "rspec-expectations", "~> 3.10"
+  s.add_development_dependency "rspec-mocks", "~> 3.10"
```
electrofelix, line 22: Maybe we could just lift the constraints and rely on what is listed in vagrant? e.g. either remove the version completely or use ">= 3.5", I don't think we're really tied that tightly, more likely it was to reduce risk of surprises.

dcermak: Sounds like a more reasonable approach, let's see what the CI says about that 😉

electrofelix (approved):
Looks good to me

PR #1334, opened 2021-08-22
Floppy support

guss77:
Add support for attaching 1 or 2 floppy drives to the VM using "vfd" image files.

### DIFF
--- a/docs/configuration.markdown
+++ b/docs/configuration.markdown
@@ -674,6 +674,25 @@ Vagrant.configure("2") do |config|
 end
 ```
 
+## Floppies
+
+You can attach up to two floppies to a VM via `libvirt.storage :file,
+:device => :floppy`. Available options are:
+
+* `path` - The path to the vfd image to be used for the floppy drive.
+* `dev` - The device to use (`fda` or `fdb`). This will be
+  automatically determined if unspecified.
+
+The following example creates a floppy drive in the VM:
+
+```ruby
+Vagrant.configure("2") do |config|
+  config.vm.provider :libvirt do |libvirt|
+    libvirt.storage :file, :device => :floppy, :path => '/path/to/floppy.vfs'
+  end
+end
+```
+
 ## Input
 
 You can specify multiple inputs to the VM via `libvirt.input`. Available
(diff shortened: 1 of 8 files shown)


electrofelix (commented):
The cdrom config function is doing the wrong thing unfortunately, as the number of devices should be checked in the validate function instead.

After moving, can you squash down the commits so it can be landed as a single commit.

electrofelix:
@guss77 are you happy to update to do the number of floppy drives check in the validate function and provide a comment in the code where the check is done as to what is known about the limit?

guss77:
Sorry for dropping out, I'll go over everything later this week.

brianjmurrell:
Does this PR being open mean there is no support in vagrant-libvirt to be able to add a floppy drive to a box?

TBH, I don't really even care to have a floppy in my VM but it seems that `mdadm` tries to scan for floppies and hangs in doing so (apologies for the screenshot instead of text but I only had access to a console on this VM) when one isn't present:

image

So the path of least resistance seems to be to just give it a floppy to scan.  It seems the GParted folks discovered this also:

https://sourceforge.net/p/gparted/mailman/message/36899927/

It seems they disabled the floppy in the BIOS.  Is that an option with vagrant-libvirt?

electrofelix:
@brianjmurrell unfortunately yes. Technically it should be possible to call attach-device after the domain is created but before it is started through the vagrant triggers via it's action hook mechanism. That would allow connecting floppy devices without them being in the original domain definition.

Alternatively if you could rebase this change, fix the couple of issues that I've commented on and submit as a new PR (could you keep the original author in place though so they get credited with the original work and add co-authored-by), I'd be happy to merge it

electrofelix:
Going to handle the validation and device allocation separately afterwards for both cdrom and floppies together.

electrofelix:
Just as a note here, libvirt only supports one floppy controller and it only supports 2 devices being attached, otherwise the following error will be emitted:
```
Error while creating domain: Error saving the server: Call to virDomainDefineXML failed: internal error: Only 1 fdc controller is supported
```

Issue #1341, opened 2021-09-09
Retrieve IP Address of management interface using qemu-guest-agent commands


abbbi:
hi,

i think this topic has been discussed several times (#1124,#1204 for example).

Currently the management network depends entirely on the internal libvirt dnsmasq dhcp service.
Its currently not possible to use vagrant-libvirt with a external dhcp service, because the only way to find out about the virtual machines IP address are the:

 * the local libvirt dnsmasq lease database
 * the arp table on the system.

However, there is another possibility. Imagine the following setup:

* libvirt setup (with multiple hosts)
* all hosts have routed network configuration or use bridged  host networking for the vagrant virtual machines
* an external DHCP server is answering requests for those bridged devices
* an guest image has the qemu-guest agent installed and the virtual machine has an agent channel setup via:

`libvirt.channel :type => 'unix', :target_name => 'org.qemu.guest_agent.0', :target_type => 'virtio'`

Now as soon as the qemu guest agent is reachable within the spinned up virtual machine, its
possible to receive the virtual machines network configuration via qemu-guest commands,
for example:

```
virsh qemu-agent-command 524 '{"execute":"guest-network-get-interfaces"}' --pretty
    {
      "name": "eth0",
      "ip-addresses": [
        {
          "ip-address-type": "ipv4",
          "ip-address": "192.168.161.21",
          "prefix": 24
        }
      ],
[..]
"hardware-address": "52:54:00:f0:85:1e"
```

this way the vagrant-libvirt plugin could be modified to:

* Allow to attach an management interface to an existing bridge or an network that has not dhcp enabled, and is merely an libvirt bridged network
* virtual machine boots, check if qemu guest agent is reachable
* if its reachable, get network configuration via qemu guest agent and wait until device configured as management_network device (with the matching mac address) gets ip address configured
* continue provisioning using this ip.

I think pretty much all of the common generic/* virtual machine images have the guest agent installed by default, so it is available during boot already (confirmed with debian9/10/11 and peru/windows-2019* images).

That would allow vagrant-libvirt to be used in bigger environments where you dont want to hide the virtual machines behind NAT and want to use an DHCP service for multiple host systems.

(ruby 'libvirt' bindings have qemu_agent_command)

abbbi:
Short example (timeout= needs to be > 0 otherwise the ruby libvirt bindings allways tell the qemu agent is not reachable):

```
require 'libvirt'

conn = Libvirt::open("qemu:///system")
[...]

PR #1343, opened 2021-09-09
Retrieve IP Address of management interface using qemu-guest-agent commands #1341

abbbi:
See issue #1341 for detailed explanation.

Following setup:

 * host system with debian 11 running, with bridge for network interface configured. Network interface is connected to home router wit dhcp enabled.

```
  brctl addbr br0
  brctl addif br0 enp0s31f6 
  ifconfig br0 up
  ifconfig enp0s31f6 up
```

As a test, dhclient -vv br0 assigns ip address via dhcp from home network router.

Now create virtual machine with following definition:

```
 # config.vagrant.plugins = "vagrant-libvirt"     
  config.vm.box = "generic/debian11"     
  config.vm.provider "libvirt" do |libvirt|    
    libvirt.qemu_use_agent = true
    libvirt.channel :type => 'unix', :target_name => 'org.qemu.guest_agent.0', :target_type => 'virtio'
    # attach management network to existing bridge
    libvirt.management_network_device = "br0"
  end
end
```

vagrant up:

```
bundle exec vagrant up --debug
[..]
DEBUG driver: Unable to receive IP via qemu agent: [Call to virDomainQemuAgentCommand failed: Guest agent is not responding: QEMU guest agent is not connected]
 INFO driver: Get IP via qemu agent
DEBUG driver: Unable to receive IP via qemu agent: [Call to virDomainQemuAgentCommand failed: Guest agent is not responding: QEMU guest agent is not connected]
 INFO driver: Get IP via qemu agent
DEBUG driver: Got Response from qemu agent
DEBUG driver: {"return":[{"name":"lo","ip-addresses":[{"ip-address-type":"ipv4","ip-address":"127.0.0.1","prefix":8}],"statistics":{"tx-packets":160,"tx-errs":0,"rx-bytes":13920,"rx-dropped":0,"rx-packets":160,"rx-errs":0,"tx-bytes":13920,"tx-dropped":0},"hardware-address":"00:00:00:00:00:00"},{"name":"eth0","hardware-address":"52:54:00:fd:ab:49"}]}
[..]
DEBUG driver: Got Response from qemu agent
DEBUG driver: {"return":[{"name":"lo","ip-addresses":[{"ip-address-type":"ipv4","ip-address":"127.0.0.1","prefix":8}],"statistics":{"tx-packets":160,"tx-errs":0,"rx-bytes":13920,"rx-dropped":0,"rx-packets":160,"rx-errs":0,"tx-bytes":13920,"tx-dropped":0},"hardware-address":"00:00:00:00:00:00"},{"name":"eth0","ip-addresses":[{"ip-address-type":"ipv4","ip-address":"192.168.178.167","prefix":24}],"statistics":{"tx-packets":4,"tx-errs":0,"rx-bytes":5268,"rx-dropped":15,"rx-packets":46,"rx-errs":0,"tx-bytes":832,"tx-dropped":0},"hardware-address":"52:54:00:fd:ab:49"}]}
DEBUG driver: Found mathing interface: [eth0]
DEBUG driver: Return IP: [192.168.178.167]
[..]
```
spinned up virtual machine was assigned ip 192.168.178.167 on management interface from external router and is fully provisioned and reachable:

```
bundle exec vagrant ssh -c "ip addr"

2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 52:54:00:fd:ab:49 brd ff:ff:ff:ff:ff:ff
    altname enp0s6
    altname ens6
    inet 192.168.178.167/24 brd 192.168.178.255 scope global dynamic eth0
       valid_lft 863870sec preferred_lft 863870sec
Connection to 192.168.178.167 closed.

```

### DIFF
--- a/lib/vagrant-libvirt/action/create_networks.rb
+++ b/lib/vagrant-libvirt/action/create_networks.rb
@@ -162,7 +162,7 @@ module VagrantPlugins
             if @interface_network[:created]
               # Just check for mismatch error here - if name and ip from
               # config match together.
-              if @options[:network_name] != @interface_network[:name]
+              if @options[:network_name] != @interface_network[:name] and @qemu_use_agent = false
                 raise Errors::NetworkNameAndAddressMismatch,
                       ip_address:   @options[:ip],
                       network_name: @options[:network_name]
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -193,6 +193,9 @@ module VagrantPlugins
       # Use QEMU session instead of system
       attr_accessor :qemu_use_session
 
+      # Use QEMU Agent to get ip address
+      attr_accessor :qemu_use_agent
+
       def initialize
         @uri               = UNSET_VALUE
         @driver            = UNSET_VALUE
@@ -332,6 +335,9 @@ module VagrantPlugins
         @qemu_env          = UNSET_VALUE
 
         @qemu_use_session  = UNSET_VALUE
+
+        # Use Qemu agent to get ip address
+        @qemu_use_agent  = UNSET_VALUE
       end
 
       def boot(device)
@@ -936,6 +942,8 @@ module VagrantPlugins
 
         # Additional QEMU commandline environment variables
         @qemu_env = {} if @qemu_env == UNSET_VALUE
+
+        @qemu_use_agent = true if @qemu_use_agent != UNSET_VALUE
       end
 
       def validate(machine)
@@ -948,6 +956,17 @@ module VagrantPlugins
             errors << "the URI and qemu_use_session configuration conflict: uri:'#{@uri}' qemu_use_session:'#{@qemu_use_session}'"
           end
         end
+
+
+        if @qemu_use_agent == true
+          # if qemu agent is used to optain domain ip configuration, at least
+          # one qemu channel has to be configured. As there are various options,
+          # error out and leave configuration to the user
+          unless machine.provider_config.channels.any? { |channel| channel[:target_name].start_with?("org.qemu.guest_agent") }
+            errors << "qemu agent option enabled, but no qemu agent channel configured: please add at least one qemu agent channel to vagrant config"
+          end
+        end
+
 
         machine.provider_config.disks.each do |disk|
           if disk[:path] && (disk[:path][0] == '/')
--- a/lib/vagrant-libvirt/driver.rb
+++ b/lib/vagrant-libvirt/driver.rb
@@ -3,6 +3,7 @@
 require 'fog/libvirt'
 require 'libvirt'
 require 'log4r'
+require 'json'
 
 module VagrantPlugins
   module ProviderLibvirt
@@ -58,7 +59,7 @@ module VagrantPlugins
 
         config = @machine.provider_config
 
-        @@system_connection = Libvirt::open_read_only(config.system_uri)
+        @@system_connection = Libvirt::open(config.system_uri)
         @@system_connection
       end
 
@@ -97,6 +98,12 @@ module VagrantPlugins
       def get_domain_ipaddress(machine, domain)
         if @machine.provider_config.qemu_use_session
           return get_ipaddress_from_system domain.mac
+        end
+
+        # attempt to get ip address from qemu agent
+        if @machine.provider_config.qemu_use_agent == true
+          @logger.info('Get IP via qemu agent')
+          return get_ipaddress_from_qemu_agent(domain, machine.id)
         end
 
         # Get IP address from dhcp leases table
@@ -143,6 +150,40 @@ module VagrantPlugins
           break if ip_address
         end
 
+        ip_address
+      end
+
+      def get_ipaddress_from_qemu_agent(domain, machine_id)
+        ip_address = nil
+        addresses = nil
+        dom = system_connection.lookup_domain_by_uuid(machine_id)
+        begin
+          response = dom.qemu_agent_command('{"execute":"guest-network-get-interfaces"}', timeout=10)
+          @logger.debug("Got Response from qemu agent")
+          @logger.debug(response)
+          addresses = JSON.parse(response)
+        rescue => e
+          @logger.debug("Unable to receive IP via qemu agent: [%s]" % e.message)
+        end
+
+        unless addresses.nil?
+          addresses["return"].each{ |interface|
+            if domain.mac == interface["hardware-address"]
+              @logger.debug("Found mathing interface: [%s]" % interface["name"])
+              if interface.has_key?("ip-addresses")
+                interface["ip-addresses"].each{ |ip|
+                  # returning ipv6 addresses might break windows guests because
+                  # winrm cant handle connection, winrm fails with "invalid uri"
+                  if ip["ip-address-type"] == "ipv4"
+                    ip_address = ip["ip-address"]
+                    @logger.debug("Return IP: [%s]" % ip_address)
+                    break
+                  end
+                  }
+              end
+            end
+          }
+        end
         ip_address
       end
 
--- a/lib/vagrant-libvirt/util/network_util.rb
+++ b/lib/vagrant-libvirt/util/network_util.rb
@@ -20,6 +20,7 @@ module VagrantPlugins
 
         def configured_networks(env, logger)
           qemu_use_session = env[:machine].provider_config.qemu_use_session
+          qemu_use_agent = env[:machine].provider_config.qemu_use_agent
           management_network_device = env[:machine].provider_config.management_network_device
           management_network_name = env[:machine].provider_config.management_network_name
           management_network_address = env[:machine].provider_config.management_network_address
--- a/tests/qemu_agent/Vagrantfile
+++ b/tests/qemu_agent/Vagrantfile
@@ -1,0 +1,13 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+#
+# frozen_string_literal: true
+
+Vagrant.configure("2") do |config|
+  config.vm.box = "generic/debian10"
+  config.vm.synced_folder ".", "/vagrant", disabled: true
+  config.vm.provider :libvirt do |libvirt|
+    libvirt.channel :type => 'unix', :target_name => 'org.qemu.guest_agent.0', :target_type => 'virtio'
+    libvirt.qemu_use_agent = true
+  end
+end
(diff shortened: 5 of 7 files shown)


lib/vagrant-libvirt/util/network_util.rb
```diff
@@ -48,7 +49,7 @@ def configured_networks(env, logger)
                   error_message: "#{management_network_address} does not include both an address and subnet mask"
           end
 
-          if qemu_use_session
+          if qemu_use_session or qemu_use_agent == true
```
electrofelix, line 52: I've a query on whether this is necessary or whether we should be able to use the qemu guest agent to retrieve the IP address for any guess that has it installed and enabled, and make the decision on what network is used for the management network to be handled explicitly rather than bundled as part of this option. Probably better to expand on this via the issue as there is more context there.

tests/qemu_agent/Vagrantfile
```diff
@@ -0,0 +1,13 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+#
+# frozen_string_literal: true
+
+Vagrant.configure("2") do |config|
+  config.vm.box = "generic/debian10"
+  config.vm.synced_folder ".", "/vagrant", disabled: true
+  config.vm.provider :libvirt do |libvirt|
+    libvirt.channel :type => 'unix', :target_name => 'org.qemu.guest_agent.0', :target_type => 'virtio'
```
electrofelix, line 10: Was wondering if you were going to update to add this as the default during the finalize if qemu_use_agent is true and there are no existing channels?

lib/vagrant-libvirt/driver.rb
```diff
@@ -58,7 +59,7 @@ def system_connection
 
         config = @machine.provider_config
 
-        @@system_connection = Libvirt::open_read_only(config.system_uri)
+        @@system_connection = Libvirt::open(config.system_uri)
```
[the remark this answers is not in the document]
electrofelix: So I think I've tracked down that the open_read_only would only ever work for local connections, and also it appears that due to libvirt being integrated with polkit for the last while, the username/password in the above connection set up would be ignored as polkit would enforce requesting auth interactively unless the user is in the expected groups.

So this likely does need to be fixed up for remote connections using `ssh`, but the proposed change isn't any more/less broken than what was there before.

electrofelix (commented):
Just want confirm if you are planning to move the required channel to be included as a default for the channels if none are defined during the finalize stage. Some notes added on the other areas.

abbbi:
> Just want confirm if you are planning to move the required channel to be included as a default for the channels if none are defined during the finalize stage. Some notes added on the other areas.

would make sense, but im not really sure where to put that at best.

abbbi:
> I think this possibly should check for @interface_network[:ip] being not nil or not an empty string rather than checking about the qemu_use_agent. I have to wonder what happens around this code if there are multiple networks on the host one of which is using the same name but is intended to be using dhcp local to the host rather than external that the machine itself is planning to use. I feel like I'm not really sure what happens here and what should happen, though maybe for now this is sufficient and try and prioritize some network unit tests for different scenarios around this code in the near future.

well, currently now sure about how to proceed, there are the following situations to consider:

1) the network might be defined by the user and already exists on the system: no ip address is set, in this case the check for the 
@interface_network[:ip_address] is never nil, bcs the module uses the default values:

```
 INFO create_networks: Using host-bridge at 192.168.121.0/24 as the management network nat is the mode
DEBUG create_networks: In config found network type forwarded_port options {:guest=>22, :host=>2222, :host_ip=>"127.0.0.1", :id=>"ssh", :auto_correct=>true, :protocol=>"tcp"}
DEBUG create_networks: Searching for network with options {:iface_type=>:private_network, :network_name=>"host-bridge", :ip=>"192.168.121.0", :netmask=>"255.255.255.0", :dhcp_enabled=>true, :forward_mode=>"nat", :guest_ipv6=>"yes", :autostart=>false, :bus=>nil, :slot=>nil, :mac=>"52:54:00:16:c0:80"}
DEBUG create_networks: looking up network with ip == 192.168.121.0
DEBUG create_networks: Checking that network name does not clash with ip
{:name=>nil, :ip_address=>#<IPAddr: IPv4:192.168.121.1/255.255.255.255>, :netmask=>"255.255.255.0", :network_address=>"192.168.121.0", :bridge_name=>nil, :domain_name=>nil, :ipv6_address=>nil, :ipv6_prefix=>nil, :created=>true, :active=>false, :autostart=>false, :guest_ipv6=>"yes", :libvirt_network=>#<Libvirt::Network:0x00005626d7e516a0 @connection=#<Libvirt::Connect:0x00005626d7af0588>>}
```

electrofelix:
I think maybe leave the networking as you currently have it. It'll have to be revisited at some point and I suspect the solution will involve delaying setting of the network address to use from being set in the configuration until the networks are being setup to allow flexibility of only setting it for a fallback. But need tests around this area that are outside the scope of this PR.

A comment beside the condition might be useful to remind us when coming back to this section that we know there is an issue there.

For setting the default channel, if you look at the following code:
https://github.com/vagrant-libvirt/vagrant-libvirt/blob/d1bad669b6e6a1c7fbddb29b0de02990d7158459/lib/vagrant-libvirt/config.rb#L900-L902

Can replace that with a something to inject the channel automatically. The only question is which of the following options to follow:
1. Always inject channel when `qemu_use_agent` is `true`, or
1. Only when no channels are defined (`qemu_use_agent` is `true` and `channels` is `UNSET_VALUE`, or
1. If `qemu_use_agent` is set and no channel with a target of `org.qemu.guest_agent.0` is defined

I'm assuming one of the latter 2, but I'm not familiar enough with this capability to understand what other values/settings could be used for it that would make sense. Possibly if going with option 2 as the least risky, can emit a message if `qemu_use_agent` is `true` and `channels` contain an entry to state that the code will not automatically inject the channel and if it's needed tell the user to add it to the definitions along with their existing definition?

abbbi (commented):
> Can replace that with a something to inject the channel automatically. The only question is which of the following options to follow:
> 
> 1. Always inject channel when `qemu_use_agent` is `true`, or
> 2. Only when no channels are defined (`qemu_use_agent` is `true` and `channels` is `UNSET_VALUE`, or
> 3. If `qemu_use_agent` is set and no channel with a target of `org.qemu.guest_agent.0` is defined
> 
> I'm assuming one of the latter 2, but I'm not familiar enough with this capability to understand what other values/settings could be used for it that would make sense. Possibly if going with option 2 as the least risky, can emit a message if `qemu_use_agent` is `true` and `channels` contain an entry to state that the code will not automatically inject the channel and if it's needed tell the user to add it to the definitions along with their existing definition?

i've now added code to at least validate if a channel is set and error out if qemu_use_agent is true but a channel is missing.
I think the used channel configuration can different between setups, some might prefer to use a local socket, but then there are things like apparmor and stuff that has to be taken care of. I think it is good to error out if its not set and provide one example in the README that works for most of the cases.

lib/vagrant-libvirt/driver.rb
```diff
@@ -167,13 +167,19 @@ def get_ipaddress_from_qemu_agent(domain, machine_id)
...
-              if ip.has_key?("ip-addresses")
-                ip_address = ip["ip-addresses"][0]["ip-address"]
-                @logger.debug("Return IP: [%s]" % ip_address)
-                break
+          addresses["return"].each{ |interface|
+            if domain.mac == interface["hardware-address"]
+              @logger.debug("Found mathing interface: [%s]" % interface["name"])
+              if interface.has_key?("ip-addresses")
+                interface["ip-addresses"].each{ |ip|
+                  # returning ipv6 addresses might break windows guests because
```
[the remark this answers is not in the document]
electrofelix: I suspect that it shouldn't return the ipv6 address as an array and instead it's likely there is bug elsewhere, possibly we need to return information in a `winrm_info` with ipv6 addresses wrapped for the hostname rather than returning `[ipv6]` as the address.
vagrant will fall back to using the ssh info if no winrm info is provided for the host:
https://github.com/hashicorp/vagrant/blob/cfa8722007b1b78ea06632db499386b5593416eb/plugins/communicators/winrm/helper.rb#L31-L38

What I'm not sure about is whether this would be fixed by changing how the ssh info is formatted for ipv6, or need to return a dedicated winrm info response.

I'd stick with your approach to only return an ipv4 address for now, until it's possible to work out what the correct behaviour is around ipv6.

lib/vagrant-libvirt/config.rb
```diff
@@ -949,6 +957,23 @@ def validate(machine)
...
+          # error out and leave configuration to the user
+          err = true
+          machine.provider_config.channels.each { |channel|
+            if channel[:target_name].start_with?("org.qemu.guest_agent")
+              err = nil
+            end
+          }
+          if err == true
+            errors << "qemu agent option enabled, but no qemu agent channel configured: please add at least one qemu agent channel to vagrant config"
+          end
```
electrofelix, line 973: I think I'd be inclined to use the `.any?` syntax here rather than `.each`
```suggestion
          unless machine.provider_config.channels.any? { |channel| channel[:target_name].start_with?("org.qemu.guest_agent") }
            errors << "qemu agent option enabled, but no qemu agent channel configured: please add at least one qemu agent channel to vagrant config"
          end
```

electrofelix (commented):
Small suggestion in the validation.

Also currently waiting for coveralls to fix there end to allow reporting unit tests, so might be tomorrow, but I'll look to keep an eye to retrigger once and will look to land this ahead of the other PRs (assuming you're happy)

electrofelix (approved):
lgtm, just need to wait for coveralls to sort itself out now.

electrofelix:
@abbbi looks like it's good to go ahead and squash merge whenever you are ready

electrofelix:
@abbbi squash or merge when ever you're happy

abbbi:
> @abbbi squash or merge when ever you're happy

mmhhh.. im not allowed to merge :)

electrofelix:
@abbbi oh, I thought I gave you sufficient privs to do that, I'll see what's needed to sort that

electrofelix:
@abbbi should be sorted now, I hope

lib/vagrant-libvirt/action/create_networks.rb
```diff
@@ -162,7 +162,7 @@ def handle_ip_option(env)
             if @interface_network[:created]
               # Just check for mismatch error here - if name and ip from
               # config match together.
-              if @options[:network_name] != @interface_network[:name]
+              if @options[:network_name] != @interface_network[:name] and @qemu_use_agent = false
```
electrofelix, line 165: @abbbi I missed that this is actually setting @qemu_use_agent to false rather than comparing, it should use `==`

abbbi: whoops :)

PR #1349, opened 2021-09-13
Fix variable in entrypoint.sh when determining user based on Vagrantfile ownership

mmguero:
In the entrypoint.sh script, the "$1" variable was used incorrectly when determining default user based on the ownership of the vagrantfile (or parents). The "$vagrantfile" variable should have been used instead, as it was just set.

Signed-off-by: SG <13872653+mmguero@users.noreply.github.com>

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -33,7 +33,7 @@ esac
 # from the current working directory anyway
 vagrantfile="${VAGRANT_VAGRANTFILE:-Vagrantfile}"
 path="$(pwd)"
-while [[ "$path" != "" && ! -e "$path/$1" ]]
+while [[ "$path" != "" && ! -e "$path/$vagrantfile" ]]
 do
     path=${path%/*}
 done


electrofelix:
Thanks, that does look like an obvious mistake.

electrofelix:
@mmguero can you amend the commit and force push to put the PR summary as the first line then a blank line and then wrap the remaining body to 72 characters. Can use the following trick to quickly wrap:
```
echo 'In the entrypoint.sh script, the "$1" variable was used incorrectly when determining default user based on the ownership of the vagrantfile (or parents). The "$vagrantfile" variable should have been used instead, as it was just set.

Signed-off-by: SG 13872653+mmguero@users.noreply.github.com' | fold -w 72 -s
```

Should look like:
```
Fix entrypoint.sh determining user based on Vagrantfile ownership

In the entrypoint.sh script, the "$1" variable was used incorrectly 
when determining default user based on the ownership of the vagrantfile 
(or parents). The "$vagrantfile" variable should have been used 
instead, as it was just set.

Signed-off-by: SG 13872653+mmguero@users.noreply.github.com
```

Makes it easier dealing with history when looking through it subsequently. I could the commit by squashing but it loses the original author information.

Issue #1357, opened 2021-09-24
support podman in addition to docker


ikke-t:
Instructions are made to run vagrant-libvirt with docker. Please guide the same for podman.

### Steps to reproduce
1. podman run -it --rm \
  -e LIBVIRT_DEFAULT_URI \
  -v /var/run/libvirt/:/var/run/libvirt/ \
  -v ~/.vagrant.d:/.vagrant.d \
  -v $(realpath "${PWD}"):${PWD} \
  -w $(realpath "${PWD}") \
  --network host \
  vagrantlibvirt/vagrant-libvirt:latest \
    vagrant status
2.
3.

### Expected behaviour
it should work

### Actual behaviour

```
IGNORE_MISSING_LIBVIRT_SOCK=true podman run -it --rm \
  -e LIBVIRT_DEFAULT_URI \
  -v /var/run/libvirt/:/var/run/libvirt/ \
  -v ~/.vagrant.d:/.vagrant.d \
  -v $(realpath "${PWD}"):${PWD} \
  -w $(realpath "${PWD}") \
  --network host \
  vagrantlibvirt/vagrant-libvirt:latest vagrant up
WARNING! Running as root, if this breaks, you get to keep both pieces
Starting with UID: 0, GID: 0
chown: invalid user: 'vagrant:users'
Unless you are using this to connect to a remote libvirtd it is
necessary to mount the libvirt socket in as /var/run/libvirt/libvirt-sock

Set IGNORE_MISSING_LIBVIRT_SOCK to silence this warning
error: failed switching to "vagrant": unable to find user vagrant: no matching entries in passwd file
```

### System configuration
**OS/Distro version:**:

RHEL8

**Libvirt version:**

libvirt-6.0.0-35.1.module+el8.4.0+11273+64eb94ef.x86_64

**Are you using upstream vagrant package or your distros package?**
Upstream / Distro
vagrant-2.2.7-1.x86_64

AverageHumanoid:
I was able to get Podman to work by manually creating `~/.vagrant.d/{boxes,data,tmp}` then adding

```
--security-opt label=disable \
--entrypoint /bin/bash \
```
to the run time options.

electrofelix:
@ikke-t can you try running with the edge tagged image `vagrantlibvirt/vagrant-libvirt:edge`, the latest image points to the latest release and there are a number of fixes on the main branch that should help.

[...]

PR #1359, opened 2021-09-25
Document running using Podman

AverageHumanoid:
Address #1357

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ vagrant(){
 ```
 
 ---
-To run with Podman you need to add
+To run with Podman you need to use a rootful container and add
 
 ```bash
   --entrypoint /bin/bash \
@@ -174,7 +174,7 @@ for example:
 
 ```bash
 vagrant(){
-  podman run -it --rm \
+  sudo podman run -it --rm \
   -e LIBVIRT_DEFAULT_URI \
   -v /var/run/libvirt/:/var/run/libvirt/ \
   -v ~/.vagrant.d:/.vagrant.d \
@@ -187,7 +187,6 @@ vagrant(){
   vagrant $@
 }
 ```
-
 
 Note that if you are connecting to a remote system libvirt, you may omit the
 `-v /var/run/libvirt/:/var/run/libvirt/` mount bind. Some distributions patch the local
--- a/README.md
+++ b/README.md
@@ -143,22 +143,51 @@ docker run -it --rm \
   -w $(realpath "${PWD}") \
   --network host \
   vagrantlibvirt/vagrant-libvirt:latest \
-    vagrant status
+  vagrant status
 ```
 
-It's possible to define an alias in `~/.bashrc`, for example:
+It's possible to define a function in `~/.bashrc`, for example:
 ```bash
-alias vagrant='
+vagrant(){
   docker run -it --rm \
-    -e LIBVIRT_DEFAULT_URI \
-    -v /var/run/libvirt/:/var/run/libvirt/ \
-    -v ~/.vagrant.d:/.vagrant.d \
-    -v $(realpath "${PWD}"):${PWD} \
-    -w $(realpath "${PWD}") \
-    --network host \
-    vagrantlibvirt/vagrant-libvirt:latest \
-    vagrant'
+  -e LIBVIRT_DEFAULT_URI \
+  -v /var/run/libvirt/:/var/run/libvirt/ \
+  -v ~/.vagrant.d:/.vagrant.d \
+  -v $(realpath "${PWD}"):${PWD} \
+  -w $(realpath "${PWD}") \
+  --network host \
+  vagrantlibvirt/vagrant-libvirt:latest \
+  vagrant $@
+}
+
 ```
+
+---
+To run with Podman you need to add
+
+```bash
+  --entrypoint /bin/bash \
+  --security-opt label=disable \
+```
+
+for example:
+
+```bash
+vagrant(){
+  podman run -it --rm \
+  -e LIBVIRT_DEFAULT_URI \
+  -v /var/run/libvirt/:/var/run/libvirt/ \
+  -v ~/.vagrant.d:/.vagrant.d \
+  -v $(realpath "${PWD}"):${PWD} \
+  -w $(realpath "${PWD}") \
+  --network host \
+  --entrypoint /bin/bash \
+  --security-opt label=disable \
+  docker.io/vagrantlibvirt/vagrant-libvirt:latest \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -143,23 +143,52 @@ docker run -it --rm \
...
+  vagrant $@
+}
+
+```
+
+---
+To run with Podman you need to add
+
+```bash
+  --entrypoint /bin/bash \
```
electrofelix, line 169: This skips https://github.com/vagrant-libvirt/vagrant-libvirt/blob/master/entrypoint.sh which does a number of set up steps that are normally required for things to work as expected e.g. have the internal user be in the correct group to access the libvirt socket. So you'll need to explain how this works as expected subsequently.

AverageHumanoid: I realized you need to run podman as root as well since a rootless container won't work. I think this is just temporary workaround until you can add official support/update `entrypoint.sh` but so far I've had no problem using this setup.

AverageHumanoid: The only effect I've encountered from this is any new files/folders created by Vagrant will be owned by root, any existing Vagrantfiles won't be affected which shouldn't break anything.

electrofelix: Have you tried the 'edge' version of the image without overriding the entrypoint? The goal of the entrypoint is to ensure that when vagrant is executed it is with the correct UID/GID so that the files remain readable/writeable by the user.

AverageHumanoid: I just tried the `edge` version which seems to work so I removed the `--entrypoint /bin/bash` argument.

README.md
```diff
@@ -143,21 +144,50 @@ docker run -it --rm \
   -w $(realpath "${PWD}") \
   --network host \
   vagrantlibvirt/vagrant-libvirt:latest \
-    vagrant status
+  vagrant status
```
electrofelix, line 147: I have a slight preference to keep the indent formatting for this and the commands below as:
```bash
<container-cli> \
  <arg-to-container-cli> \
    <arg-to-container-entrypoint>
```

As a subtle hint that the line `vagrant status` isn't being passed to the docker command, but rather what ever docker itself is executing, since I think a note is needed below on the change to the mount path as it might be useful information if rootless docker behaves similar to podman.

README.md
```diff
@@ -143,21 +144,50 @@ docker run -it --rm \
...
+
+for example:
+
+```bash
+vagrant(){
+  podman run -it --rm \
+  -e LIBVIRT_DEFAULT_URI \
+  -v /var/run/libvirt/:/var/run/libvirt/ \
+  -v ~/.vagrant.d/boxes:/vagrant/boxes \
+  -v ~/.vagrant.d/data:/vagrant/data \
```
electrofelix, line 182: I think probably a good idea to add an explanation below about this being different to the docker command, basically that since podman has the root user in the container mapped to the host user, it's possible to simply mount the needed data directories directly onto the target paths. While in the docker one it is mounting them to `/.vagrant.d` and the entrypoint is handling the mapping as part of setting up the correct user.

It's not perfect, because `~/.vagrant.d/Vagrantfile` won't be picked up in the current example if someone is using it, but perhaps such a note can help them know to add that to the list of the volume mounts.

electrofelix (commented):
last minor comments, think a small note on the reason behind the change in mount paths between docker and podman would be useful here for others.

README.md
AverageHumanoid: How about this explanation:

"Running Podman in rootless mode maps the `root` user inside the container to your host user so we need to bypass `entrypoint.sh` and mount persistent storage directly to `/vagrant`."

I did a bit of testing and was able to use a Vagrantfile in `~/.vagrant.d/` without problem.

electrofelix: That sounds good. I wasn't expecting that it could pick up the user `~/.vagrant.d/Vagrantfile` because I thought it used `ENV['VAGRANT_HOME']/Vagrantfile`, where the env var `VAGRANT_HOME` is altered inside the container to be `/vagrant/`, so that's good to know that it works.

AverageHumanoid: It should pickup whatever Vagrantfile is the current directory which is the purpose of 

```
  -v $(realpath "${PWD}"):${PWD} \
  -w $(realpath "${PWD}") \
```

right?

electrofelix: I think we're talking about 2 different Vagrantfile's. When vagrant loads it looks in multiple locations for Vagrantfiles and combines them together, so it reads and processes in the following order:
* `<box>/Vagrantfile`
* `${PWD}/Vagrantfile`
* `${VAGRANT_HOME}/Vagrantfile`

By default VAGRANT_HOME resolves to ~/.vagrant.d, which is also where the plugins get installed under.

As an example I have the following Vagrantfile in that location:
```
    libvirt.storage_pool_name = "vms"
  end

  if Vagrant.has_plugin?("vagrant-proxyconf")
    if !ENV['VAGRANT_ENABLE_PROXY'].nil?
        config.proxy.enabled  = true
        config.proxy.http     = ENV['http_proxy'] || ENV['HTTP_PROXY'] || nil
        config.proxy.https    = config.proxy.http
        config.proxy.no_proxy = ENV['no_proxy'] || ENV['NO_PROXY'] || nil
    else
        config.proxy.enabled  = false
    end
  end
end
```

I have faster storage available elsewhere other than the default storage pool and I created a storage pool there called `vms`. Since I don't want to edit each Vagrantfile that someone sends me, I use this to ensure the storage pool is overridden for me without having to think about it. Secondly I used to work from behind a proxy as well as sometimes need to validate that some things didn't need external access so I needed to be able to control whether the proxy env was passed through. Both of these get merged into every project Vagrantfile that I run subsequently.

So that is the Vagrantfile I was referring to previously about not being picked up automatically where as when using the entrypoint it is possible to mount in the entire ~/.vagrant.d directory and the entrypoint would remap the directories.

However now that I think of it, it doesn't remap the Vagrantfile if it exists, so it doesn't currently make difference. Another small thing for me to look at if I can tidy up the entrypoint so that it is useful to handle this sort of thing for podman based execution as well.

AverageHumanoid: Thanks for for clarification. My use case is not so complex that I need to combine multiple Vagrantfiles but I can see how that can be useful in your circumstance.

electrofelix:
made some minor adjustments to change how some of the info is grouped (toc gets auto-regenerated for me due to local helper plugin which showed that the subsequent header was getting mis-grouped as being a child of podman).

Squashing should still leave you as the author though

electrofelix:
Oh, I also switched the podman command back to using `latest` from edge, because with the override of the entrypoint, there is no issue with using the latest image anymore

PR #1361, opened 2021-09-27
#1341 fix condition check on @qemu_use_agent

abbbi:
As discussed in:

https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1343#pullrequestreview-764153975


electrofelix (approved):
look good to me

Issue #508, opened 2015-11-18
Destroying VMs after reboot failed


vills:
As VMs doesn't autostart after reboot, so no networks are active too.

```
 Name                 State      Autostart     Persistent
----------------------------------------------------------
 default              active     yes           yes
 t1_external          inactive   no            yes
 t1_internal          inactive   no            yes
 t2_external          inactive   no            yes
 t2_internal          inactive   no            yes
 vagrant-libvirt      inactive   no            yes
```

As a workaround, it should be something like `vagrant reload`. But if you'll try `vagrant destroy` after host-machine reboot, then you get error:
> Error while removing network vagrant-libvirt. Call to virNetworkDestroy failed: Requested operation is not valid: network is not active.

My opinion: *vagrant-libvirt* should handle such situation and correctly undefine networks if its not active.

P.S. All VMs destroyed successfully, but networks - no.

Some info:
```
~# vagrant plugin list
vagrant-libvirt (0.0.32)
vagrant-omnibus (1.4.1)
vagrant-share (1.1.4, system)
```
```
~# dpkg -l | egrep 'libvirt|vagrant'
ii  libvirt-bin                         1.2.2-0ubuntu13.1.14                  amd64        programs for the libvirt library
ii  libvirt-dev                         1.2.2-0ubuntu13.1.14                  amd64        development files for the libvirt library
ii  libvirt0                            1.2.2-0ubuntu13.1.14                  amd64        library for interfacing with different virtualization systems
ii  vagrant                             1:1.7.4                               amd64        no description given
```
```
~# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.3 LTS
Release:	14.04
Codename:	trusty
```

RyanHartje:
I'm having the same issue on Centos 7, here is my debug info:

```sh
# vagrant plugin list
vagrant-libvirt (0.0.32)
vagrant-share (1.1.5, system)
```

```sh
# rpm -qa|egrep "libvirt|vagrant"
libvirt-daemon-driver-network-1.2.17-13.el7_2.2.x86_64
libvirt-daemon-driver-secret-1.2.17-13.el7_2.2.x86_64
vagrant-1.8.1-1.x86_64
libvirt-daemon-1.2.17-13.el7_2.2.x86_64
libvirt-daemon-driver-nwfilter-1.2.17-13.el7_2.2.x86_64
[...]

PR #1381, opened 2021-10-14
Add a :libvirt__always_destroy flag to network definitions

jamie01:
This ensures that a network is always appended to the created_networks
file, to avoid the case where a domain that created the network gets
destroyed first during cleanup, therefore skipping the network destroy
because some other domain is still using it. Since the second domain did
not create the network, the existing behaviour is that it gets left
behind.

Seting :libvirt__always_destroy to true will ensure that the domain will
destroy the network even if it did not create it, if there are no other
users of the network remaining.

----

This addresses an issue we experience in multi-node setups, where often the first node to create a network is also the first to be destroyed during a vagrant destroy, hence it leaves the network up due to there being another domain using it.

Since the created_networks file is the current way of tracking networks I have reused this to allow a flag that causes it to be written even if a network already exists when the machine is created. This has the effect of ensuring the network is always destroyed, if all machines that use it set this flag.

Any comments appreciated if I could think about this in a different way! :)

Fixes: #508

### DIFF
--- a/lib/vagrant-libvirt/action/handle_box_image.rb
+++ b/lib/vagrant-libvirt/action/handle_box_image.rb
@@ -34,10 +34,11 @@ module VagrantPlugins
             box_format = env[:machine].box.metadata['format']
             HandleBoxImage.verify_box_format(box_format)
 
+            image_path = HandleBoxImage.get_box_image_path(env[:machine].box, 'box.img')
             env[:box_volume_number] = 1
             env[:box_volumes] = [{
-              :path => HandleBoxImage.get_box_image_path(env[:machine].box, 'box.img'),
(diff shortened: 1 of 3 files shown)


electrofelix:
I think I'll want to test some of the behaviour around halt and resume machines using the same network, but it might be that the flag is unnecessary as this looks like a fix for a number of logged issues over time.

Hesitation is due to the lack of unit/integration tests around network creation/destroy.

electrofelix:
@jamie01 just means that even though it passes the unit tests need to do some manual experiments with it locally to see what happens when other directories contains machines using the same network which are in a halted state when other machines with this setting enabled are destroyed. Will try to get through the testing today.

PR #1385, opened 2021-10-18
Add basic support to control serial console settings

Itxaka:
Support limited user defined serial console settings to redirect boot
time messages to a log file.

Closes: #695
Authored-by: @electrofelix
Signed-off-by: Itxaka <igarcia@suse.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -910,6 +910,9 @@ starts with `libvirt__` string. Here is a list of those options:
   If not specified the default is 'false'.
 * `:bus` - The bus of the PCI device. Both :bus and :slot have to be defined.
 * `:slot` - The slot of the PCI device. Both :bus and :slot have to be defined.
+* `:libvirt__always_destroy` - Allow domains that use but did not create a
+  network to destroy it when the domain is destroyed (default: `true`). Set to
+  `false` to only allow the domain that created the network to destroy it.
 
 When the option `:libvirt__dhcp_enabled` is to to 'false' it shouldn't matter
 whether the virtual network contains a DHCP server or not and vagrant-libvirt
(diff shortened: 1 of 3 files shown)


electrofelix:
I think this is the first time I've tried squash merge with a PR with a commit authored by someone other than the PR creator, so not entirely sure what will be the final result.

mazzy89:
I'm having the permission issue and reading the discussion here is not very clear to me exactly how you are proposing to solve. Could you advice?

electrofelix:
@mazzy89 Previously I was hoping that simply pre-creating the file would be sufficient for it to retain the ownership, however recent testing suggests that it ends up owned by root.

I've done some experimenting with security labels and also with some settings and I may have a workaround. That is, if `append='on'` is set on the source element, and the file exists, libvirt/qemu appears to leave the file with the initial user/group and permissions, it's only if libvirt/qemu looks to replace the file when `append='off'` (which is the default if append is not set).

Updating the code to always add `append='on'` when this is enabled appears to work, but it's unclear to me if the output still gets added to the log. I can put up a draft PR that will generate a gem file that can be installed locally if you are in a position to validate?

mazzy89:
@electrofelix to understand better `who` (which user) does make an attempt to create the file? is it `qemu` or `libvirt`?

electrofelix:
@mazzy89 I believe it's libvirt, based on when I look to see the qemu process started, there is no reference to the log file instead the command line contains `-chardev file,id=charserial0,path=/dev/fdset/3,append=on -device isa-serial,chardev=charserial0,id=serial0`, so I'd presume that it is libvirt that created the `/dev/fdset/3` for qemu to stream back the output sent to the device in the VM and it is therefore libvirt that is writing out to the file.

PR #1392, opened 2021-11-03
Fix a FrozenError in prepare_nfs_settings.rb

rledisez:


electrofelix (approved):
Thanks, probably missed this due to missing tests in this area.

Issue #1342, opened 2021-09-09
VM creation fails 2nd time


Yajo:
### Steps to reproduce
1. `vagrant up`
2. `vagrant destroy`
3. `vagrant up`

### Expected behaviour
VM properly setup, with SSH and internet connectivity.

### Actual behaviour
VM only works 1st time, not 2nd.

**Workaround 1**: logout, login. Then it works again 1st time.

**Workaround 2**: randomize vm name, because it works always the 1st time per vm name.

### System configuration
**OS/Distro version:**: Fedora 34 Silverblue.

**Libvirt version:** libvirt-daemon-7.0.0-6.fc34.x86_64

**Output of `vagrant version; vagrant plugin list`:**
Installed Version: 2.2.16

Vagrant was unable to check for the latest version of Vagrant.
Please check manually at https://www.vagrantup.com
vagrant-libvirt (0.5.3, system)

**Attach Output of `VAGRANT_LOG=debug vagrant ... --provider=libvirt >vagrant.log 2>&1`**

[up-destroy-up.log](https://github.com/vagrant-libvirt/vagrant-libvirt/files/7135351/up-destroy-up.log)

**A Vagrantfile to reproduce the issue:**
```
Vagrant.configure('2') do |config|
  if Vagrant.has_plugin?('vagrant-cachier')
    
    config.cache.scope = 'machine'
    
  end

  config.vm.define "server0" do |c|
    ##
    # Box definition
    ##
    c.vm.box = "generic/ubuntu2004"
    
    ##
    # Config options
    ##
    c.vm.synced_folder ".", "/vagrant", disabled: true
    c.ssh.insert_key = true
    c.vm.hostname = "server0"

    ##
    # Network
    ##
    
    c.vm.network "public_network", dev: "virbr0", mode: "bridge", type: "bridge"
    
    ##
    # Provider
    ##
    c.vm.provider "libvirt" do |libvirt, override|
      libvirt.memory = 512
      libvirt.cpus = 2
      libvirt.management_network_autostart = true
      libvirt.management_network_mode = "open"
      libvirt.qemu_use_session = true
    end
  end
end
```

**Are you using upstream vagrant package or your distros package?**
Distro

Yajo:
Analyzing the logs, comparing the 1st with the 2nd times `up` is executed, I see these interesting diffs.

2nd time it doesn't add mac address for virbr0:

```diff
[...]

PR #1396, opened 2021-11-07
Reorder qemu agent usage for use with sessions

electrofelix:
Adjust the order of checks around use of qemu sessions to allow use of
the agent as a priority when enabled, which should remove the need to
retrieve the address from the system connection when enabled.

Additionally adjust the call to the agent to ensure it uses the default
connection to retrieve the correct domain, rather than forcing the
system connection, which will fail to find the domain if it was created
via a user session.

Add tests that validate most of this behaviour, as well as resulting in
some minor fixes around downcasing the mac address for comparisons, and
also using instance mocks with rspec instead of pure doubles to help
catch false positives where mocks are allowing calls that done exist.

Fixes: #1342

### DIFF
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -970,7 +970,7 @@ module VagrantPlugins
         # Additional QEMU commandline environment variables
         @qemu_env = {} if @qemu_env == UNSET_VALUE
 
-        @qemu_use_agent = true if @qemu_use_agent != UNSET_VALUE
+        @qemu_use_agent = false if @qemu_use_agent == UNSET_VALUE
 
         @serials = [{:type => 'pty', :source => nil}] if @serials == []
       end
@@ -986,6 +986,9 @@ module VagrantPlugins
           end
         end
 
+        unless @qemu_use_agent == true || @qemu_use_agent == false
+          errors << "libvirt.qemu_use_agent must be a boolean."
+        end
 
         if @qemu_use_agent == true
           # if qemu agent is used to optain domain ip configuration, at least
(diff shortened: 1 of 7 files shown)


Yajo:
This one still asks for password:

https://user-images.githubusercontent.com/973709/140706189-3e55a1a9-8563-4c61-aebb-0e2aadfb12af.mp4

[vagrant.log](https://github.com/vagrant-libvirt/vagrant-libvirt/files/7495285/vagrant.log)

To test it, I did:

```sh
vagrant plugin uninstall vagrant-libvirt
vagrant plugin install vagrant-libvirt
rm -Rf ~/.vagrant.d/gems/3.0.2/gems/vagrant-libvirt-0.6.3/
git clone https://github.com/vagrant-libvirt/vagrant-libvirt.git ~/.vagrant.d/gems/3.0.2/gems/vagrant-libvirt-0.6.3/
cd ~/.vagrant.d/gems/3.0.2/gems/vagrant-libvirt-0.6.3/
git fetch origin refs/head/1396/head
git checkout FETCH_HEAD
```

Just in case I did something wrong.

electrofelix:
@Yajo looks like you are using the system connection rather than the session, from the log file:
```
 INFO driver: Connecting to Libvirt (qemu:///system) ...
```

When using the direct plugin you need to either set `qemu_use_session` in the Vagrantfile, or set export the following env var `export LIBVIRT_DEFAULT_URI=qemu:///session` to your shell before running vagrant up, to get the same behaviour as the distro packaged plugin.

If you have either of those set and the log message above is still appearing instead of containing `qemu:///session`, then there is definitely another bug with regarding to how the setting is getting picked up.

Yajo:
Yes, you're right. Changed it and it finally works! 🎉 😻 

[Vagrantfile](https://github.com/vagrant-libvirt/vagrant-libvirt/files/7496519/Vagrantfile.txt)

Yajo (commented):
I think it would be nice if you include in the `README.md` some info about how to use this plugin flawlessly on Fedora, now that it finally works.

PR #1397, opened 2021-11-08
Switch back to read-only system connection

electrofelix:
With the rework to use the default connection to retrieve the domain to 
allow running of the qemu agent, no longer necessary to use a read 
write system connection, and can revert to a read-only connection that 
requires less privileges.

This should remove an occurrence of polkit requesting a password when 
bringing a VM up using a session connection.


Yajo:
Would you be able to push a new release with this patch? Thanks :)

PR #1413, opened 2021-12-02
Make destroy network locale independent

Elyfler:
My machine uses French locales. 

When I try to run a `vagrant destroy -f mymachine`, the 'Network not found' error is raised because my error message is in French.

The comparison is then `if 'Réseau non trouvé' == 'Network not found'` as 'Network not found' is hardcoded which obviously fails everytime.

I could switch to an english locale but as it is a quick fix and may be used by others, I chose to switch from message comparison to libvirt_code comparison as eveything was already there, making it work with any locale.


electrofelix (approved):
Thanks, that's clearly a much better approach.

PR #1414, opened 2021-12-02
Double the UDPSocket in cases where there's no network

dcermak:
In PrepareNFSSettings.read_host_ip we use a UDPSocket to connect to the host in
question. However, this fails completely if the system executing the tests has
no network adapters enable, like builders in OBS or Koji. If we double the calls
to this, then we no longer face that issue.


electrofelix:
Thanks, had a quick look at OBS there and I'm wondering if there is anyway for this project to trigger a build run automatically after merge and get a status check report to pick up if something has been added that will cause problems subsequently when trying to build a package of this project?

Issue #1415, opened 2021-12-02
Unit tests fail in isolated OBS workers


dcermak:
**Describe the bug**
I am currently trying to update vagrant-libvirt in openSUSE from 0.5.3 to 0.6+ and am facing a plethora of test failures, for example:
```
[   60s]   1) VagrantPlugins::ProviderLibvirt::Action#action_halt not created should execute without error                                                                                                                                   
[   60s]      Failure/Error: expect { runner.run(subject.action_halt) }.not_to raise_error                                                                                                                                                   
[   60s]                                                                                                                                                                                                                                     
[   60s]        expected no Exception, got #<RSpec::Mocks::MockExpectationError: #<InstanceDouble(::Libvirt::Connect) (anonymous)> received unexpected message :libversion with (no args)> with backtrace:                                   
[   60s]          # ./lib/vagrant-libvirt/cap/synced_folder_virtiofs.rb:29:in `usable?'                                                                                                                                                      
[   60s]          # /usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/builtin/mixin_synced_folders.rb:39:in `block in default_synced_folder_type'                                                                               
[   60s]          # /usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/builtin/mixin_synced_folders.rb:38:in `each'                                                                                                              
[   60s]          # /usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/builtin/mixin_synced_folders.rb:38:in `default_synced_folder_type'                                                                                        
[   60s]          # /usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/builtin/mixin_synced_folders.rb:168:in `synced_folders'                                                                                                   
[   60s]          # /usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/machine.rb:631:in `synced_folders'                                                                                                                               
[   60s]          # /usr/share/vagrant/gems/gems/vagrant-2.2.19/plugins/kernel_v2/config/vm.rb:730:in `validate'                                                                                                                             
[   60s]          # /usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/config/v2/root.rb:71:in `block in validate'                                                                                                                      
[...]

PR #1416, opened 2021-12-03
Allow the connection.client.libversion call

dcermak:
When the tests are executed and rsync is not installed on the machine executing the tests, then vagrant-libvirt will automatically fallback to virtiofs or 9p. Both of these perform the following call to check the libvirt version:
```Ruby
libvirt_version = machine.provider.driver.connection.client.libversion
```
However, this mock was not setup and thus those tests would fail.

This fixes https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1415


electrofelix:
@pvalena I think you'll want this patch as well for the latest RPM, it should fix most of the issues not relating to changing the default connection to enable session instead of system.

PR #1418, opened 2021-12-03
Make simplecov a source development dependency only

electrofelix:
Remove the need for downstream packaging to either require or patch
out simplecov from existing files by making it an optional development
dependency.


electrofelix:
@dcermak would this patch help remove the need to carry the patch https://build.opensuse.org/package/view_file/openSUSE:Factory/vagrant-libvirt/0001-Remove-simplecov-from-tests.patch?expand=1

dcermak:
@electrofelix yes it does, thanks!

PR #1424, opened 2021-12-11
Reduce patching for distro default session use

electrofelix:
Reduce the patching needed should a distro wish to switch the default
from using the system connection by default to using a session
connection by default.

Should now only require patching the default value and a single test
checking the defaults.


electrofelix:
@pvalena I think if you also carry this PR ahead of your current patch to use session by default should vastly reduce the amount needed to switch from system to session going forward.

pvalena:
Thanks!

PR #1425, opened 2021-12-17
Management network keep option

ssasso:
Starting from version *0.7.0* (https://newreleases.io/project/github/vagrant-libvirt/vagrant-libvirt/release/0.7.0), *always_destroy* is set to *true* by default for any network.

However, in certain cases, it would be useful to keep the management network, especially when it has been already configured (externally) with Static IP leases.

This option allows to change the *always_destroy* behaviour for the management network.

### DIFF
--- a/README.md
+++ b/README.md
@@ -978,6 +978,8 @@ used by this network are configurable at the provider level.
 * `management_network_domain` - Domain name assigned to the management network.
 * `management_network_mtu` - MTU size of management network. If not specified,
   the Libvirt default (1500) will be used.
+* `management_network_keep` - Starting from version *0.7.0*, *always_destroy* is set to *true* by default for any network.
+  This option allows to change this behaviour for the management network.
 
 You may wonder how vagrant-libvirt knows the IP address a VM received.  Libvirt
 doesn't provide a standard way to find out the IP address of a running domain.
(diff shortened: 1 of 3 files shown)


electrofelix:
Looks like you might need to rebase as you have a commit included from a previous PR that was merged.

electrofelix:
@ssasso just realised after merging that you mention that it's useful to not destroy networks that were created externally, I think that itself is another bug as should only destroy networks that have been created.

Probably there is another mistake in how the network is written out to track whether it was created by vagrant-libvirt or not. Can you confirm that destroying a custom managed network definition is the main problem you are looking to solve here?

ssasso:
@electrofelix correct, I'm trying to solve the fact that networks created externally (in particular, the default "vagrant-libvirt" network) should not be deleted.
However, to my understanding that was correct according to "Destroying domains will now attempt to always remove associated networks that are no longer in use. If this cases any issues set :libvirt__always_destroy => false on the impacted interfaces and report an issue."

electrofelix:
The intent was that destroying networks should only remove associated networks that were created though. The problem being solved there was that if multiple Vagrantfiles or a multi machine Vagrantfile created a network, only the first machine that was brought up would perform the destroy because it was associated with that one. Failing to teardown in the reverse order would then leave the network created as it would skip the delete if it's still in use and only attempt it with the machine that originally created it when any could have created it.

Need to detect that the network wasn't created by vagrant-libvirt even if it uses the same name.

PR #1434, opened 2022-01-17
Make disabling and configuring memballoon possible

rubenk:
The memballoon settings didn't have any effect

### DIFF
--- a/lib/vagrant-libvirt/action/create_domain.rb
+++ b/lib/vagrant-libvirt/action/create_domain.rb
@@ -133,6 +133,12 @@ module VagrantPlugins
           # RNG device passthrough
           @rng = config.rng
 
+          # Memballoon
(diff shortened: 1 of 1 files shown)


electrofelix:
You PR doesn't impact functionality, only reporting to the console whether the settings are enabled, which doesn't align with the summary and description provided, so I'm wondering what problem are you trying to solve.

electrofelix:
Just re-read the code there, I see what it's doing now. Hasn't realise the variables were never defined

rubenk:
The funny thing is there are unit tests for memballoon and they passed before this change. I haven't looked too deeply into why this is though.

electrofelix:
They are setting the domain directly instead of setting the config object, definitely worth fixing up at some point

PR #1436, opened 2022-01-17
Fix static ipv6 ips on private networks

rubenk:
When the :ip is ipv6, set the 'static6' type, so that we select the
correct _static6.erb template to configure the interface in the guest.

With the following configuration, ipv6 works out of the box on CentOS 8:

  (1..2).each do |i|
    config.vm.define "node#{i}" do |node|
      node.vm.hostname = "node#{i}"

      node.vm.network :private_network,
        :ip => "fc00::#{i+9}",
        :netmask => "64",
        :libvirt__dhcp_enabled => false,
        :libvirt__forward_mode => "veryisolated",
        :libvirt__network_name => "test"
    end
  end

Fixes: #914

### DIFF
--- a/lib/vagrant-libvirt/action/create_network_interfaces.rb
+++ b/lib/vagrant-libvirt/action/create_network_interfaces.rb
@@ -216,6 +216,9 @@ module VagrantPlugins
                   gateway: options[:gateway],
                   route: options[:route]
                 }.merge(network)
+                if IPAddr.new(options[:ip]).ipv6?
+                  network[:type] = :static6
+                end
               else
                 network[:type] = :dhcp
               end


electrofelix:
That possibly fixes a few issues, I'll try to link them later so those that ran into the issue are alerted to the fix landing

rubenk:
alright, thanks. I just tested this with Debian 11 and there it works as well.

Note that this only works if you set the netmask explicitly. It would be nice if we could parse :ip "fc00::10/64" as well.

dereknola:
@electrofelix When is the next release coming out? I would love to have this PR, right now I am stuck manually building.

electrofelix:
@dereknola finally done there this evening. I've been trying to wait until I thought I would have time to respond to any critical issues that might appear shortly after release.

Issue #529, opened 2015-12-22
Support vagrant 1.8 snapshots


varac:
Vagrant 1.8 added a feature to make snapshots.

from https://github.com/mitchellh/vagrant/blob/master/CHANGELOG.md:

```
Snapshots: The vagrant snapshot command can be used to checkpoint and restore point-in-time snapshots.
```

Using latest vagrant-libvirt release (0.0.32), i get: 

```
--- vms/jessie » vagrant snapshot list  
==> jessie: This provider doesn't support snapshots.
==> jessie: 
==> jessie: This may be intentional or this may be a bug. If this provider
==> jessie: should support snapshots, then please report this as a bug to the
==> jessie: maintainer of the provider.

```

I'm using sahara for working with snapshots, so this is not urgent, but a feature request.

shaleh:
Is calling the equivalent of ``virsh snapshot-create-as myvm snapshot1 "snapshot1 description"`` sufficient here? Or do we need some kind of Vagrant variable setting which type of snapshot to use.

infernix:
For the base box this shouldn't be very difficult to implement since it's always qcow2. There are some unanswered questions though for additional disks:

- Should we snapshot additional disk together with the base box snapshot? 
- If the answer is maybe, do we want to make that an option for each extra disk? What default value?
- What happens when a snapshot is made for a disk of type `raw`? Libvirt creates a `qcow2` snapshot of it going forward but not at all sure what happens on the next `vagrant up` in this case.
- Do we `sync` before snapshotting live outside of whatever `libvirt` does, and if so, how?
- Do we need to do `atomic` snapshots, and if that fails, how do we handle failure?

I am also not sure about the difference between `snapshot-create-as ...` and `snapshot-create-as ... --live. This will need some testing.

shaleh:
I am trying to imagine a case where I would want the primary disk and secondary disks to not be in sync. Not having much success.

As for the other questions I do not have enough libvirt experience to make a reasonable suggestion.

infernix:
- Secondary disk is shared
- Secondary disk is part of something like ceph/gluster/other SDS
- Secondary disk contains generated data across multiple test iterations

Probably rare, and likely unexpected based on `vagrant snapshot`functionality description, so we are probably fine with assuming snapshots are for the entire domain, at least initially.

[...]

PR #840, opened 2017-12-11
Add provider-native snapshot support 

randomvariable:
Resolves #529

PR #1456, opened 2022-02-25
Add support for snapshots

r0x0d:
This PR is intended to continue the work left by @randomvariable,
addressing the fewer comments left there.

This also tries to resolve the #529.

Signed-off-by: Rodolfo Olivieri <rodolfo.olivieri3@gmail.com>


lib/vagrant-libvirt/action.rb
```diff
@@ -384,6 +390,51 @@ def self.action_ssh_run
...
+          b.use ConfigValidate
+          b.use Call, IsCreated do |env, b2|
+            unless env[:result]
+              raise Vagrant::Errors::VMNotCreatedError
+            end
+            
+            # Using as an example the virtualbox plugin, we don't delete the
+            # snapshot after restoring it 
+            # https://github.com/hashicorp/vagrant/blob/main/plugins/providers/virtualbox/action/snapshot_restore.rb
+            b2.use SnapshotRestore
```
r0x0d, line 419: In the original PR, #840, there was a comment questioning that if it's really necessary to delete the snapshots after restoring then. 

I removed that piece of code here because following the example of the virtualbox plugin in vagrant repository, they don't delete the snapshot after restoring them. 

Link to comment: https://github.com/vagrant-libvirt/vagrant-libvirt/pull/840#discussion_r156145695
Link to the snapshot_restore file in virtualbox plugin: https://github.com/hashicorp/vagrant/blob/main/plugins/providers/virtualbox/action/snapshot_restore.rb

r0x0d:
As I'm not 100% familiar with ruby and unit testing in this language, I tried my best to add a basic test case for the snapshot plugin action. 

Let me know if there is anything I can improve here.

locales/en.yml
```diff
@@ -181,6 +181,16 @@ en:
       management_network_required: |-
         Management network can't be disabled when VM use box.
         Please fix your configuration and run vagrant again.
+      snapshot_deletion_error: |-
+        Error while deleting snapshot: %{error_message}.
```
r0x0d, line 185: So, I'm getting an InterpolationError here because of this variable. Everything looks corret to me... What might be causing that?

electrofelix: When raising the exception, need to pass something as the `error_message` value, see how the following is raised:
https://github.com/vagrant-libvirt/vagrant-libvirt/blob/1b8a70163bef05759c42bb7912810c6d8fc22c19/lib/vagrant-libvirt/action/create_network_interfaces.rb#L33-L36

electrofelix: I thought it was something simple like forgetting to pass it, but it's not. Still looking.

lib/vagrant-libvirt/action.rb
r0x0d: yeah, it looked strange to me to delete the snapshot after a restore (thought about what if the user wants to restore the snapshot multiple times? It couldn't in the old PR as it would be deleted after the first one).

locales/en.yml
r0x0d: I will try to take a look into that this week when I have some time. I tried to look it up when I opened the PR, but I completely forgot about this... If you find out why let me know and I can fix it as well.

electrofelix:
Now that I've pushed 0.8.0 out the door, I'll look to take a closer look at getting this merged in about a week. Just want to allow a bit of time for any critical bugfixes to 0.8.0 to land on mainline and ship.

r0x0d:
> Now that I've pushed 0.8.0 out the door, I'll look to take a closer look at getting this merged in about a week. Just want to allow a bit of time for any critical bugfixes to 0.8.0 to land on mainline and ship.

No problem! I will be looking forward to this. Thanks a lot!

electrofelix (dismissed):
Small adjustment in the driver code to pass around the `machine` rather than the `machine_id` as this allows surfacing the machine name in errors, which is useful for those working with multi machine Vagrantfile's.

Additionally I think one integration test needs to be added, can be based on the complex packaging test:
https://github.com/vagrant-libvirt/vagrant-libvirt/blob/e65aec6b0bbcb39d384c0b50d6d061d473fb0de5/tests/runtests.bats#L182-L217

Will want to exercise the steps:
- bring up a machine
- create a snapshot
- create or copy a file (can be empty) in the guest (and confirm it exists)
- restore to snapshot and confirm file no longer exists
- list snapshots
- delete snapshot and confirm no longer exists (presumably by listing and comparing to previous)

It's possible the list and delete may be better split into a separate integration test, that's more a call when writing whether it appears the test is getting excessively long.

This will ensure that the functionality is working as expected.

r0x0d:
Thanks for your review, @electrofelix! I will work on this as soon as possible!

r0x0d:
@electrofelix, while testing the changes you mentioned, I started receiving the following error. 

```
Traceback (most recent call last):
        14: from /usr/bin/vagrant:23:in `<main>'
        13: from /usr/bin/vagrant:23:in `load'
        12: from /usr/share/rubygems-integration/all/gems/vagrant-2.2.14/bin/vagrant:200:in `<top (required)>'
        11: from /usr/share/rubygems-integration/all/gems/vagrant-2.2.14/lib/vagrant/environment.rb:290:in `cli'
        10: from /usr/share/rubygems-integration/all/gems/vagrant-2.2.14/lib/vagrant/cli.rb:67:in `execute'
         9: from /usr/share/rubygems-integration/all/gems/vagrant-2.2.14/plugins/commands/snapshot/command/root.rb:65:in `execute'
         8: from /usr/share/rubygems-integration/all/gems/vagrant-2.2.14/plugins/commands/snapshot/command/save.rb:39:in `execute'
         7: from /usr/share/rubygems-integration/all/gems/vagrant-2.2.14/lib/vagrant/plugin/v2/command.rb:232:in `with_target_vms'
         6: from /usr/share/rubygems-integration/all/gems/vagrant-2.2.14/lib/vagrant/plugin/v2/command.rb:232:in `each'
         5: from /usr/share/rubygems-integration/all/gems/vagrant-2.2.14/lib/vagrant/plugin/v2/command.rb:243:in `block in with_target_vms'
         4: from /usr/share/rubygems-integration/all/gems/vagrant-2.2.14/plugins/commands/snapshot/command/save.rb:53:in `block in execute'
         3: from /usr/share/rubygems-integration/all/gems/vagrant-2.2.14/lib/vagrant/capability_host.rb:111:in `capability'
         2: from /usr/share/rubygems-integration/all/gems/vagrant-2.2.14/lib/vagrant/capability_host.rb:111:in `call'
         1: from /root/.vagrant.d/gems/2.7.4/gems/vagrant-libvirt-0.6.4.pre.55/lib/vagrant-libvirt/cap/snapshots.rb:7:in `snapshot_list'
/root/.vagrant.d/gems/2.7.4/gems/vagrant-libvirt-0.6.4.pre.55/lib/vagrant-libvirt/driver.rb:163:in `list_snapshots': undefined method `list_snapshots' for #<Fog::Libvirt::Compute::Server:0x0000555d9c39e1b8> (NoMethodError)
```

Do you know anything about this?

electrofelix:
@r0x0d I hope I didn't confuse the `domain` variable name with direct access to the libvirt object when doing the original review. I suspect I have and I need to check what correct call to get access to that from the fog-libvirt domain (server) object is.

lib/vagrant-libvirt/driver.rb
```diff
@@ -125,6 +140,65 @@ def get_domain_ipaddress(machine, domain)
...
+      end
+
+      def create_new_snapshot(machine, snapshot_name)
+        snapshot_desc = <<-EOF
+        <domainsnapshot>
+          <name>#{snapshot_name}</name>
+          <description>Snapshot for vagrant sandbox</description>
+        </domainsnapshot>
+        EOF
+        get_domain(machine).snapshot_create_xml(snapshot_desc)
```
electrofelix, line 181: I made a mistake, the domain returned above by `get_client_domain` is not the same domain as `get_domain`, in fact it's possible that function was badly named and should have been `get_server`, since it is the fog server object that is returned there.

So will need to switch back, however can you make sure to pass the `machine` variable to `get_client_domain` and make that function private for now until sure it should be exposed directly.

lib/vagrant-libvirt/driver.rb
```diff
@@ -65,6 +65,21 @@ def system_connection
...
+          if e.libvirt_code == ProviderLibvirt::Util::ErrorCodes::VIR_ERR_NO_DOMAIN
+            @logger.debug("machine #{mid} not found #{e}.")
+            return nil
+          else
+            raise e
+          end
+        end
+
+        domain
+      end
```
electrofelix, line 81: Turns out this is returning a different object to `get_domain` below.

Suggest renaming slightly to make it a little more obvious for future me that it's not the same and is retrieving the libvirt object itself. Also could you make this method private for now and can decide in the future to make it publicly accessible.
```suggestion
      def get_libvirt_domain(machine)
        begin
          libvirt_domain = connection.client.lookup_domain_by_uuid(machine.id)
        rescue Libvirt::RetrieveError => e
          if e.libvirt_code == ProviderLibvirt::Util::ErrorCodes::VIR_ERR_NO_DOMAIN
            @logger.debug("machine #{machine.name} not found #{e}.")
            return nil
          else
            raise e
          end
        end

        libvirt_domain
      end
```

electrofelix (commented):
Sorry, I made a mistake in thinking the reference to `domain` in `get_client_domain` was the same domain as in `get_domain`, different objects returned. Suggested some renames to the method and variables contained to prevent future me making the same dumb mistake.

r0x0d:
Just passing by to tell that I will continue this when I have some time left! This PR is not dead, I just had some problems with vagrant on my F36.

electrofelix:
@r0x0d anything that I can help with? There is a script at https://github.com/vagrant-libvirt/vagrant-libvirt-qa/blob/0771dcb057be7d1c361069abcfcfa7e13c04ed32/scripts/install.bash that verifies getting vagrant and vagrant-libvirt installed on a number of versions of fedora

electrofelix:
@r0x0d are you using the distro vagrant or upstream vagrant? I have some better steps when working with the distro vagrant that ensures the correct dependency packages get installed while allowing for the latest vagrant-libvirt. Or are you trying to use vagrant direct from source as in full development mode. If the latter it's probably easier to make use of rvm which will allow you to install a version of ruby for just you user and switch between it and the system installed version. I really must update the development instructions to suggest this as a way of avoiding issues where the distro patches vagrant to work with newer versions of ruby than upstream is currently set up for

r0x0d:
@electrofelix I'm using the distro vagrant, just a basic `dnf install vagrant` and that's it! 

I'm not on my personal laptop right now, but I can send you the error message that I get from the distro vagrant running on F36.

r0x0d:
@electrofelix, now that everything is working as expected on my machine, I finally applied the suggestions from your review. 

Tested locally and it seems to be working fine. Now I'm taking a look into the integration test and trying to wrap my head around that.

locales/en.yml
r0x0d: Essentially, the problem on this one was because I was not passing the variable `error_message` when raising the `Errors::SnapshotMissing` class

tests/runtests.bats
```diff
@@ -215,3 +215,64 @@ cleanup() {
...
+  echo "status = ${status}"
+  # This means that the file does not exist on the box.
+  [ "$status" -eq 1 ]
+  run ${VAGRANT_CMD} snapshot delete default test
+  echo "${output}"
+  echo "status = ${status}"
+  [ "$status" -eq 0 ]
+
+  cleanup
+}
```
r0x0d, line 278: @electrofelix I tried to implement the integration tests you mentioned in https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1456#pullrequestreview-950098062, let me know if something is off in my understanding of it.

locales/en.yml
r0x0d: I'm resolving this conversation since it was me not paying enough attention to the code.

tests/runtests.bats
```diff
@@ -223,3 +223,64 @@ cleanup() {
 
   cleanup
 }
+
+@test "bring up and save a snapshot" {
+  export VAGRANT_CWD=tests/snapshot
```
electrofelix, line 228: I think maybe skip this test as you end up testing the same thing in the following one. Unless there was something specific for the snapshot saving that needed to be tested here or due to performance when testing locally that it was useful to test save by itself.

tests/runtests.bats
```diff
@@ -223,3 +223,64 @@ cleanup() {
...
+
+@test "bring up and save a snapshot and restore it" {
+  export VAGRANT_CWD=tests/snapshot
+
+  cleanup
+  run ${VAGRANT_CMD} up ${VAGRANT_OPT}
+  echo "${output}"
+  echo "status = ${status}"
+  [ "$status" -eq 0 ]
+  run ${VAGRANT_CMD} snapshot save default test
```
electrofelix, line 259: Might need to add another file of a different name before snapshot save here, then save, then delete and verify the file is gone, then after adding the file below that we're looking to be removed by restoring, also want to check that the file that previously existed and was deleted is restored.

Ideally this would ensure that the restore is actually restoring the snapshot that contained previous files and not just rolling back to the original boot state.

It might not work with the tinycore image as I've just discovered it doesn't persist changes if you reboot.

electrofelix (commented):
Looks good to me, if you get to adjust the integration test within the next day or two, otherwise I'm happy to tweak so we can merge since I've finally fixed the major issue in the 0.8 release that broken use of additional storage and released 0.9 to cover it, should merge this in the next couple of days along with the other outstanding PRs after it

tests/runtests.bats
r0x0d: Actually, there is no hidden purpose here. I was trying to test that those three operations were working properly. But I can remove it, since the full snapshot test I will do most of the operations here.

electrofelix:
@r0x0d I made a couple of tweaks just to remove the progress bar as it doesn't get executed and also the method `is_snapshot_mode_on?` as it was never called by anything.

r0x0d:
Thanks, @electrofelix! 

I hope to contribute more with vagrant-libvirt in the future :)

miguellavalle:
In what release can I expect to see the snapshots feature available? In Fedora 37 I get 0.7.0.5

electrofelix:
@miguellavalle you can follow https://vagrant-libvirt.github.io/vagrant-libvirt/installation.html#fedora to ensure that all dependencies for vagrant-libvirt come from fedora, remove the vagrant-libvirt package supplied by the distro and then just install the plug-in using vagrant directly and you'll get access to it now. I don't know why they stick with the 0.7 releases and are only using to 0.11 in rawhide now.

The only thing you need to remember is the default for the plug-in install is to use qemu:///system connection while fedora packaged defaults to qemu:///session. There is a setting to switch between the two, Just need to determine which behaviour you want and if it's not the default set the config flag for use_qemu_session as needed

PR #1457, opened 2022-02-28
improvements and fixes for container image

mmguero:
* Added `Dockerfile` to `.dockerignore` so that changes to it don't invalidate the Docker cache.
* Changed base image from `ubuntu:bionic` (last updated 2019) to `debian:stable-slim` as it's stable and slow-moving.
* Added some `-dev` packages to base layer to support building vagrant plugins.
* Allow additional vagrant plugins to be built and installed by specifying the `DEFAULT_OTHER_PLUGINS` build argument; default is `vagrant-mutate` as it aids in being able to convert non-libvirt images to libvirt format. Example: `--build-arg DEFAULT_OTHER_PLUGINS=vagrant-mutate,vagrant-reload,vagrant-scp,vagrant-sshfs`
* Ensure default `vagrant` user (UID 1000) is created in Docker image.
* Some minor fixes in the `entrypoint.sh` script: remove unused variable, fixed incorrect flag being passed to `usermod` command, and use `$VAGRANT_HOME` environment variable rather than hard-coding directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,31 @@
 # syntax = docker/dockerfile:1.0-experimental
 ARG VAGRANT_VERSION=2.2.19
 
+FROM debian:stable-slim as base
 
-FROM ubuntu:bionic as base
-
-RUN apt update \
-    && apt install -y --no-install-recommends \
-        bash \
-        ca-certificates \
-        curl \
-        git \
-        gosu \
-        kmod \
-        libvirt-bin \
-        openssh-client \
-        qemu-utils \
-        rsync \
-    && rm -rf /var/lib/apt/lists \
+RUN apt-get -y -qq update \
+    && apt-get -y --no-install-recommends install \
+      bash \
+      ca-certificates \
+      curl \
+      git \
+      gosu \
+      kmod \
+      libguestfs-tools \
+      libvirt0 \
+      libvirt-clients \
+      libvirt-dev \
+      libxml2-dev \
+      libxslt-dev \
+      openssh-client \
+      openssh-sftp-server \
+      qemu-system \
+      qemu-utils \
+      rsync \
+      ruby-dev \
+      zlib1g-dev \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
     ;
 
 RUN mkdir /vagrant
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,47 +1,58 @@
 # syntax = docker/dockerfile:1.0-experimental
 ARG VAGRANT_VERSION=2.2.19
 
-
-FROM ubuntu:bionic as base
```
electrofelix, line 5: The choice of `ubuntu:bionic` was to ensure that the client library wouldn't be trying to use newer features when talking to the service side when talking via the socket when run on likes of CentOS/RHEL 7 / Ubuntu 18.04. While it might be fine to switch debian flavour that uses a more recent libvirt (v7 vs v4) I think it might need more testing to check that it won't have issues when run against an older version of the libvirt server. That would need to be done before this PR could land, so perhaps better dealing with that as a separate PR where we can either have an additional set of images to have the newer libvirt if it doesn't work with the older distros or switch once confirmed that it will work.

Dockerfile
```diff
@@ -1,47 +1,58 @@
...
+      curl \
+      git \
+      gosu \
+      kmod \
+      libguestfs-tools \
+      libvirt0 \
+      libvirt-clients \
+      libvirt-dev \
+      libxml2-dev \
+      libxslt-dev \
```
electrofelix, line 19: The dev packages are installed as part of the build layer, including them as part of the base will bloat the slim image.

Dockerfile
```diff
@@ -1,47 +1,58 @@
...
+      openssh-client \
+      openssh-sftp-server \
+      qemu-system \
+      qemu-utils \
+      rsync \
+      ruby-dev \
+      zlib1g-dev \
```
electrofelix, line 26: Same here, the dev packages are installed as part of the build layer, including them as part of the base will bloat the slim image.

Dockerfile
```diff
@@ -1,47 +1,58 @@
...
-    && apt update \
-    && apt install -y ./vagrant.deb \
-    && rm -rf /var/lib/apt/lists/* \
-    && rm -f vagrant.deb \
+    && apt-get -y -qq update \
+    && curl -sSL -o /tmp/vagrant.deb "https://releases.hashicorp.com/vagrant/${VAGRANT_VERSION}/vagrant_${VAGRANT_VERSION}_x86_64.deb" \
+    && apt-get install -y /tmp/vagrant.deb \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
+    && useradd -M --uid ${DEFAULT_UID} --gid ${DEFAULT_GROUP} ${DEFAULT_USER} \
```
electrofelix, line 47: Can you elaborate on what problem you were solving here by having the user pre-created as it should be created by the entrypoint script instead?

Dockerfile
```diff
@@ -1,47 +1,58 @@
...
 
-
 ENV VAGRANT_DEFAULT_PROVIDER=libvirt
 
 FROM base as build
 
-# allow caching of packages for build
-RUN rm -f /etc/apt/apt.conf.d/docker-clean; echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache
-RUN sed -i '/deb-src/s/^# //' /etc/apt/sources.list
-RUN --mount=type=cache,target=/var/cache/apt --mount=type=cache,target=/var/lib/apt \
```
electrofelix, line 43: The mount caches are intended to help speed up rebuilds, prefer if they are kept in place.

Dockerfile
```diff
@@ -52,19 +63,27 @@ RUN --mount=type=cache,target=/var/cache/apt --mount=type=cache,target=/var/lib/
...
-        touch /vagrant/${dir}/.remove; \
-    done \
-    ;
+COPY . .
+RUN rake build \
+    && vagrant plugin install ./pkg/vagrant-libvirt*.gem \
+    && for plugin in $(echo "$DEFAULT_OTHER_PLUGINS" | sed "s/,/ /g"); \
+         do \
+           vagrant plugin install ${plugin} ; \
+         done \
```
electrofelix, line 81: This should probably be combined into a single call as otherwise I recall it repeats performing the install of all existing plugins along with each new plugin that is installed.

Dockerfile
```diff
@@ -52,19 +63,27 @@ RUN --mount=type=cache,target=/var/cache/apt --mount=type=cache,target=/var/lib/
...
+    && for dir in boxes data tmp; \
+         do \
+           touch /vagrant/${dir}/.remove; \
+         done \
+         ;
```
electrofelix, line 86: Not sure why this is moved as it doesn't benefit from being done after the plugin installs.

electrofelix (commented):
It's not clear if the impact of changing the base distro has been looked at, I currently don't have the time to test that so I'm reluctant to have it rolled in with some of the other changes.

mmguero:
Thanks for looking over the PR. I had some misunderstandings about the way things worked that you cleared up. I'll resubmit at another time or pull the updated base image debian one in for a different PR for testing. Thanks for the time spent explaining things.

Issue #1467, opened 2022-03-17
if ssh forwarding is enabled, ssh process is left running after destroy


abbbi:
VM has enabled ssh port forwarding:

`config.vm.network :forwarded_port, guest: 22, host: 2000, host_ip: "*", gateway_ports: true`

spin up the virtual machine, i can connect the ssh port from remote to host:

```
ssh vagrant@host -p 2000
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
```

now destroy the domain:

```
vagrant destroy -f
[2022-03-17 21:51:56] ==> default: Clearing any previously set forwarded ports...
[2022-03-17 21:51:56] ==> default: Removing domain...
[2022-03-17 21:51:57] ==> default: Deleting the machine folder
```

the ssh process is still running on the host system afterwards:

```
ps axw | grep ssh.*vagrant
3935651 pts/1    S      0:00 ssh -n -o User=vagrant -o Port=22 -o UserKnownHostsFile=/dev/null -o ExitOnForwardFailure=yes -o ControlMaster=no -o StrictHostKeyChecking=no -o PasswordAuthentication=no -o ForwardX11=no -o IdentitiesOnly=yes -o IdentityFile="/home/abi/git/integration-tests/vagrant/debian11/.vagrant/machines/default/libvirt/private_key" -L *:2000:192.168.28.101:22 -N 192.168.28.101 -g
```

now, if virtual machine is spinned up again, the ssh connection wont succeed because the leftover ssh process will accept the connection but forwarding fails due to other ip address in vm.

abbbi:
it seems the wrong pid is killed during destroy.
On a pristine system, spin up virtual machine, i can see following ssh processes:

```
ps axw | grep ssh.*vagrant*
[...]

PR #1468, opened 2022-03-17
(#1467) kill parent ssh process pid if ssh forwarding is enabled

abbbi:
see #1467 : untested attempt to fix situation.


abbbi:
that wont help, because pids to kill are found elsehwere

PR #1473, opened 2022-03-31
Added fix for issue #1472

### LINKED ISSUE #1472: Crash: driver.rb:138  undefined method
**Describe the bug**
Given `vagrant up --provider=libvirt` has been invoked and some (to me unknown) event causes the  vagrant domain to be lost
when I invoke a vagrant command like `vagrant status`
then I get the error message
```
root@a1dfd93e66e1:/# vagrant status
Traceback (most recent call last):
        14: from /opt/vagrant/embedded/gems/2.2.19/gems/vagrant-2.2.19/bin/vagrant:231:in `<main>'
         ...
         1: from /root/.vagrant.d/gems/2.7.4/gems/vagrant-libvirt-0.7.0/lib/vagrant-libvirt/provider.rb:98:in `state'
/root/.vagrant.d/gems/2.7.4/gems/vagrant-libvirt-0.7.0/lib/vagrant-libvirt/driver.rb:138:in `state': undefined method `to_sym' for nil:NilClass (NoMethodError)
```
**Expected behavior**
```
root@a1dfd93e66e1:/# vagrant status
Current machine states:
ubuntu-win10              not created (libvirt)
The Libvirt domain is not created. Run `vagrant up` to create it
```

**Versions:**
- Libvirt version: `vagrant-libvirt (0.7.0, global)`
- Vagrant version [output of `vagrant version`]: `Installed Version: 2.2.19`
  
- Vagrant flavour [Upstream or Distro]: `https://releases.hashicorp.com/vagrant/2.2.19/vagrant_2.2.19_x86_64.deb`
- Vagrant plugins versions (including vagrant-libvirt) [output of `vagrant plugin list`]:
   - `vagrant-libvirt (0.7.0, global)` 

**Additional info**
This issue is not about the root cause of why the domain is lost.
This issue only focuses on a missing check in `.../lib/vagrant-libvirt/driver.rb:138`
The line is currently 
   `return :not_created if domain.nil? || domain.state.to_sym == :terminated`
but I propose that is should be
   `return :not_created if domain.nil? || domain.state.nil? || domain.state.to_sym == :terminated`

I am currently working around the issue by patching my vagrant installation:
```
root@a1dfd93e66e1:/# F='/root/.vagrant.d//gems/2.7.4/gems/vagrant-libvirt-0.7.0/lib/vagrant-libvirt/driver.rb'
root@a1dfd93e66e1:/# sed -i '138s/^.*$/        return :not_created if domain.nil? || domain.state.nil? || domain.state.to_sym == :terminated/' "${F}"
```

ebaklund:
**Purpose**
Avoid crash caused by power management suspension where `fog` fails to translate domain state `pmsuspended`.

**Issues**
Closes #1472

**Changes**
- Changed the `driver.state()` function so that it does not crash on undefined domain state. Just return `nil` instead. The provider translates `nil` to `:unknown`.

### DIFF
--- a/lib/vagrant-libvirt/driver.rb
+++ b/lib/vagrant-libvirt/driver.rb
@@ -135,7 +135,9 @@ module VagrantPlugins
         end
 
         # TODO: terminated no longer appears to be a valid fog state, remove?
-        return :not_created if domain.nil? || domain.state.to_sym == :terminated
+        return :not_created if domain.nil?
+        return nil if domain.state.nil?
+        return :not_created if domain.state.to_sym == :terminated
 
         state = domain.state.tr('-', '_').to_sym
         if state == :running
--- a/lib/vagrant-libvirt/driver.rb
+++ b/lib/vagrant-libvirt/driver.rb
@@ -136,7 +136,7 @@ module VagrantPlugins
 
         # TODO: terminated no longer appears to be a valid fog state, remove?
         return :not_created if domain.nil?
-        return nil if domain.state.nil?
+        return :unknown if domain.state.nil?
         return :not_created if domain.state.to_sym == :terminated
 
         state = domain.state.tr('-', '_').to_sym
--- a/spec/unit/driver_spec.rb
+++ b/spec/unit/driver_spec.rb
@@ -35,10 +35,10 @@ describe VagrantPlugins::ProviderLibvirt::Driver do
   # name for the test machines above.
   let(:machine)    { iso_env.machine(:test1, :libvirt) }
   let(:machine2)   { iso_env.machine(:test2, :libvirt) }
-  let(:connection1) { double("connection 1") } 
-  let(:connection2) { double("connection 2") } 
-  let(:system_connection1) { double("system connection 1") } 
-  let(:system_connection2) { double("system connection 2") } 
+  let(:connection1) { double("connection 1") }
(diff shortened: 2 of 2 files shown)


lib/vagrant-libvirt/driver.rb
```diff
@@ -135,7 +135,9 @@ def state(machine)
         end
 
         # TODO: terminated no longer appears to be a valid fog state, remove?
-        return :not_created if domain.nil? || domain.state.to_sym == :terminated
+        return :not_created if domain.nil?
+        return nil if domain.state.nil?
```
electrofelix, line 139: While the provider code correctly handles nil, I do think it might be clearer to return `:unknown` here rather than rely on the subsequent conversion

electrofelix: Looks like the unit tests need some adjustments as the expected state being passed in no longer matches in a few cases resulting in slightly different behaviour. Possibly also the tests are a little fragile in a couple of cases

ebaklund: Thank you for feedback.
* I have translated the nil value to `:unknown` in the driver as requested.
* Added a unit test for nil handling.
* Made some unit tests less brittle with respect to domain.state call count.

ebaklund:
Thank you for your feedback.
Ready for new review.

Issue #1475, opened 2022-04-04
Order of interfaces with pci passthrough


hprem:
**Description**
With pci passthrough, the management interface comes after the passed-through interfaces. But I would like to make the management interface as the first interface (as I use cloud-init, which enables dhcp on first interface, by default)

**Issue**
On creating a VM with 2 passed-through NICs using the following vagrantfile
```ruby
Vagrant.configure("2") do |config|

  config.vm.define "ubuntu" do |ubuntu|

    ubuntu.vm.box = "/space/disk/ubuntu.21.10.box"
    ubuntu.vm.hostname = "host"
    ubuntu.vm.provider :libvirt do |libvirt|
        libvirt.machine_type = "q35"
        libvirt.pci :domain => '0x0000', :bus => "0x31", :slot => "0x00", :function => "0x0"
        libvirt.pci :domain => '0x0000', :bus => "0x31", :slot => "0x00", :function => "0x1"

    end
  end
end
```

the management interface gets slot number "06", which is higher than that of the passed through nics ("03" and "04")

```
gnos@host:~$ lspci | grep -i eth
03:00.0 Ethernet controller: Intel Corporation Ethernet Controller E810-C for QSFP (rev 02)
04:00.0 Ethernet controller: Intel Corporation Ethernet Controller E810-C for QSFP (rev 02)
06:00.0 Ethernet controller: Red Hat, Inc. Virtio network device (rev 01)   <===<<< Management intf
```

But I am using cloud-init, which tries to enable dhcp on the first interface; so I would like to make the management interface to be the first interface

So I tried adding the following extra config, to specify the slot number I want the management interface to appear at

```ruby
        libvirt.management_network_domain = "0x0000"
        libvirt.management_network_pci_bus = "0x03"
        libvirt.management_network_pci_slot = "0x00"
```

But  this resulted in the below mentioned error
```
Error while attaching new device to domain. Call to virDomainAttachDeviceFlags failed: XML error: Attempted double use of PCI Address 0000:03:00.0
```

**Expected behavior**
The management interface should get the specified domain/slot/port and the passed-through interfaces should appear after it.

electrofelix:
The error is coming from libvirt itself, so you will need to generate a debug log so I can see what XML was being sent to libvirt and I can then look at providing suggestions on how to adjust the parameters to get what you want.

hprem:
Here is the full vagrant log file
[...]

PR #1481, opened 2022-04-21
Guest PCI address options for passthrough devices

jamie01:
When assigning a pci passthrough device, the domain, bus, slot and
function parameters pertain to the host device address. Add guest_
flavours of these to allow setting the address of the device in the
guest as well.

Fixes: #1475

### DIFF
--- a/lib/vagrant-libvirt/templates/domain.xml.erb
+++ b/lib/vagrant-libvirt/templates/domain.xml.erb
@@ -230,15 +230,20 @@
       <backend model='random'>/dev/random</backend>
     </rng>
 <%- end -%>
-<%- @pcis.each do |pci| -%>
+<%-
+  @pcis.each do |pci|
+    pci = pci.reject { |k,v| v.nil? }
+    guest_prefix = "guest_"
+    pci_host = pci.select { |k,_| !k.to_s.start_with?(guest_prefix) }
+    pci_guest = pci.select { |k,_| k.to_s.start_with?(guest_prefix) }
+-%>
     <hostdev mode='subsystem' type='pci' managed='yes'>
       <source>
-	<address <%= pci.select { |k,_| [:domain, :bus, :slot, :function].include? k }
-	  .reject { |k,v| v.nil? }
-	  .map { |k,v| "#{k.to_s}='#{v}'" }
-	  .join(' ')
-	-%>/>
+	<address <%= pci_host.map { |k,v| "#{k.to_s}='#{v}'" } .join(' ') -%>/>
       </source>
+    <%- if !pci_guest.empty? -%>
+      <address type='pci' <%= pci_guest.map { |k,v| "#{k.to_s[guest_prefix.length..-1]}='#{v}'" } .join(' ') -%>/>
+    <%- end -%>
     </hostdev>
 <%- end -%>
 <%- @usbs.each do |usb| -%>
--- a/lib/vagrant-libvirt/templates/domain.xml.erb
+++ b/lib/vagrant-libvirt/templates/domain.xml.erb
@@ -239,7 +239,7 @@
 -%>
     <hostdev mode='subsystem' type='pci' managed='yes'>
       <source>
-	<address <%= pci_host.map { |k,v| "#{k.to_s}='#{v}'" } .join(' ') -%>/>
+        <address <%= pci_host.map { |k,v| "#{k.to_s}='#{v}'" } .join(' ') -%>/>
       </source>
     <%- if !pci_guest.empty? -%>
       <address type='pci' <%= pci_guest.map { |k,v| "#{k.to_s[guest_prefix.length..-1]}='#{v}'" } .join(' ') -%>/>
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -528,10 +528,14 @@ module VagrantPlugins
           pci_domain = options[:domain]
         end
 
-        @pcis.push(domain:    pci_domain,
-                   bus:       options[:bus],
-                   slot:      options[:slot],
-                   function:  options[:function])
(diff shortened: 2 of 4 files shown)


jamie01:
Hm, not sure if I broke something in the tests.. `package_simple` failed trying to connect to the machine but `package_complex_example` passed

electrofelix:
> Hm, not sure if I broke something in the tests.. `package_simple` failed trying to connect to the machine but `package_complex_example` passed

Yeah, as these are running for real using full emulation rather than a hypervisor, there is a slightly greater tendency for spurious failures

lib/vagrant-libvirt/templates/domain.xml.erb
```diff
@@ -239,6 +239,13 @@
 	  .join(' ')
 	-%>/>
       </source>
+    <%- if [pci[:guest_domain], pci[:guest_bus], pci[:guest_slot], pci[:guest_function]].any? -%>
```
electrofelix, line 242: I think it should be possible to adjust this block a bit to only need it to be aware of keys starting with the `guest_` suffix and allow control of what is acceptable via the config definition. At least until something more explicit is needed it's a little easier to only need to modify one place to add new attributes rather than keeping multiple in sync.

electrofelix (commented):
I'll try to put together an actual suggestion later once now reviewing by phone, but I think can adjust the block in the template slightly to only need the attributes to be defined in the config object.

lib/vagrant-libvirt/templates/domain.xml.erb
jamie01: I see what you mean, that would certainly be neater. My ruby knowledge still doesn't extend far beyond cut and paste examples so any suggestions are more than welcome :)

lib/vagrant-libvirt/templates/domain.xml.erb
```diff
@@ -239,6 +239,13 @@
...
+	<address type='pci' <%= pci.select { |k,_| [:guest_domain, :guest_bus, :guest_slot, :guest_function].include? k }
+	  .reject { |k,v| v.nil? }
+	  .map { |k,v| "#{k.to_s[6..-1]}='#{v}'" }
```
electrofelix, line 245: If at the top of the loop over the pcis you add the line:
```
pci = pci.reject { |k,v| v.nil? }
```
I believe it will remove the need to reject any empty values up front for the entire block. This is a slight tweak to the main pci block, and should allow removing both lines containing `.reject { |k,v| v.nil? }`.

Then by switching the code here to the following should allow skipping the need to explicitly identify the keys.
```suggestion
<%-   prefix = 'guest_'
      pci_guest = pci.select { |k,_| k.to_s.start_with?(prefix) }                                                                            
      if !pci_guest.empty?                                                                                                                   
-%>   
      <address type='pci' <%= pci_guest.map { |k,v| "#{k.to_s[prefix.length..-1]}='#{v}'" }.join(' ') -%>/>
```

If you want to do something similar above where you use `pci.select { |k,_| !k.to_s.start_with?(prefix) }` instead of the explicit keys `[:domain, :bus, :slot, :function]`, that would be great, but not necessary for this PR.

electrofelix (commented):
To see what happens clearly when best to update the all settings test by adding a line after:
https://github.com/vagrant-libvirt/vagrant-libvirt/blob/master/spec/unit/templates/domain_spec.rb#L94

should look something like:
```
domain.pci(domain: '0x0002', bus: '0x04', slot: '0x00', function: '0x0', guest_domain: '0x0000', guest_bus: '0x01', guest_slot: '0x01', guest_function: '0x0')
```

And subsequently when you run the tests using:
```
bundle exec rspec --color --format documentation --fail-fast spec/unit/templates/domain_spec.rb
```

Should see something like:
```
  1) templates/domain when all settings enabled renders template
     Failure/Error: expect(domain.to_xml('domain')).to eq xml_expected
     
       expected: "<domain type='kvm' xmlns:qemu='http://libvirt.org/schemas/domain/qemu/1.0'>\n  <name></name>\n  <tit...env name='QEMU_PA_SERVER' value='/run/user/1000/pulse/native'/>\n  </qemu:commandline>\n</domain>\n"
            got: "<domain type='kvm' xmlns:qemu='http://libvirt.org/schemas/domain/qemu/1.0'>\n  <name></name>\n  <tit...env name='QEMU_PA_SERVER' value='/run/user/1000/pulse/native'/>\n  </qemu:commandline>\n</domain>\n"
     
       (compared using ==)
     
       Diff:
       @@ -115,6 +115,11 @@
        	<address domain='0x0001' bus='0x03' slot='0x00' function='0x0'/>
              </source>
            </hostdev>
       +    <hostdev mode='subsystem' type='pci' managed='yes'>
       +      <source>
       +	<address domain='0x0002' bus='0x04' slot='0x00' function='0x0'/>
       +      </source>
       +      <address type='pci' domain='0x0000' bus='0x01' slot='0x01' function='0x0'/>
       +    </hostdev>
            <hostdev mode='subsystem' type='usb'>
              <source startupPolicy='mandatory'>
        	<vendor id='0x1234'/>
       
     # ./spec/unit/templates/domain_spec.rb:129:in `block (3 levels) in <top (required)>'

```

You can then use add the snippet (assuming it is correct, as I've only given it a cursory pass) to the file spec/unit/templates/domain_all_settings.xml and then provided the test passes, adjust the template as needed until the code is as desired.

lib/vagrant-libvirt/templates/domain.xml.erb
jamie01: Thank you for the detailed response, appreciate it! I've incorporated the change for both the host and guest attributes and it does make it a fair bit cleaner. Will push the changes now, and include the unit-test as well as you suggested, which is passing.

Issue #1445, opened 2022-02-04
Bringing up all boxes in multi-box configuration fails with Ruby 3.1


ryan-ronnander:
Vagrant fails to bring up one or more boxes in a multi box format with the libvirt provider. Running multiple `vagrant up` commands will eventually bring up all vagrant boxes. The following error occurs:

```text
An error occurred while executing multiple actions in parallel.
Any errors that occurred are shown below.

An unexpected error occurred when executing the action on the
'testbox-2' machine. Please report this as a bug:

constant VagrantPlugins::ProviderLibvirt::Action::HandleBox not defined

                b2.use HandleBox
                       ^^^^^^^^^

/home/rronnander/.vagrant.d/gems/3.1.0/gems/vagrant-libvirt-0.7.0/lib/vagrant-libvirt/action.rb:38:in `block (2 levels) in action_up'
/usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/builtin/call.rb:47:in `call'
/usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/warden.rb:48:in `call'
/usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/builtin/box_check_outdated.rb:93:in `call'
/usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/warden.rb:48:in `call'
/usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/builtin/config_validate.rb:25:in `call'
/usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/warden.rb:48:in `call'
/usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/builder.rb:149:in `call'
/usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/runner.rb:89:in `block in run'
/usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/util/busy.rb:19:in `busy'
/usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/action/runner.rb:89:in `run'
/usr/share/vagrant/gems/gems/vagrant-2.2.19/lib/vagrant/machine.rb:246:in `action_raw'
[...]

PR #1485, opened 2022-04-28
Use require for HandleBox, SyncedFolders, WaitForCommunicator

tserong:
Using `autoload` fairly consistently results in 'constant not defined' errors for each of HandleBox, SyncedFolders, WaitForCommunicator when using ruby 3.1.  Presumably there's some sort of race condition. Explicitly using `require` for these three makes the problem go away. It might be a bit ugly to have those `require` lines right at the point where those symbols are used, vs. the neat `autoload` section at the top of the file, but I'm not sure how better to structure this. Suggestions most welcome :-)

Fixes: #1445
Signed-off-by: Tim Serong <tserong@suse.com>

### DIFF
--- a/lib/vagrant-libvirt/action.rb
+++ b/lib/vagrant-libvirt/action.rb
@@ -48,15 +48,12 @@ module VagrantPlugins
       autoload :TimedProvision, action_root.join('timed_provision')
       autoload :WaitTillUp, action_root.join('wait_till_up')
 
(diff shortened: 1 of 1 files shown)


electrofelix:
while I'm not ecstatic about this approach, it seems prudent to land, do a bugfix release and then look at what can be done for the long term subsequently.

tserong:
Thanks!

PR #1486, opened 2022-04-28
Update steps in the addtional notes

r0x0d:
While trying to use the steps provided in the `Additional Notes for Fedora and
Similar Linux Distributions` I noticed that some of the versions referenced
there were not up-to-date, as well as a missing script call for the krb5.

Signed-off-by: Rodolfo Olivieri <rolivier@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,11 +353,11 @@ into_srpm.sh -d c8s
 cd SRPMS
 
 # common commands (make sure to adjust verison accordingly)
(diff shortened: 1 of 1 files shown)


electrofelix:
At some point need to completely rewrite the install instructions to tell people to default to using the distro vagrant where available and if they need to install the upstream one, https://github.com/vagrant-libvirt/vagrant-libvirt-qa/blob/main/scripts/install.bash should contain up to date steps.

PR #1489, opened 2022-05-05
Replace tabs with spaces in domain_all_settings.xml

dcermak:
The xml renderer stopped rendering tabs and now uses spaces instead, which made the test fail.


electrofelix (approved):

electrofelix:
@dcermak I spotted I forgot to release 0.8.3 when this was merged, would that still be useful or given that 0.9.0 is essentially just large bugfixes a better choice?

Issue #1477, opened 2022-04-13
vagrant up error talking to libvirt - unable to get index for interface eth0


cryptonaut01:
**Describe the bug**
I started with vagrant on a fresh ubuntu server 21.10 instance. Configured bridge br0 via yaml file. Now I am struggeling to distribute a first virt. box. How to tell vagrant to use that existing bridge ? vagrant up always creates a new network config on additional bridge virbr0 in nat mode, independent from what I configure within Vagrantfile config.

Error message:
==> linux: Starting domain.
There was an error talking to Libvirt. The error message is shown below:
Call to virDomainCreateWithFlags failed: Unable to get index for interface eth0: No such device

$ vagrant status
Current machine states:
linux                     shutoff (libvirt)
The Libvirt domain is not running. Run `vagrant up` to start it.

$ ip a
2: eno1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast master br0 state UP group default qlen 1000
    link/ether 68:1d:ef:25:a8:23 brd ff:ff:ff:ff:ff:ff
    altname enp1s0
    inet 192.168.2.156/24 brd 192.168.2.255 scope global eno1
       valid_lft forever preferred_lft forever
3: br0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 52:c7:b0:83:35:c2 brd ff:ff:ff:ff:ff:ff
    inet 192.168.2.156/24 brd 192.168.2.255 scope global br0
       valid_lft forever preferred_lft forever
    inet6 fe80::50c7:b0ff:fe83:35c2/64 scope link 
       valid_lft forever preferred_lft forever
11: virbr0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default qlen 1000
    link/ether 52:54:00:dd:44:6e brd ff:ff:ff:ff:ff:ff
    inet 192.168.121.1/24 brd 192.168.121.255 scope global virbr0
       valid_lft forever preferred_lft forever

$ virsh net-dumpxml vagrant-libvirt
```
<network ipv6='yes'>
  <name>vagrant-libvirt</name>
  <uuid>e0b582d2-1f90-411c-a784-34a90d9786dd</uuid>
  <forward mode='nat'>
    <nat>
      <port start='1024' end='65535'/>
    </nat>
[...]

PR #1499, opened 2022-05-21
Flag when source device for public network does not exist

electrofelix:
Detect when the device provided for connecting the VM to a public
network via the host does not exist and attempt to provide a useful
error message to give end users a hint of how to correct.

Fixes: #1477

### DIFF
--- a/lib/vagrant-libvirt/config.rb
+++ b/lib/vagrant-libvirt/config.rb
@@ -195,6 +195,9 @@ module VagrantPlugins
       # serial consoles
       attr_accessor :serials
 
+      # internal helper attributes
+      attr_accessor :host_device_exclude_prefixes
+
       def initialize
         @uri               = UNSET_VALUE
         @driver            = UNSET_VALUE
@@ -342,6 +345,9 @@ module VagrantPlugins
         @qemu_use_agent  = UNSET_VALUE
 
         @serials           = UNSET_VALUE
+
+        # internal options to help override behaviour
+        @host_device_exclude_prefixes = UNSET_VALUE
       end
 
       def boot(device)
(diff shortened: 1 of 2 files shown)


nexusVI:
My remote libvirt configuration doesn't work anymore: 
```
Libvirt Provider:
* network configuration 0 for machine default is a public_network referencing host device 'br0' which does not exist, consider adding ':dev => ....' referencing one of eth0
```

The device `br0` exists on the remote server but not on the server where i execute `vagrant`, this used to work before. 

Maybe related to this?

electrofelix:
*sigh* yes, it is, sorry I missed making sure this was correct for remote

electrofelix:
@nexusVI currently testing a fix which seems valid, but might take a little while before I can get it to push it up and land. In the mean time if you want to modify the code locally with the following, it should allow you to continue working:
```
@@ -1212,9 +1212,7 @@ module VagrantPlugins
 
       def host_devices
         @host_devices ||= begin
-          require 'socket'
-
-          Socket.getifaddrs.map { |iface| iface.name }.uniq.select do |dev|
+          machine.provider.driver.connection.client.list_all_interfaces().map { |iface| iface.name }.uniq.select do |dev|
             dev != "lo" && !@host_device_exclude_prefixes.any? { |exclude| dev.start_with?(exclude) }
           end
         end
```

Should also be possible to patch at the top of your Vagrantfile to contain the following to solve as well:
```
require 'vagrant-libvirt/config'

# since ruby doesn't seal classes, you can reopen and modify them
module VagrantPlugins
  module ProviderLibvirt
    class Config
      def host_devices
        @host_devices ||= begin
          machine.provider.driver.connection.client.list_all_interfaces().map { |iface| iface.name }.uniq.select do |dev|
            dev != "lo" && !@host_device_exclude_prefixes.any? { |exclude| dev.start_with?(exclude) }
          end
        end
      end
    end
  end
end

# place config after here
```

electrofelix:
@nexusVI https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1550 will be the fix

Issue #1508, opened 2022-06-07
No box and PXE boot example from documentation fails


mikfire:
**Describe the bug**
I am attempting to PXE boot a VM with no defined box, as described [in the docs](https://www.rubydoc.info/gems/vagrant-libvirt#no-box-and-pxe-boot).

`vagrant up pxeboot` fails with
```
resolve_disk_settings.rb:101:in `lookup_domain_by_uuid': no implicit conversion of nil into String (TypeError)
```
I believe this is a side effect of PR #1502. It looks like there is an assumption that if the config.vm.box isn't set, then the ID will be. In the PXE case, both the box and the ID are nil which is causing line 101 to throw the error. If I comment out lines 101 - 106, the machine comes up as expected. I am uncertain if that is a good idea or not.

**To Reproduce**
1. Install vagrant 2.2.19
2. say `vagrant plugin install vagrant-libvirt`
3. Verify vagrant-libvirt v 0.9.0 is installed
4. Say vagrant up pxeclient

**Expected behavior**
The expected behavior is that the client machine will start, recognize its network and PXE boot, as it did in v0.7.0.

**Versions (please complete the following information):**:
- Libvirt version: 0.9.0
- Vagrant version: `Installed Version: 2.2.19`
- Vagrant flavour: gentoo
- Vagrant plugins versions (including vagrant-libvirt):
```
pkg-config (1.4.7, global)
  - Version Constraint: > 0
vagrant-libvirt (0.9.0, global)
```

**Debug Log**
I would, but github keeps throwing errors when I try. The full error message is:

```
 INFO runner: Running action: environment_unload #<Vagrant::Action::Builder:0x00007f8c904e9db8>
/home/****/.vagrant.d/gems/2.7.6/gems/vagrant-libvirt-0.9.0/lib/vagrant-libvirt/action/resolve_disk_settings.rb:101:in `lookup_domain_by_uuid': no implicit conversion of nil into String (TypeError)
	from /home/****/.vagrant.d/gems/2.7.6/gems/vagrant-libvirt-0.9.0/lib/vagrant-libvirt/action/resolve_disk_settings.rb:101:in `call'
	from /usr/lib64/ruby/gems/2.7.0/gems/vagrant-2.2.19/lib/vagrant/action/warden.rb:48:in `call'
[...]

Issue #1371, opened 2021-10-05
Cannot undefine domain with nvram attached: Call to virDomainUndefineFlags failed:


abbbi:
hi,

attempting to get windows 11 going (sight), it needs to use UEFI. While that works nicely on debian, it
seems i need to use nvram settings on centos8* hosts, unfortunately, i cant undefine a virtual
machine if nvram setting is active:

```
`/home/buildtest/.vagrant.d/gems/2.6.6/gems/fog-libvirt-0.9.0/lib/fog/libvirt/requests/compute/vm_action.rb:7:in `undefine': Call to virDomainUndefineFlags failed: Requested operation is not valid: cannot undefine domain with nvram (Libvirt::Error
```

Example vagrant file settings:
```

        domain.loader = "/usr/share/OVMF/OVMF_CODE.secboot.fd"
        domain.nvram = true

```

Guess thats an issue in fog-libvirt, not setting VIR_DOMAIN_UNDEFINE_NVRAM flag while callling virDomainUndefine

electrofelix:
#1329 should solve this, just needs the related to land in fog-libvirt and then for the dependency long with the PR setting this flag to land here. Might be no harm pinging the fog-libvirt devs

electrofelix:
Actually looks like the PR for fog-libvirt fog/fog-libvirt#102 needs a minor update

Jeansen:
Is there any workaround this? I know this from destroying a VM and added a trigger that simply manipulates the XML and removes relevant entreis. But currently, I get this error when a domain is being created ....

Jeansen:
Looks like this problem came in with the latest 0.7.0 version. Up to 0.6.3 my Vagrant script works fine.

electrofelix:
@Jeansen can you share a debug log? There wasn't anything done between 0.6.3 and 0.7.0 that should have impacted nvram definitions

Jeansen:
@electrofelix  Hm..strange. I'll try around a bit with different version to eliminate any side effects. I'll post my results then here.

Jeansen:
Ah, I am sorry. I had a typo. It is version 0.5.3 that works for me. Here are some excerpts of my script.

From version 0.4.1 to 0.5.3 some more Error output appeared. But it still works. But with version 0.6 and above, I geht the stack trace you see.

Unfortunately, this project is still something I use for my personal playground and not yet on github and with some additional scripts involved, quite complex. 

[...]

PR #1509, opened 2022-06-08
Only populate domain_name if not provided

electrofelix:
Ensure that when no box is defined for a pxe install, that the domain
name is only populated for subsequent boots and if it is already
provided by a create action, skip attempting to retrieve from a
non-existing domain XML.

Fixes: #1508

### DIFF
--- a/lib/vagrant-libvirt/action/resolve_disk_settings.rb
+++ b/lib/vagrant-libvirt/action/resolve_disk_settings.rb
@@ -97,13 +97,15 @@ module VagrantPlugins
             # the default storage prefix is typically: /var/lib/libvirt/images/
             storage_prefix = "#{File.dirname(domain_volumes[0][:absolute_path])}/" # steal
           else
-            # Ensure domain name is set for subsequent steps if restarting a machine without a box
-            libvirt_domain = env[:machine].provider.driver.connection.client.lookup_domain_by_uuid(
-              env[:machine].id
-            )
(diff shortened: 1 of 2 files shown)


Jeansen:
I have something similar. My Vagrantfile works fine in previous versions. Actually, I could only run it for months in vers 4.1 because of #1371. Now, with 0.9.0 all errors there are gone, but I get the following:

Error while creating volume for domain: Call to virStorageVolCreateXMLiled: storage volume 'disk1_1-1.qcow2' exists already

In my Vagrantfile I have something like this:

 v.storage :file, path: disk, bus: 'scsi', allow_existing: true, size: size.to_i, serial: disk_name

So, there should be no error. And as I said, in version 4.1 - the last version I could test with - it works fine.

electrofelix:
@Jeansen that sounds like a different issue. The handling of the storage was reworked extensively over the last number of versions, so it is likely that I introduced an issue when the user specifies the path. Can you file a separate issue please.

Jeansen:
Master builds fine. Tag 6.0.0, too. I wasn't aware that I had to clean the modules folder. Thank you @electrofelix :smile:

Jeansen:
> @electrofelix OK. I'll try to create a minimal Vagrant file and create a new issue then.

@electrofelix  Just wanted to create a new issue but it seems with the latest versions 0.10.1, those errors are gone.

PR #1523, opened 2022-07-04
Migrate documentation to github pages

electrofelix:
Switch from having all documentation contained within the README to
publishing a jekyll static site of documentation under github pages.

This allows for configuration and installation guides to be handled
separately to make for the relevant pieces to be in smaller chunks and
therefore easier to follow for users. Additionally a table of contents
can now be included in a left navigation section that ensures it should
be possible to quickly jump from the start to any section and back
again.

Include support for publishing previews and releases under separate
directories to allow for them to exist at the same time as the other
latest version of the documents.

The navigation section also includes support for accessing any of the
other versions published so that it easier to see what configuration
options exist for a given release.


electrofelix:
@abbbi wonder if you would mind casting an eye over the general idea of this to migrate to GitHub pages for documentation to facilitate version specific docs as well as breaking up the giant README into more digestible chunks.

I have a couple of bugs to fix around how it switches between latest and specific versions, as well as the select element for switching not appearing on mobile devices by default.

Docs are at https://vagrant-libvirt.github.io/vagrant-libvirt

Also need to decide on whether it would be better to put the select element in an iframe so that all the releases can point to the same page for that in case there are mistakes that require being fixed in the future. As it stands what appears for each version is completely isolated from the others, it relies on the same code remaining working in the future.

abbbi:
i only had a quick look,but this looks promising and better to digest than the big readme.
Not so sure about the dark background/light font theme.

electrofelix:
I can make use of https://just-the-docs.github.io/just-the-docs/docs/customization/#switchable-custom-scheme to make the colour scheme switchable. I prefer dark as I find it easier to read, but no reason to enforce that on anyone else that prefers dark text on light.

PR #1533, opened 2022-07-29
Cache requests and style plugin version menu

electrofelix:
Add caching support to reduce the number of requests to lookup the
github APIs when generating the plugin version select menu.


electrofelix:
Looks like I broke the querying

PR #1536, opened 2022-08-08
Simple maintenance improvements

a1346054:


docs/_includes/upstream-vagrant-install.html
```diff
@@ -8,8 +8,8 @@
 * Alternatively install vagrant following [https://www.vagrantup.com/downloads](https://www.vagrantup.com/downloads):
 {% if include.distro == "debian" or include.distro == "ubuntu" -%}
 ```shell
-curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo apt-key add -
-sudo apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main"
+curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo tee /usr/share/keyrings/hashicorp-archive-keyring.asc
+echo "deb [ signed-by=/usr/share/keyrings/hashicorp-archive-keyring.asc ] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/hashicorp.list
```
electrofelix, line 12: Two quick questions on this, does this work for older releases, and is asc usually an ASCII armoured key rather than the keyring?

I haven't experimented to check either of these, just wondering if you might have the answers to hand.

a1346054: Oldest supported version I checked was ubuntu 18.04 and debian 10, likely works on older unsupported versions of debian/ubuntu too.

https://apt.releases.hashicorp.com/gpg is actually an ASCII armored key, not a keyring, hence it is suitable to use .asc as extension. Apt behaves differently when the extension is .gpg (which makes apt expect it to be a keybox) and it would not work. Everything works when the extension is .key or .asc

a1346054: Just checked in a container that debian 9 supports it too.

It also means that there's no dependency on the full gnupg package because there's no use of the `gpg` command any more.

PR #1565, opened 2022-08-29
Use compare-xml instead of equivalent-xml

dcermak:
`equivalent-xml` is pretty much dead upstream, `compare-xml` on the other hand has some upstream activity

### DIFF
--- a/lib/vagrant-libvirt/action/start_domain.rb
+++ b/lib/vagrant-libvirt/action/start_domain.rb
@@ -2,7 +2,7 @@
 
 require 'log4r'
 
-require 'equivalent-xml'
+require 'compare-xml'
 require 'rexml/document'
 
 module VagrantPlugins
@@ -437,7 +437,7 @@ module VagrantPlugins
                   proposed = Nokogiri::XML(new_xml, &:noblanks)
                   applied = Nokogiri::XML(applied_xml, &:noblanks)
 
-                  if !EquivalentXml.equivalent?(proposed, applied)
+                  if CompareXML.equivalent?(proposed, applied, { force_children: true })
                     require 'diffy'
 
                     # pretty print the XML as even though there can be additional changes,
--- a/vagrant-libvirt.gemspec
+++ b/vagrant-libvirt.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'fog-libvirt', '>= 0.6.0'
   s.add_runtime_dependency 'fog-core', '~> 2'
   s.add_runtime_dependency 'rexml'
-  s.add_runtime_dependency 'equivalent-xml'
+  s.add_runtime_dependency 'compare-xml'
   s.add_runtime_dependency 'diffy'
 
   # Make sure to allow use of the same version as Vagrant by being less specific


electrofelix:
Unfortunately I've been a bit premature in merging this change, it doesn't quite work as expected. It will no longer catch libvirt failing to apply XML fully.

I've put up https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1566 which has a test that is marked pending but fails with CompareXML, because it fails to correctly identify equivalent and different XML docs. It appears that unless the verbose mode is turned on and the check is against whether there is a list of differences returned or not makes me a little wary that the basic behaviour is wrong.

electrofelix:
Running the following via irb:
```
require 'compare-xml'
doc1 = Nokogiri::XML('''<a href="/admin" id="button_1" class="blue button">Link</a>''')
doc2 = Nokogiri::XML('''<a href="/admin" id="button_1" class="blue button">Link</a>''')
CompareXML.equivalent?(doc1, doc2)
```
displays false

Similarly:
```
require 'compare-xml'
doc1 = Nokogiri::XML('''<a href="/admin" id="button_1" class="blue button">Link</a>''')
doc2 = Nokogiri::XML('''<a href="/admin" id="button_1" class="blue yellow">Link</a>''')
CompareXML.equivalent?(doc1, doc2)
```
displays false.

If for the first set I instead run `CompareXML.equivalent?(doc1, doc2, {verbose: true})` I get an empty array returned, and for the second I get an array of differences.

This suggests the library isn't behaving as described when not in verbose mode.

electrofelix:
@dcermak experimenting a bit I find that it's possible to use xml-simple

It appears I can replace the following:
https://github.com/vagrant-libvirt/vagrant-libvirt/blob/64910dbd90a40735ce59ba9595fb68015cfb0be4/lib/vagrant-libvirt/action/start_domain.rb#L431-L450

With the following code and it will catch the added usecase
```ruby
                  require 'xmlsimple'
                  proposed = XmlSimple.xml_in(new_xml, {'NormaliseSpace' => 2})
                  applied = XmlSimple.xml_in(libvirt_domain.xml_desc(1), {'NormaliseSpace' => 2})

                  # content elements that are empty are preserved by xml-simple and will result
                  # in the structures being considered different even if functionality the same
                  # strip any empty ones to avoid.
                  def compact_content(node)
                    if node.is_a?(Array)
                      node.map! do |element|
                        compact_content(element)
                      end
                    elsif node.is_a?(Hash)
                      if node['content'] and node['content'].empty?
                        node.delete('content')
                      end
                      node.each do |k, v|
                        node[k] = compact_content(v)
                      end
                    else
                      return node
                    end
                  end

                  if compact_content(proposed) != compact_content(applied)
                    require 'diffy'

                    diff = Diffy::Diff.new(XmlSimple.xml_out(proposed), XmlSimple.xml_out(applied), :context => 3).to_s(:text)
```

I'd move the helper function to another file of course, but seems like ruby-xml-simple is likely well used already and this would remove pulling in a dependency that is not already available for a few distros.

Would this work for you?

dcermak:
> Would this work for you?

Yes of course and I'm very sorry for not testing this properly and pushing a broken PR causing you more work down the line.

electrofelix:
@dcermak I had plenty of opportunity to ensure there was a test case added for this scenario, given the whole purpose of this section of the code is to deal with the fun of libvirt not applying what it was asked to apply ;-)

PR #1566, opened 2022-08-29
Add test for libvirt update failure mode

electrofelix:
To ensure the XML comparison is validated add a test where libvirt
returns XML that is different to what was requested, and change that
both the error is raised, an error message emitted and that the domain
define would be reverted to the previous state if possible.


electrofelix:
Have a full PR ready to land, not just the test.

PR #1574, opened 2022-09-03
Ensure XML diff output if not fully applied

electrofelix:
Make sure to perform a diff of the actual XML strings and not the object
representations if encountering the bug where the XML failed to be
updated correctly.


electrofelix:
Wrong commit pushed up, will need to switch around

electrofelix:
Correct changes pushed as different PRs

PR #1602, opened 2022-09-21
[Dockerfile] Fix creation of plugins.json with buildah

dcermak:
Multi-line RUN instructions are unsupported by default docker (works 
with buildkit) and buildah, instead one has to use other tricks like 
ANSI-C quoting for the shell via \n\ and multiple echo commands.

Includes a buildah github workflow to help ensure future
compatibility.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,20 +61,18 @@ RUN find /opt/vagrant/embedded/ -type f | grep -v /opt/vagrant/embedded/plugins.
 
 RUN /opt/vagrant/embedded/bin/gem install --install-dir /opt/vagrant/embedded/gems/${VAGRANT_VERSION} ./pkg/vagrant-libvirt*.gem
 
-RUN cat <<EOF > /opt/vagrant/embedded/plugins.json
-{
-    "version": "1",
-    "installed": {
-        "vagrant-libvirt": {
-            "ruby_version":"$(/opt/vagrant/embedded/bin/ruby -e 'puts RUBY_VERSION')",
-            "vagrant_version":"${VAGRANT_VERSION}",
-            "gem_version":"",
-            "require":"",
-            "sources":[]
-        }
-    }
-}
-EOF
+RUN echo -n '{\n\
+    "version": "1",\n\
+    "installed": {\n\
+        "vagrant-libvirt": {\n\
+            "ruby_version": "'$(/opt/vagrant/embedded/bin/ruby -e 'puts "#{RUBY_VERSION}"')'",\n\
+            "vagrant_version": "'${VAGRANT_VERSION}'",\n\
+            "gem_version":"",\n\
+            "require":"",\n\
+            "sources":[]\n\
+        }\n\
+    }\n\
+}' > /opt/vagrant/embedded/plugins.json
 
 FROM build as pruned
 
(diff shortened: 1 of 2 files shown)


electrofelix:
Multiline notation is working with docker buildkit, which is what is being used to build the resulting image.

Is there a way to validate that the dockerfile is compatible with buildah without needing to also support doing a build with it, e.g. buildah lint, otherwise there is a hidden constraint on being compatible with something that isn't checked through CI.

dcermak:
> Multiline notation is working with docker buildkit, which is what is being used to build the resulting image.

Ah, that explains why this was not working with either docker nor buildah.

> Is there a way to validate that the dockerfile is compatible with buildah without needing to also support doing a build with it, e.g. buildah lint, otherwise there is a hidden constraint on being compatible with something that isn't checked through CI.

Afaik buildah has no builtin linter (yet), so I looked at different linters:
- https://github.com/hadolint/hadolint will immediately complain about the non-standard mounts in line 41 (I removed these but did not submit this change, as I thought docker would support that)
- https://github.com/RedCoolBeans/dockerlint finds the multiline instruction as an issue and also yells about other things:
```
❯ podman run -it --rm -v "$PWD/Dockerfile":/Dockerfile:Z redcoolbeans/dockerlint
ERROR: Multiple ENTRYPOINT instructions found, only line 96 will take effect
ERROR: { is invalid on line 65

ERROR: /Dockerfile failed.
```

electrofelix:
Looks like there is a buildah github action, so might be able to add that subsequently to ensure this remains compatible with it in the future.

Dockerfile
```diff
@@ -61,20 +61,20 @@ RUN find /opt/vagrant/embedded/ -type f | grep -v /opt/vagrant/embedded/plugins.
...
-            "sources":[]
-        }
-    }
-}
-EOF
+RUN echo -n '{\n\
+    "version": "1",\n\
+    "installed": {\n\
+        "vagrant-libvirt": {\n\
+            "ruby_version":"' > /opt/vagrant/embedded/plugins.json \
```
electrofelix, line 68: Instead of outputting to a file here, does the following work?
```
        "ruby_version": "'$(/opt/vagrant/embedded/bin/ruby -e 'puts "#{RUBY_VERSION}"')'",\n\
        "vagrant_version": "'${VAGRANT_VERSION}'",\n\
            "gem_version":"",\n\
            "require":"",\n\
            "sources":[]\n\
        }\n\
    }\n\
}' >> /opt/vagrant/embedded/plugins.json
```

Basically making use of `'` to switch back to allowing shell expansion and then re-opening with `'` for the echo to continue.

electrofelix (commented):
I think when switching to be buildah compatible need to switch the syntax reference at the top of the dockerfile back to `# syntax = docker/dockerfile:1.0-experimental`

Dockerfile
dcermak: unfortunately shell expansion does not work in between `'` braces

dcermak:
> Basically means I'll look to get a buildah action going as well to ensure it remains working after this PR lands.

I have created one, could you please approve the CI runs to see if it works?

Dockerfile
electrofelix: The change above seems to work when I tried it with buildah locally, the `'` are being closed temporarily to allow the next piece use shell expansion and then re-opened again.

electrofelix:
@dcermak I've adjusted the approval settings on actions, so hopefully it should just run for you each time now.

Dockerfile
dcermak: Ah, indeed! Sorry, I did not read your comment correctly. I've used your changes and pushed them.

PR #1603, opened 2022-09-21
Spec fixes

dcermak:
- the sysprep.sh script was not executable (flagged by rpmlint)
- Need to explicitly `require 'tmpdir'` in `spec_helper.rb` on openSUSE Tumbleweed,
  otherwise you get tons of failures like:
```
An error occurred while loading ./spec/unit/action/prepare_nfs_settings_spec.rb.
Failure/Error: temp_dir = Dir.mktmpdir("rspec-")

NoMethodError:
  undefined method `mktmpdir' for Dir:Class

    temp_dir = Dir.mktmpdir("rspec-")
                  ^^^^^^^^^
  Did you mean?  mkdir
# ./spec/spec_helper.rb:44:in `block in <top (required)>'
# ./spec/spec_helper.rb:41:in `<top (required)>'
# ./spec/unit/action/prepare_nfs_settings_spec.rb:3:in `require'
# ./spec/unit/action/prepare_nfs_settings_spec.rb:3:in `<top (required)>'
```


dcermak:
Darragh Bailey ***@***.***> writes:

> For the rspec tmodir is tumbleweed using a newer ruby? I'm scratching my head a little as to why this hasn't bitten me yet, perhaps I'm lagging behind too much.

Yes, Tumbleweed is using Ruby 3.1 (and has been for months now). My
guess is that it's a combination of a newer rspec and maybe a newer
Ruby? But this issue is unfortunately not that easy to reproduce, as you
have keep patching vagrant (it did not support Ruby 3.1 until a few
hours ago).

> Also any idea why the script started to cause issues as it wasn't executable before being moved. Is it the acceptance tests weren't being run before during the package build on tumbleweed, or is it that a newer libguestfs tools is executing the script directly instead of by passing it to a shell?

The script did not cause any issues at all. This only appeared because
rpmlint complained about a non-executable script. It's really just a
cosmetic change and not necessary at all.

> Both changes look fine btw, I'm just curious if there are gaps in what is being tested that I need to look at covering to prevent other issues slipping through

It would be nice if we could keep testing vagrant-libvirt on all mayor
Linux distributions, or at least on those that are packaging
it. Unfortunately, that is not really a simple endeavor, as at least
both openSUSE and Fedora carry a few downstream patches for vagrant
(e.g. to make it compatible with Ruby 3.1). E.g. for openSUSE
Tumbleweed, you'd have to keep patching Vagrant since January to make
the tests run and the whole process is probably going to become
necessary again, when Ruby 3.2 is going to be released.

electrofelix:
> Yes, Tumbleweed is using Ruby 3.1 (and has been for months now). My guess is that it's a combination of a newer rspec and maybe a newer Ruby? But this issue is unfortunately not that easy to reproduce, as you have keep patching vagrant (it did not support Ruby 3.1 until a few hours ago).

We patch for ruby 3.0, so possibly can just apply the same fix there.

> It would be nice if we could keep testing vagrant-libvirt on all mayor Linux distributions, or at least on those that are packaging it. Unfortunately, that is not really a simple endeavor, as at least both openSUSE and Fedora carry a few downstream patches for vagrant (e.g. to make it compatible with Ruby 3.1). E.g. for openSUSE Tumbleweed, you'd have to keep patching Vagrant since January to make the tests run and the whole process is probably going to become necessary again, when Ruby 3.2 is going to be released.

Presumably could look at using docker images for the distros to execute the rspec tests with vagrant installed from the distro packages as a way of solving. I'll look at taking some of the work I did for the vagrant-libvirt-qa repo, as I've started recommending that users use the distro vagrant rather than upstream to avoid various linking issues between ruby <-> ruby-libvirt <-> libvirt.
Actions support using a container image as the base env https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container so that's probably a path to help catch these issues here before releases. I'll see what I can do

electrofelix:
Did a quick test and the failures don't appear on ruby 3.1.2 when just running `bundle exec rspec`, so it's probably something specific to the combination of packages in the distro. It's still a valid change though as it makes sense to require it before attempting to use, just puzzling as to how it is currently working without triggering an error all the time.

dcermak:
> Presumably could look at using docker images for the distros to execute the rspec tests with vagrant installed from the distro packages as a way of solving. I'll look at taking some of the work I did for the vagrant-libvirt-qa repo, as I've started recommending that users use the distro vagrant rather than upstream to avoid various linking issues between ruby <-> ruby-libvirt <-> libvirt.
> Actions support using a container image as the base env https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container so that's probably a path to help catch these issues here before releases. I'll see what I can do

Can I help you in any way?

PR #1613, opened 2022-09-28
Restore use of cache mounts

electrofelix:
Adjust syntax used for cache mounts to align with a format supported
by both buildah and docker.


electrofelix:
This will need buildah 1.26 available in the runner.

electrofelix:
@dcermak it looks like the version of buildah on the Ubuntu runners work with the mount cache. When I tested this before with https://github.com/vagrant-libvirt/vagrant-libvirt/actions/runs/3147154713/jobs/5116377272 I got the impression that the buildah version wasn't quite new enough and thought 1.26 would be required.

However looking through the history of https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md I'm not seeing any indication that buildah was changed in the mean time, nor when I put up a test workflow did it appear that the version had moved on from 1.23.1.

Since it's a bit unclear why it started to work before a new minor release, wondering if you could check this PR change locally to make sure I'm not re-introducing something that will break for other distros where buildah is in use? I've checked it locally and I'm still seeing the same error as before and I'm running 22.04 myself, so I'm a bit wary now of what buildah is actually in the runners.

electrofelix:
@dcermak never mind, I just spotted my mistake, forgot to update buildah to use the correct ref given I've moved the workflow to use `pull_request_target`.

dcermak:
> @dcermak never mind, I just spotted my mistake, forgot to update buildah to use the correct ref given I've moved the workflow to use `pull_request_target`.

Ok, I got confused :sweat_smile: because it builds fine with buildah 1.27 on my machine.

PR #1672, opened 2022-11-12
feat: add websocket graphics config

Tooa:
Hi :wave:, 

this commit adds websocket functionality for VNC. The websocket attribute may be used to specify the port to listen on (with -1 meaning auto-allocation and autoport having no effect due to security reasons).

Let me know what you think or if you would like to have anything changed.

### DIFF
--- a/lib/vagrant-libvirt/action/create_domain.rb
+++ b/lib/vagrant-libvirt/action/create_domain.rb
@@ -271,7 +271,7 @@ module VagrantPlugins
           end
 
           env[:ui].info(" -- Graphics Type:     #{@graphics_type}")
-          env[:ui].info(" -- Graphics Websocket: #{@graphics_websocket}")
+          env[:ui].info(" -- Graphics Websocket: #{@graphics_websocket}") if @graphics_websocket != -1
           if !@graphics_autoport
             env[:ui].info(" -- Graphics Port:     #{@graphics_port}")
             env[:ui].info(" -- Graphics IP:       #{@graphics_ip}")
--- a/lib/vagrant-libvirt/action/create_domain.rb
+++ b/lib/vagrant-libvirt/action/create_domain.rb
@@ -65,6 +65,7 @@ module VagrantPlugins
           @graphics_type = config.graphics_type
           @graphics_autoport = config.graphics_autoport
           @graphics_port = config.graphics_port
+          @graphics_websocket = config.graphics_websocket
           @graphics_ip = config.graphics_ip
           @graphics_passwd = config.graphics_passwd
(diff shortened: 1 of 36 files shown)


lib/vagrant-libvirt/action/create_domain.rb
```diff
@@ -270,6 +271,7 @@ def call(env)
           end
 
           env[:ui].info(" -- Graphics Type:     #{@graphics_type}")
+          env[:ui].info(" -- Graphics Websocket: #{@graphics_websocket}")
```
electrofelix, line 274: Probably should only be displayed if not set to `-1`, otherwise should look to update https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1672/files#diff-5590e6b922ec8d4bccf50817db5e075b6445e9fcdd7ad70a1a3b2b1f6702c1b9R551-R559 to output the port assigned once the VM is started.

electrofelix (commented):
main tests that will be useful would be in https://github.com/vagrant-libvirt/vagrant-libvirt/blob/main/spec/unit/action/start_domain_spec.rb and https://github.com/vagrant-libvirt/vagrant-libvirt/blob/main/spec/unit/action/create_domain_spec.rb, should be some examples of how to do some partial matching of the rendered XML to avoid needing to match the entire doc.

lib/vagrant-libvirt/action/create_domain.rb
Tooa: The commit 78b4cf7864bf36bc8c514a86a058d85b077a44d2 should address your review comment. Let me know if I got you wrong. 

Still figuring out how I can make the tests pass without adding `websocket = '-1'` to all test fixtures. Can you give an example for the test `should execute correctly`?

electrofelix: Likely need to update the fixtures as this is changing the default output

Tooa: Done, thanks for your input. Mind taking a look again?

Tooa:
Are there any open issues or things you want me to address @electrofelix ? Just want to make sure there is no obvious blocker I haven't noticed yet. Take your time otherwise.

electrofelix:
@Tooa nothing needed, just needed a few minutes to tidy up the message being used for squash, sorry for the delay.

Tooa:
> @Tooa nothing needed, just needed a few minutes to tidy up the message being used for squash, sorry for the delay.

No worries about the delay. Thanks for assisting and pointing me in the right direction to get this PR merged.

PR #1676, opened 2022-11-19
Add libvirt-dev to Ubuntu install

daviddavis:
Installation of vagrant-libvirt was failing for me until I installed the libvirt-dev package. 

[Here's](https://packages.ubuntu.com/jammy/all/libvirt-dev/filelist) a list of files from the libvirt-dev package. I believe that vagrant-libvirt needs some of these files such as libvirt.pc.


electrofelix:
Thanks, lgtm

PR #1677, opened 2022-11-21
Use append to a truncated file to retain owner

electrofelix:
Change the default behaviour when logging output from the console to a
file to ensure the file is pre-created and unless append is explicitly
set to true in the source config, truncate the file. Combined with
having the domain template default to enabling append, this causes
libvirt/qemu to not attempt to create the file and thus retains the
owner/group and permissions of the original file.

Relates-to: #1385

### DIFF
--- a/lib/vagrant-libvirt/action/create_domain.rb
+++ b/lib/vagrant-libvirt/action/create_domain.rb
@@ -198,6 +198,8 @@ module VagrantPlugins
             dir = File.dirname(serial[:source][:path])
             begin
               FileUtils.mkdir_p(dir)
+              FileUtils.touch(serial[:source][:path])
+              File.truncate(serial[:source][:path], 0) unless serial[:source].fetch(:append, false)
             rescue ::Errno::EACCES
               raise Errors::SerialCannotCreatePathError,
                     path: dir
--- a/lib/vagrant-libvirt/templates/domain.xml.erb
+++ b/lib/vagrant-libvirt/templates/domain.xml.erb
@@ -208,9 +208,12 @@
 <%- end -%>
 <%- @serials.each_with_index do |serial, port| -%>
     <serial type='<%= serial[:type] %>'>
-      <%- unless serial[:source].nil? -%>
-      <source path='<%= serial[:source][:path] %>'/>
-      <%- end -%>
+  <%- unless serial[:source].nil?
+    source = serial[:source]
+    seclabel = serial[:source][:seclabel]
+-%>
+      <source path='<%= source[:path] %>' append='<%= source.fetch(:append, 'on') %>'/>
+  <%- end -%>
       <target port='<%= port %>'/>
     </serial>
 <%- end -%>
--- a/spec/unit/templates/domain_all_settings.xml
+++ b/spec/unit/templates/domain_all_settings.xml
@@ -105,7 +105,7 @@
       <target dev='fda' bus='fdc'/>
     </disk>
     <serial type='file'>
-      <source path='/var/log/vm_consoles/machine.log'/>
+      <source path='/var/log/vm_consoles/machine.log' append='on'/>
       <target port='0'/>
     </serial>
     <console type='file'>


electrofelix:
@mazzy89 with the reload fix from the issue you mentioned, I'd hope this might be sufficient to solve the file permissions issue for the serial console. I don't have a box set up that sends the console output to the serial to completely validate the expected behaviour. Can try to get to that at some stage, unless you can point me in the direction of one known to have it enabled. Or alternatively if you can confirm whether this improves or hinders, that would also help.

mazzy89:
@electrofelix thanks a lot for providing such fix promptly. I'm going to make some adjustment and drop Libvirt as a binary in favour of running it as a docker container so then I can swap that container for the one backed here.

mazzy89:
@electrofelix but I have a question. this container comes with the required dependencies? I mean in my case I run vagrant-libvirt in fedora. so it's not clear which dependencies are required to be installed in the host machine and which are already baked in the container.

electrofelix:
@mazzy89 to use the image as described at https://vagrant-libvirt.github.io/vagrant-libvirt/installation.html#docker--podman, host machine just needs to have docker and libvirt installed, and the user needs to be in the libvirt group if using the system libvirt context.

If you were using vagrant-libvirt from the Fedora distribution previously it would have enabled the `qemu_use_session` config option by default, so you may need to explicitly enable that to ensure when connecting it uses `qemu:///session` instead of `qemu:///system`, which is the upstream default, and would be used by the docker image based installation.

mazzy89:
@electrofelix I'm ready to test the image but I cannot download that image because the registry requires auth.

electrofelix:
@mazzy89 I think I've changed the right settings to make it public now

mazzy89:
@electrofelix the container misses dependency

```
The executable 'ip' Vagrant is trying to run was not
found in the PATH variable. This is an error. Please verify
this software is installed and on the path.
```

electrofelix:
@mazzy89 do you need to sync the current folder into the VMs? I'm guessing it's trying to use nfs by default instead of rsync?

Given that nfs is unlikely to work nicely (need to call exportfs, access /etc/exports on the host, and potentially start the nfs-server), if you don't need it you can disable it using the following config.
```
config.vm.synced_folder ".", "/vagrant", disabled: true
```

If you definitely need it, can try to see whether it is possible after #1681 is merged to include the ip tool, but might be worth initially disabling and then we can look at other mount types such as virtiofs or 9p as alternatives if nfs is a bit tricky from inside a container.

The other alternative will be just to use the gem file attached to the PR and install using `vagrant plugin install ./vagrant-libvirt-0.10.9.pre.63.gem` as there are instructions on how to set up to use Fedora's distro vagrant package with the latest plugin from rubygems https://vagrant-libvirt.github.io/vagrant-libvirt/#installation

electrofelix:
@mazzy89 I did an experiment earlier today and it's unlikely that it will be possible for NFS to work without moving where /etc/exports is located on the host because docker won't correctly overwrite it the way vagrant creates a new file and then looks to move it to replace the existing. The alternative is to patch the vagrant behaviour so that for hosts that support /exports.d/ it will write a file per machine to that directory instead of trying to update /etc/exports directly.

But unlikely to be straightforward to resolve to use NFS for synced folders when testing this via the docker image

mazzy89:
In my case I do not need any NFS and in my Vagrantfile three is no definition of synced folder. Likely it's the default behavior. Later today I'll test with your recommendation to explicitly disable NFS

LKHN:
@electrofelix 

Hi, I still having the permission issue:

```ruby
/home/vagrant/.vagrant.d/gems/2.7.6/gems/fog-libvirt-0.11.0/lib/fog/libvirt/requests/compute/vm_action.rb:7:in `create': Call to virDomainCreateWithFlags failed: Unable to open file: test.log: Permission denied (Libvirt::Error)
```

An empty `test.log` file is created with this stats:

```sh
[vagrant@almalinux8-kvm almalinux-8]$ stat test.log
  File: test.log
  Size: 0               Blocks: 0          IO Block: 4096   regular empty file
Device: 802h/2050d      Inode: 18038182    Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1000/ vagrant)   Gid: ( 1000/ vagrant)
Context: unconfined_u:object_r:user_home_t:s0
Access: 2023-01-23 10:49:17.649215162 +0000
Modify: 2023-01-23 10:48:57.319084178 +0000
Change: 2023-01-23 10:48:57.319084178 +0000
 Birth: 2023-01-23 10:48:57.319084178 +0000

```

No luck with disabling `synced_folder` option too.

Full output:

[vagrant_libvirt_serial.log](https://github.com/vagrant-libvirt/vagrant-libvirt/files/10478956/vagrant_libvirt_serial.log)

Vagrantfile:

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

    config.vm.box = "almalinux/8"
    config.vm.hostname = "almalinux8-kvm.test"
    config.vm.synced_folder ".", "/vagrant",
    type: "virtiofs"

    config.vm.provider "libvirt" do |libvirt|
        libvirt.qemu_use_session = false
        libvirt.memory = 2048
        libvirt.memorybacking :access, :mode => "shared"
        libvirt.machine_type = "q35"
        libvirt.nested = true
        libvirt.cpu_mode = "host-passthrough"
        libvirt.cpus = 2
        libvirt.disk_bus = "scsi"
        libvirt.serial :type => "file", :source => {:path => "test.log"}
    end
end

```
Host: AlmaLinux OS 8.7

Vagrant: Upstream

```sh
vagrant version
Installed Version: 2.3.4
Latest Version: 2.3.4
 
You're running an up-to-date version of Vagrant!

```
Vagrant-libvirt: Upstream

```sh
vagrant plugin list
vagrant-libvirt (0.11.2, global)
```

Thanks!

electrofelix:
@LKHN you would need to run the latest development version, see installation instructions at https://github.com/vagrant-libvirt/vagrant-libvirt#latest-development-version

Looks like you are running an official release which wouldn't include this change yet.

If the following block doesn't disable the default synced folder it would mean a big with vagrant itself:
```
config.vm.synced_folder ".", "/vagrant", disabled: true
```

LKHN:
I replaced these files with the release ones:

`lib/vagrant-libvirt/action/create_domain.rb`
`lib/vagrant-libvirt/templates/domain.xml.erb`

As I see from the domain definition, it works. Isn't it? If it is not enough I will install the development version and let you know soon.

```xml
    <serial type='file'>
      <source path='test.log' append='on'/>
      <target port='0'/>
    </serial>
```

I disabled the `synced_folder` on the different run still same error:

`config.vm.synced_folder ".", "/vagrant", disabled: true`

electrofelix:
@LKHN thanks, that wasn't obvious from the initial comment. Looking at the xml, the problem is use of relative paths. I must have ensured I was resolving the path via the Vagrantfile upfront instead of doing it in the code. Try using an absolute path and that should take care of the initial problem.

Will have to revisit with a change to resolve relative to absolute paths.

LKHN:
@electrofelix 
Thanks for the quick replies. No luck with the absolute path too.

**Vagrantfile:**

```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

    config.vm.box = "almalinux/8"
    config.vm.hostname = "almalinux8-serial.test"
    config.vm.synced_folder ".", "/vagrant", disabled: true

    config.vm.provider "libvirt" do |libvirt|
        libvirt.qemu_use_session = false
        libvirt.memory = 2048
        libvirt.machine_type = "q35"
        libvirt.nested = true
        libvirt.cpu_mode = "host-passthrough"
        libvirt.cpus = 2
        libvirt.disk_bus = "scsi"
        libvirt.serial :type => "file", :source => {:path => "/home/vagrant/almalinux-8/test.log"}
        libvirt.management_network_address = "192.168.125.0/24"
    end
end
```

**Result:**

[vagrant_libvirt_serial_absolute_path.log](https://github.com/vagrant-libvirt/vagrant-libvirt/files/10479504/vagrant_libvirt_serial_absolute_path.log)

**Domain:**

```xml
<domain type='kvm' xmlns:qemu='http://libvirt.org/schemas/domain/qemu/1.0'>
  <name>almalinux-8_default</name>
  <title></title>
  <description>Source: /home/vagrant/almalinux-8/Vagrantfile</description>
  <uuid></uuid>
  <memory>2097152</memory>
  <vcpu>2</vcpu>
  <cpu mode='host-passthrough'>
    <feature policy='optional' name='vmx'/>
    <feature policy='optional' name='svm'/>
  </cpu>
  <os>
    <type machine='q35'>hvm</type>
    <bootmenu enable='no'/>
    <kernel></kernel>
    <initrd></initrd>
    <cmdline></cmdline>
  </os>
  <features>
    <acpi/>
    <apic/>
    <pae/>
  </features>
  <clock offset='utc'>
  </clock>
  <devices>
      <alias name='ua-box-volume-0'/>
      <driver name='qemu' type='qcow2' cache='default'/>
      <source file='/var/lib/libvirt/images/almalinux-8_default.img'/>
      <target dev='sda' bus='scsi'/>
    </disk>
    <controller type='scsi' model='virtio-scsi' index='0'/>
    <serial type='file'>
      <source path='/home/vagrant/almalinux-8/test.log' append='on'/>
      <target port='0'/>
    </serial>
    <console type='file'>
      <source path='/home/vagrant/almalinux-8/test.log'/>
      <target port='0'/>
    </console>
    <input type='mouse' bus='ps2'>
    </input>
    <graphics type='vnc' port='-1' autoport='yes' websocket='-1' listen='127.0.0.1' keymap='en-us'/>
    <video>
      <model type='cirrus' vram='16384' heads='1'/>
    </video>
  </devices>
</domain>
```

electrofelix:
Are you able to watch the file from before hand and capture what the permissions are and what they get changed to? It's possible that there needs to be an selinux block added to the domain def for this file to allow qemu but it's worth seeing what the file is created with first.

PR #1682, opened 2022-11-30
extra disk :device is optional, causes nil to String error

BlackIkeEagle:
When you add an additional disk with all default settings you get an error when issuing `vagrant reload`

example extra disk:

```
libvirt.storage :file, :size => '20G'
```

on `vagrant reload` you get:

```
/home/ike/.vagrant.d/gems/3.0.4/gems/vagrant-libvirt-0.10.8/lib/vagrant-libvirt/action/start_domain.rb:52:in `+': no implicit conversion of nil into String (TypeError)
        from /home/ike/.vagrant.d/gems/3.0.4/gems/vagrant-libvirt-0.10.8/lib/vagrant-libvirt/action/start_domain.rb:52:in `block in call'
        from /home/ike/.vagrant.d/gems/3.0.4/gems/vagrant-libvirt-0.10.8/lib/vagrant-libvirt/action/start_domain.rb:49:in `each'
```

So this now first checks if the :device is set before trying to do the xpath query for the device

Signed-off-by: BlackEagle <ike.devolder@gmail.com>

### DIFF
--- a/lib/vagrant-libvirt/action/start_domain.rb
+++ b/lib/vagrant-libvirt/action/start_domain.rb
@@ -47,14 +47,16 @@ module VagrantPlugins
 
           # additional disk bus
           config.disks.each do |disk|
-            device = disk[:device]
-            bus = disk[:bus]
-            REXML::XPath.each(xml_descr, '/domain/devices/disk[@device="disk"]/target[@dev="' + device + '"]') do |disk_target|
-              next unless disk_target.attributes['bus'] != bus
-              @logger.debug "disk #{device} bus updated from '#{disk_target.attributes['bus']}' to '#{bus}'"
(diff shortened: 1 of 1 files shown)


electrofelix:
This should be already fixed by #1680, though not yet released. You can install the plugin from a local gen by downloading the latest asset from https://github.com/vagrant-libvirt/vagrant-libvirt/packages/1659776 to confirm it is fixed for you.

Unfortunately vagrant doesn't support installing from private rubygems repositories and GitHub doesn't support making rubygem package repositories public.

electrofelix:
No need to apologise, I need to update the docs and issue template so that the availability of a gem of the latest code to be installed is shared

PR #1702, opened 2022-12-24
Fix issue where "iommu" was configured for non-virtio network devices

Terr:
When setting the `model_type` of a network to something else than `virtio`, the `iommu` parameter was still set for the device which isn't allowed.

I noticed that the if-statement for setting the parameter was looking at the default NIC model type (in `@nic_model_type`) instead of final, configured value in `@model_type`. This PR changes that.

I wanted to add a test but my experience with the Ruby world is 0. I couldn't find an existing test that was easily adopted. I think I only saw tests that go from XML -> configuration object, and not the other way around? If you can guide me in where/how I could add a test I'd happily do so.


electrofelix:
I think the network tests are a bit lacking to be asking for you to add to them. The change is sufficiently obviously correct that I think we can merge without adding a specific test for it.

PR #1706, opened 2023-01-16
[Fix] forward_ports.rb@redirect_port: no implicit conversion of String into Array (TypeError)

btall:
Hi 👋🏾 ,

When attempting to do port forwarding with the `gateway_ports` option enabled, this error occurs.

Screenshot_2023-01-16_12-29

Regards

### DIFF
--- a/lib/vagrant-libvirt/action/forward_ports.rb
+++ b/lib/vagrant-libvirt/action/forward_ports.rb
@@ -93,7 +93,7 @@ module VagrantPlugins
             -N
             #{ssh_info[:host]}
           )
-          params += '-g' if gateway_ports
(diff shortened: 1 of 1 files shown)


lib/vagrant-libvirt/action/forward_ports.rb
```diff
@@ -93,7 +93,7 @@ def redirect_port(machine, host_ip, host_port, guest_ip, guest_port,
             -N
             #{ssh_info[:host]}
           )
-          params += '-g' if gateway_ports
+          params <<= '-g' if gateway_ports
```
electrofelix, line 96: while I was familiar with with `<<` I had to check about `<<=`, thanks for the fix

PR #1715, opened 2023-01-26
Add feature-complete /domain/clock support

bkus:
Feature reference: https://libvirt.org/formatdomain.html#time-keeping

### DIFF
--- a/docs/configuration.markdown
+++ b/docs/configuration.markdown
@@ -1046,9 +1046,18 @@ end
 
 ## Clock
 
-Clock offset can be specified via `libvirt.clock_offset`. (Default is utc)
+The clock can be configured using one of the following methods:
 
-Additionally timers can be specified via `libvirt.clock_timer`.
+* Set nothing, and the clock will default to UTC.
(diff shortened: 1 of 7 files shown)


bkus:
@electrofelix Is there anything else I should do to have this reviewed?

electrofelix:
@bkus no, I'm just not available at the moment and for the next couple of weeks as much as I would usually would be to review/merge. I think the only nice addition would be to add an entry to the validate function to alert users that only one of the options clock_absolute, clock_adjustment, clock_basis, or clock_offset is valid at any given time https://github.com/vagrant-libvirt/vagrant-libvirt/blob/6b907ab6e050d44e5f8ae24a90d54d3784adc33f/lib/vagrant-libvirt/config.rb#L1141

As otherwise people may set an option that is ignored without realizing it. I usually find it's better to be alerted up front that it's not valid to have two options that cannot be set together to be rejected and then decide.

bkus:
@electrofelix Good idea.  Validator has been added.

PR #1722, opened 2023-02-14
Fix duplicate provisioning of running VMs

### LINKED ISSUE #1710: Ansible playbooks run twice when re-provisioning
**Describe the bug**
When provisioning VMs with Ansible, the specified playbook(s) are run twice when re-provisioning.

This seems to occur starting with version `0.8.1` of the plugin (was using `0.7.0` previously). Using a gem of the latest code doesn't remedy the problem either.

**To Reproduce**
1. Create the following Vagrantfile and its provisioner playbook (named: `playbook.yml`)
2. Run `vagrant up --provision` or `vagrant up`
3. Notice how the play only ran _once_
4. Run `vagrant up --provision` (again)
5. Notice how the play only ran **_twice_**

**Vagrantfile**
```ruby
VAGRANTFILE_API_VERSION = "2"

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
  config.vm.box = "debian/bullseye64"
  config.vm.synced_folder ".", "/vagrant", disabled: true

  config.vm.provider "libvirt" do |domains|
    domains.default_prefix = "playground-ansible_"
    domains.memory = 512
    domains.machine_virtual_size = 10 # GBs
  end

  config.vm.define "vm1" do |machine|
    machine.vm.hostname = "vm1"

    machine.vm.provider "libvirt" do |domain|
      domain.cpus = 1
    end

    machine.vm.provision "ansible" do |ansible|
      ansible.playbook = "playbook.yml"
      ansible.limit = "all"
    end
  end
end

```

**playbook.yml**
```yml
---
- name: Run play
  hosts: all

  tasks:
    - name: Show message 'foo'
      ansible.builtin.debug:
        msg: foo
```

**Expected behavior**
I expect the playbooks to run once when provisioned again.

**Versions (please complete the following information):**:
- Libvirt version: 8.0.0
- Vagrant version [output of `vagrant version`]: 2.2.19
- Vagrant flavour [Upstream or Distro]: Upstream package (https://apt.releases.hashicorp.com)
    - will add that I tried using my distro's Vagrant (Ubuntu), same outcome

- Vagrant plugins versions (including vagrant-libvirt) [output of `vagrant plugin list`]:
```
vagrant-libvirt (0.11.3.pre.12, global)
  - Version Constraint: 0.11.3.pre.12
```

**Debug Log**
[vagrant.log](https://github.com/vagrant-libvirt/vagrant-libvirt/files/10470716/vagrant.log)

cavcrosby:
Fixes #1710.

I found either deleting the block mentioned in `actions.rb` or removing calls to `Provision` from blocks that also call `action_start` to fix the issue. That said, I went with @electrofelix's recommendation on the former.

I _attempted_ to add behavior to the unit tests that would check that for the actions tested, they would not call other action methods unless explicitly needed to. Though it sounds like actions can call each other _at least_ once. This might not be whats desired so suggestions would be highly appreciated.

### DIFF
--- a/lib/vagrant-libvirt/action.rb
+++ b/lib/vagrant-libvirt/action.rb
@@ -128,9 +128,7 @@ module VagrantPlugins
       private_class_method def self.action_start
         Vagrant::Action::Builder.new.tap do |b|
           b.use Call, IsRunning do |env, b2|
-            # If the VM is running, run the necessary provisioners
             if env[:result]
-              b2.use action_provision
               next
             end
 
--- a/spec/unit/action_spec.rb
+++ b/spec/unit/action_spec.rb
@@ -58,7 +58,7 @@ describe VagrantPlugins::ProviderLibvirt::Action do
   end
 
   def receive_and_call_next(&block)
-    return receive(:call) { |cls, env| call_next(cls, env, &block) }
+    return receive(:call) { |cls, env| call_next(cls, env, &block) }.exactly(1).times
   end
 
   def call_next(cls, env)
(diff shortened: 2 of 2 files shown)


lib/vagrant-libvirt/action.rb
```diff
@@ -128,12 +128,6 @@ def self.action_up
       private_class_method def self.action_start
         Vagrant::Action::Builder.new.tap do |b|
           b.use Call, IsRunning do |env, b2|
-            # If the VM is running, run the necessary provisioners
-            if env[:result]
-              b2.use action_provision
-              next
```
electrofelix, line 134: I hadn't actually checked whether `next` could be removed as well.

electrofelix (commented):
I probably should have been a little more clear, I meant that the individual expects should either have `.exactly(1).times` appended to each, e.g.:
```
expect_any_instance_of(VagrantPlugins::ProviderLibvirt::Action::Provision).to receive_and_call_next.exactly(1).times
```
Which would ensure that a anything calling Provision a second time would be treated as an error.

Or possibly could be handled across the board using:
```
diff --git spec/unit/action_spec.rb spec/unit/action_spec.rb
index 7459d38..46b9b42 100644
--- spec/unit/action_spec.rb
+++ spec/unit/action_spec.rb
@@ -58,7 +58,7 @@ describe VagrantPlugins::ProviderLibvirt::Action do
   end
 
   def receive_and_call_next(&block)
-    return receive(:call) { |cls, env| call_next(cls, env, &block) }
+    return receive(:call) { |cls, env| call_next(cls, env, &block) }.exactly(1).times
   end
```

What would still be missing is though is a test on the `action_reload` similar to the existing tests on `action_up`, `action_halt`, `action_destroy`, etc.

lib/vagrant-libvirt/action.rb
cavcrosby: yea, just removing `b2.use action_provision` and leaving the rest of the block still solved the issue when testing. I just readded it back unless you feel otherwise.

PR #1729, opened 2023-03-22
Feature: Allow specifying type for primary disk

kdunn926:
This PR has changes to allow specifying the `type` for `disk_driver_opts`, rather than hardcoding it to `qcow2`.

### DIFF
--- a/lib/vagrant-libvirt/templates/domain.xml.erb
+++ b/lib/vagrant-libvirt/templates/domain.xml.erb
@@ -152,12 +152,11 @@
 <%- @domain_volumes.each_with_index do |volume, index| -%>
     <disk type='file' device='disk'>
       <alias name='ua-box-volume-<%= index -%>'/>
-      <driver name='qemu'<%=
-        '%{type}%{cache}%{io}%{copy_on_read}%{discard}%{detect_zeroes}' \
-          % { :type => 'qcow2', :cache => "#{volume[:cache]}", :io => nil, :copy_on_read => nil, :discard => nil, :detect_zeroes => nil }
-            .merge(@disk_driver_opts)
-            .map { |k,v| v.nil? ? [k, ""] : [k, " #{k}='#{v}'"]}.to_h
-      -%>/>
+      <driver name='qemu' <%=
+        {:type => 'qcow2', :cache => "#{volume[:cache]}"}.merge(@disk_driver_opts)
+          .reject { |k,v| v.nil? }
+          .map { |k,v| "#{k}='#{v}'"}
+          .join(' ') -%>/>
       <source file='<%= volume[:absolute_path] %>'/>
       <%# we need to ensure a unique target dev -%>
       <target dev='<%= volume[:device] %>' bus='<%= volume[:bus] %>'/>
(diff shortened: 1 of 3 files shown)


lib/vagrant-libvirt/templates/domain.xml.erb
```diff
@@ -152,11 +152,13 @@
...
-        @disk_driver_opts.empty? ? "cache='#{volume[:cache]}'" :
-            @disk_driver_opts.reject { |k,v| v.nil? }
-                             .map { |k,v| "#{k}='#{v}'"}
-                             .join(' ') -%>/>
+      <driver name='qemu'<%=
+        '%{type}%{cache}%{io}%{copy_on_read}%{discard}%{detect_zeroes}' \
+          % { :type => 'qcow2', :cache => "#{volume[:cache]}", :io => nil, :copy_on_read => nil, :discard => nil, :detect_zeroes => nil }
+            .merge(@disk_driver_opts)
+            .select { |k,_| [:type, :cache, :io, :copy_on_read, :discard, :detect_zeroes].include? k }
+            .map { |k,v| v.nil? ? [k, ""] : [k, " #{k}='#{v}'"]}.to_h
```
electrofelix, line 160: I didn't think you need to interpolate all of the results, think the following achieves the same without needing to specify all of the options either:
```
      <driver name='qemu' <%=
        {
            :type => 'qcow2',
            :cache => "#{volume[:cache]}",
         }.merge(@disk_driver_opts).
             reject { |k,v| v.nil? }.
             map { |k,v| "#{k}='#{v}'"}.
             join(' ') -%>/>
```

lib/vagrant-libvirt/templates/domain.xml.erb
```diff
@@ -152,11 +152,13 @@
...
-      <driver name='qemu' type='qcow2' <%=
-        @disk_driver_opts.empty? ? "cache='#{volume[:cache]}'" :
-            @disk_driver_opts.reject { |k,v| v.nil? }
-                             .map { |k,v| "#{k}='#{v}'"}
-                             .join(' ') -%>/>
+      <driver name='qemu'<%=
+        '%{type}%{cache}%{io}%{copy_on_read}%{discard}%{detect_zeroes}' \
+          % { :type => 'qcow2', :cache => "#{volume[:cache]}", :io => nil, :copy_on_read => nil, :discard => nil, :detect_zeroes => nil }
+            .merge(@disk_driver_opts)
+            .select { |k,_| [:type, :cache, :io, :copy_on_read, :discard, :detect_zeroes].include? k }
```
electrofelix, line 159: The valid options are already controlled from config.rb, I'd prefer not to be hardcoding them here again.

electrofelix (changes requested):
I'd prefer to avoid adding the list of options back to the domain template when they are already managed from the config file, made a suggestion that should resolve.

lib/vagrant-libvirt/templates/domain.xml.erb
kdunn926: The approach used in the PR was to specifically to avoid sorting issues that generated keys in a different order than the test artifacts. I think if the original `map()/join()` approach is used the test artifacts have to be modified instead. I'm perfectly happy with whichever you think is better.

lib/vagrant-libvirt/templates/domain.xml.erb
kdunn926: Ah yes, I see how this is redundant.

electrofelix:
Sorry been busy with new job for a while, spotted that the docs was switching from the config option that end users should call `disk_driver` to the internal attribute that is updated by the function called when `disk_driver` is configured.

lib/vagrant-libvirt/templates/domain.xml.erb
electrofelix: Think I'm managed to update in a way that avoid impacting the tests while also avoiding explicit references to the only config options that are valid.

PR #1731, opened 2023-04-08
specify qcow2 compatibility version to not use very old 0.10 fallback

hfinger:
Libvirt creates qcow2 volumes with very old `compat` version `0.10` if no `compat` version is specified in the xml (see [here](https://libvirt.org/formatstorage.html#id11)). It seems libvirt has backwards compatibility reasons to still use `0.10` if nothing is specified in the xml.

Unfortunately this has the effect that VMs created by vagrant-libvirt are using the 10 year old `0.10` qcow2 version, instead of the `1.1` version that has **better performance and stability fixes**. 
This is even happening when the backing file of the vagrant box is in compat `1.1` format.

This is the `qemu-img info` of the backing file provided by the vagrant box:
```
# qemu-img info /var/lib/libvirt/images/vagrant_box_image_0_box.img
[...]
Format specific information:
    compat: 1.1
    [...]
```
Although the backing file provided by the box is using compat `1.1`, the created volume by vagrant-libvirt has only compat `0.1` as shown here:
```
# qemu-img info /var/lib/libvirt/images/vagrant-01-test_default.img
[...]
backing file: /var/lib/libvirt/images/vagrant_box_image_0_box.img
Format specific information:
    compat: 0.10
    [...]
```

With this MR, the volume of the VM will use the same `compat` version as provided by the backing image of the vagrant box.
If `qemu-img info` does not return a `compat` flag, it falls back to the very old `0.10` version.

### DIFF
--- a/lib/vagrant-libvirt/action/create_domain_volume.rb
+++ b/lib/vagrant-libvirt/action/create_domain_volume.rb
@@ -60,6 +60,7 @@ module VagrantPlugins
                       xml.group storage_gid(env)
                       xml.label 'virt_image_t'
                     end
+                    xml.compat '1.1'
                   end
                   xml.backingStore do
                     xml.path(@backing_file)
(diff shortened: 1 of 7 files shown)


electrofelix (approved):
Thanks, that should be a nice improvement for everyone to pick up.

PR #1784, opened 2023-10-15
fix additional disk path support

### LINKED ISSUE #259: Creating additional disks belongs in a seperate action
#178 added the ability to create additional disks. The code that was added to create_domain should be in  new action,  similar to create_domain_volume or create_network_interfaces. create_domain should be limited to creating the VM itself.

We should also look at moving methods added to config into another module.

### LINKED ISSUE #1783: additional disk configs aren't well aligned to libvirt
**Is your feature request related to a problem? Please describe.**

libvirt assumes volumes exist only in storage pools, making `:path` a confusing option. 

**Describe the solution you'd like**

Instead, `:name` should be exposed to users and `:path` should support uploading images
when the provided `:path` exists.

**Describe alternatives you've considered**

Using fog directly in the vagrantfile to prepare the storage pool. It's a difficult workaround and this is a better long term solution.

cornfeedhobo:
This PR seeks to add support for importing volumes when `disk[:path]` exists.

***Disclaimer: tests are still not succeeding, and I could use some help figuring it out. If anyone wants to help, drop me a note here, or on libera irc.***

Fixes #259
Fixes #1783


cornfeedhobo:
@purpleidea @sciurus @infernix @electrofelix

I've seen you comment or work on related issues. If you have time to review this, I would appreciate it. Thanks.

purpleidea:
@cornfeedhobo 

> I've seen you comment or work on related issues. If you have time to review this, I would appreciate it. Thanks.

I haven't hacked on it recently, but until vagrant is under a real Free Software license, I'm definitely out. Sorry.

cornfeedhobo:
@purpleidea I hear you, but [you did say to ping you for a review for this one specific case](https://github.com/vagrant-libvirt/vagrant-libvirt/issues/259#issuecomment-59674314) :sweat_smile:

purpleidea:
haha! So true! Well it was _2014_ so things have changed slightly. I can help you with an alternative idea for rapid virt provisioning if you are interested. Ping me elsewhere if you want to chat about it. Cheers

cornfeedhobo:
@sciurus @infernix @electrofelix Any hope?

electrofelix:
@cornfeedhobo right now I'm pretty slammed with two young children under 2 years of age, and I don't expect to have time for any projects for at least another 3 months, though could be longer.

Given time constraints, as I hopefully start to get a little time in the future, smaller PRs will be picked up first just due most likely they are the ones I can review in the time available. I don't want to suggest splitting this PR into a load of separate PRs as that could be a lot of work and I may still not be able to review them in a timely manner. Though maybe there is one commit you can push up as a separate PR that is nice and small and I can try with that. Keep this one up to date and then we can try and iterate through the individual commits as PRs. I will try to look at just one or two of the early commits to do a first pass there first to see if I can manage it from my phone before you do a chunk of work.

I thought there was an option for triggering the remaining workflows to ensure that you'd have a gem published to the GitHub rubygem packages and could point users to the instructions to install from it for a while, not sure what is going on there.

I can give some general comments about what I prefer to see:.
* Avoid adding more references to fog in variable names. At some point the backend should be abstracted to allow rewriting to use xmlrpc or possibly golang to allow a conversion. It makes sense to avoid saying something is a fog volume if what will create it in the future will change. Where just moving exisiting code around that already has these, don't worry about them, it's just new references.
* Try to avoid moving code to send messages to the UI into until functions, best to keep these within the action calling. Not always possible, but generally preferred as then the actions can decide what to display. Typically I'd think a callback for those pieces that can't be moved where the default is to pass an empty object, test where it is UI type, and only write something like progress to it in that case.

spec/acceptance/snapshots_spec.rb
```diff
@@ -31,6 +31,10 @@
     status('Test: restore snapshot')
     expect(environment.execute('vagrant', 'snapshot', 'restore', 'test')).to exit_with(0)
 
+    # KVM needs a moment for IO to catch up.
+    # If anyone comes up with a better way to wait, please update this.
+    sleep(3)
```
electrofelix, line 36: I suspect an email to get libvirt devs would be needed. I assumed that revert was blocking, but perhaps that doesn't cover flushing or the state being fully switched. Hopefully they could enlighten on what would need to be done after the revert call to confirm the revert is fully completed.

This should be good enough for the tests for now though.

spec/acceptance/support-skeletons/additional_storage/Vagrantfile
```diff
@@ -5,6 +5,7 @@
 
 Vagrant.configure("2") do |config|
   config.vm.box = "infernix/tinycore"
+  config.vm.box_check_update = false
```
electrofelix, line 8: Meant to ask as well, how does this help with stability? Or is it just handy to disable when running a lot?

Generally I'm thinking that the CI here should be using the latest images rather than those that it was run with the first time. There should be caching in place to avoid this costing a lot of time by default, bit of it's useful to disable in certain cases I could see using an env var instead being better.

cornfeedhobo: I was curious if this was intentional. Maybe I can switch to using an environment variable to toggle? Either way you hit the nail on the head - it was painful when running the test suite dozens of times in a day.

I'll put some thinking into this one.

PR #1787, opened 2023-11-05
Allow video resolution hinting

trams242:
Add support for custom video resolution

Libvirt supports setting custom video resolution by setting the
resolution subelement in the model tag. This changes the preferred EDID
mode exposed to the vm.

Changing this is usefull when running GUI systems that does not provide
UI elements for changing the resolution. Examples would be cage or other
wayland compositors where initialization uses the preferred EDID mode.

### DIFF
--- a/docs/configuration.markdown
+++ b/docs/configuration.markdown
@@ -194,6 +194,10 @@ end
   "cirrus", "vmvga", "xen", "vbox", or "qxl".
 * `video_vram` - Used by some graphics card types to vary the amount of RAM
   dedicated to video.  Defaults to 16384.
+* `video_resolution_x` - sets the default EDID/monnitor resolution hinting for the X
+  axis. Defaults to qemu default. (1280x800).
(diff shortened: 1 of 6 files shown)


docs/configuration.markdown
```diff
@@ -194,6 +194,10 @@ end
   "cirrus", "vmvga", "xen", "vbox", or "qxl".
 * `video_vram` - Used by some graphics card types to vary the amount of RAM
   dedicated to video.  Defaults to 16384.
+* `video_resolution_x` - sets the default EDID/monnitor resolution hinting for the X
```
electrofelix, line 197: While spitting the resolution into two separate variables makes it easier in some ways, I'm not sure it's particularly consistent with how many would expect it to work with `1280x800` being a more common form. Should be possible to validate the value matches `[[:digit:]]{3,}x[[:digit:]]{3,}` (or similar regex) if not nil in the config class and subsequently split when specifying the individual XML elememts

Issue #1725, opened 2023-03-02
qemu: could not load PC BIOS '/usr/share/OVMF/OVMF_CODE_4M.fd'


ju2wheels:
**Describe the bug**
I am unable to get the qemu to load the OVMF UEFI BIOS even though its the same one I select via virt-manager.

I am getting the following output:
```
==> box: Successfully added box 'tvision/tvision-vagrant-libvirt-ubuntu-22.04-amd64-ansible-2.14.1' (v2.9.0) for 'libvirt'!
Bringing machine 'ansible_test' up with 'libvirt' provider...
==> ansible_test: Checking if box 'tvision/tvision-vagrant-libvirt-ubuntu-22.04-amd64-ansible-2.14.1' version '2.9.0' is up to date...
==> ansible_test: Uploading base box image as volume into Libvirt storage...
==> ansible_test: Creating image (snapshot of base box volume).
==> ansible_test: Creating domain with the following settings...
==> ansible_test:  -- Name:              ansible-galaxy-tvision.repository_ansible_test
==> ansible_test:  -- Description:       Source: /home/jlajara/repos/TVision/ansible-galaxy-tvision.repository/Vagrantfile
==> ansible_test:  -- Domain type:       kvm
==> ansible_test:  -- Cpus:              1
==> ansible_test:  -- Feature:           acpi
==> ansible_test:  -- Feature:           apic
==> ansible_test:  -- Feature:           pae
==> ansible_test:  -- Clock offset:      utc
==> ansible_test:  -- Memory:            512M
==> ansible_test:  -- Loader:            /usr/share/OVMF/OVMF_CODE_4M.fd
==> ansible_test:  -- Base box:          tvision/tvision-vagrant-libvirt-ubuntu-22.04-amd64-ansible-2.14.1
==> ansible_test:  -- Storage pool:      default
==> ansible_test:  -- Image(vda):        /var/lib/libvirt/images/ansible-galaxy-tvision.repository_ansible_test.img, virtio, 59G
==> ansible_test:  -- Disk driver opts:  cache='default'
==> ansible_test:  -- Graphics Type:     spice
==> ansible_test:  -- Graphics Websocket: 
==> ansible_test:  -- Graphics Port:     
==> ansible_test:  -- Graphics IP:       
==> ansible_test:  -- Graphics Password: Not defined
==> ansible_test:  -- Video Type:        qxl
==> ansible_test:  -- Video VRAM:        16384
[...]

Issue #1723, opened 2023-02-19
Wrong domain definition with UEFI boot


jeka:
**Describe the bug**
Vagrant generates wrong domain XML when setting up domain with UEFI boot.

**To Reproduce**

with Vagrantfile: (adopted [Secure Encryption Virtualization (SEV))](https://vagrant-libvirt.github.io/vagrant-libvirt/examples.html#secure-encryption-virtualization-sev)
```
Vagrant.configure("2") do |config|
  config.vm.provider :libvirt do |libvirt|
    libvirt.loader = "/usr/share/edk2/ovmf/OVMF_CODE.cc.fd"
    libvirt.nvram = "/usr/share/edk2/ovmf/OVMF_VARS.fd"
    libvirt.machine_type = "q35"
  end

  config.vm.define "uefi-node" do |vm|
    vm.vm.provider :libvirt do |domain|
      domain.cpus = 1
      domain.memory = 256
      domain.storage :file, :size => '1G', :device => 'sda', :type => 'raw', :bus => 'scsi', :discard => 'unmap', :detect_zeroes => 'on'
    end
  end

end
```
vagrant generates domain XML with **os** section as follows:
```
  <os>
    <type arch='x86_64' machine='pc-q35-rhel8.6.0'>hvm</type>
    <loader readonly='yes' type='pflash'>/usr/share/edk2/ovmf/OVMF_CODE.cc.fd</loader>
    <nvram>/usr/share/edk2/ovmf/OVMF_VARS.fd'</nvram>
    <boot dev='hd'/>
    <bootmenu enable='no'/>
  </os>
```
Which is wrong. While that will work if you are booting only 1 machine in UEFI mode, it still modifies a file that should not be touched.

The correct XML part would be:
```
  <os>
    <type arch='x86_64' machine='pc-q35-rhel8.6.0'>hvm</type>
    <loader readonly='yes' type='pflash'>/usr/share/edk2/ovmf/OVMF_CODE.cc.fd</loader>
    <nvram template='/usr/share/edk2/ovmf/OVMF_VARS.fd'>/var/lib/libvirt/qemu/nvram/bug_uefi-node-VARS.fd</nvram>
    <boot dev='hd'/>
    <bootmenu enable='no'/>
```
The result should be that every single machine booted in UEFI mode should have dedicated EFI vars storage.
I guess `/var/lib/libvirt/qemu/nvram` is specific to my host OS  (CentOS Stream 8)

[...]

PR #1789, opened 2023-11-11
Add nvram template configuration option

sony-aufsehe:
Adds an nvram_template option to allow dynamic creation of nvram file in /var/lib/libvirt/qemu/nvram based on OVMF VARS.
Provides a way to launch multiple domain with a same var template.
This may be a solution for https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1725 and https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1723.


electrofelix:
I appreciate I'm not able to timely review anything due to small child duties, could you share a bit about how it turned out this doesn't solve the problem. Just in case someone else is able to pick and help move it forward

sony-aufsehe:
I stumbled upon contributiuon instructions and decided to resubmit following it from a feature branch. Seems i have not gone through with resubmission. This fix works great for me so i'll resubmit the pull request.
EDIT: reopened pull request is available here : https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1797

PR #1793, opened 2023-12-25
Prevent error on windows virtiofs mount

z0rb:
After installing WinFSP and the virtiofs driver in the provisioning phase, then configuring the service to autostart with `sc config VirtioFsSvc binPath="C:\Program Files\Virtio-Win\VioFS\virtiofs.exe" start=auto depend=VirtioFsDrv` ... the virtiofs share gets mounted as a separate drive in Windows 10.  
Guide: https://github.com/virtio-win/kvm-guest-drivers-windows/wiki/Virtiofs:-Shared-file-system#setup-with-installer

image

I have no experience writing vagrant plugins. In my fork I have just tried to implement a NOOP so that the libvirt plugin doesn't error out because of the missing guest capability. This works for me as `vagrant up` succeeds and the share works properly.

Question for maintainers. Would this be ok to merge into main, to prevent the error? The capability obviously exists, but I do not know what would be acceptable for main.

### DIFF
--- a/lib/vagrant-libvirt/cap/mount_virtiofs.rb
+++ b/lib/vagrant-libvirt/cap/mount_virtiofs.rb
@@ -34,6 +34,14 @@ module VagrantPlugins
           end
         end
       end
+
+      class MountVirtioFSWin
+        extend Vagrant::Util::Retryable
+
+        def self.mount_virtiofs_shared_folder(machine, folders)
+          # Do nothing
+        end
+      end
     end
   end
 end
--- a/lib/vagrant-libvirt/plugin.rb
+++ b/lib/vagrant-libvirt/plugin.rb
@@ -41,6 +41,8 @@ module VagrantPlugins
       end
       
       guest_capability('windows', 'mount_virtiofs_shared_folder') do
+        require_relative 'cap/mount_virtiofs'
+        Cap::MountVirtioFSWin
       end
 
       provider_capability(:libvirt, :nic_mac_addresses) do
--- a/lib/vagrant-libvirt/plugin.rb
+++ b/lib/vagrant-libvirt/plugin.rb
@@ -39,9 +39,8 @@ module VagrantPlugins
         require_relative 'cap/mount_virtiofs'
         Cap::MountVirtioFS
       end
+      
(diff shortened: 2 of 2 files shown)


z0rb:
As a detail, the box I am using was build from the main branch of: https://github.com/ruzickap/ansible-role-virtio-win
With the addition of the following tasks:
```
  tasks:
    - name: Download WinFSP
      win_get_url:
        url: https://github.com/winfsp/winfsp/releases/download/v2.0/winfsp-2.0.23075.msi
        dest: "%HOMEDRIVE%\\winfsp.msi"
      register: result
      until: result is succeeded

    - name: Install WinFSP
      win_command: msiexec /i winfsp.msi /qn
      args:
        chdir: C:\

    - name: Download VirtioFS
      win_get_url:
        url: https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/latest-virtio/virtio-win-guest-tools.exe
        dest: "%HOMEDRIVE%\\virtio-win-guest-tools.exe"
      register: result
      until: result is succeeded
      
    - name: Install VirtioFS
      win_command: virtio-win-guest-tools.exe -s
      args:
        chdir: C:\

    - name: Configure VirtioFS Service
      win_command: sc config VirtioFsSvc binPath="C:\Program Files\Virtio-Win\VioFS\virtiofs.exe" start=auto depend=VirtioFsDrv
```

Dragonsong3k:
I guess this never got merged in? I am having the same exact problem now. 

vagrant up -> hostname changes -> causes reboot -> capabilities check happens. Error out that mounting is not supported in Windows. 

```
==> vagrant-dev-vm: Mounting virtiofs shared folders...
Vagrant attempted to execute the capability 'mount_virtiofs_shared_folder'
on the detect guest OS 'windows', but the guest doesn't
support that capability. This capability is required for your
configuration of Vagrant. Please either reconfigure Vagrant to
avoid this capability or fix the issue by creating the capability.
```

The mounting actually happens in the OS before the rename step by vagrant and works fine. Once the rename happens - Error above.

I tried using the mount: false option as described in the docs but it seems to ignore it.

Issue #1792, opened 2023-12-10
Can I set the DHCP lease time via vagrant-libvirt?


johanneskastl:
**Is your feature request related to a problem? Please describe.**

TL;DR: Can I influence the DHCP leasetime for the network created by vagrant-libvirt?

When using vagrant-libvirt on e.g. a laptop, the VM's IP addresses might change if you put the laptop to sleep/suspend and wake it up again. Apparently, the DHCP lease time is only 1 hour,  so after that the IPs might change.

In my case it is not possible to set static IPs, so I would like to increase the DHCP lease time to e.g. 24h or even 1 week.

**Describe the solution you'd like**
Apparently one can set lots of things on private networks, but not for the public network.
https://vagrant-libvirt.github.io/vagrant-libvirt/configuration.html#private-network-options
https://vagrant-libvirt.github.io/vagrant-libvirt/configuration.html#public-network-options

Although there are some DHCP-related options for private networks, there is nothing regarding the leasetime.

A sample Vagrantfile without the network options might look like this:

```
Vagrant.configure("2") do |config|

  # name the VMs
  config.vm.define "leap155" do |node|

    # which image to use
    node.vm.box = "opensuse/Leap-15.5.x86_64"

    # disable synced folders
    node.vm.synced_folder ".", "/vagrant", disabled: true

    # sizing of the VMs
    node.vm.provider "libvirt" do |lv|
      lv.random_hostname = false
      lv.memory = 2048
      lv.cpus = 2
    end

    # set the hostname
    node.vm.hostname = "leap155"

  end # config.vm.define nodes
end
```

This results in a new network `vagrant-libvirt` being created whose xml looks like this:

```
network ipv6='yes'>                                                                                                                                                                         
  <name>vagrant-libvirt</name>                                                                                                                                                               
  <uuid>34d7cbbd-4cf4-4c4f-b9f6-8524f9101fc1</uuid>                                                                                                                                          
  <forward mode='nat'/>                                                                                                                                                                      
  <bridge name='virbr3' stp='on' delay='0'/>                                                                                                                                                 
  <mac address='52:54:00:77:6b:c8'/>                                                                                                                                                         
  <ip address='192.168.121.1' netmask='255.255.255.0'>                                                                                                                                       
    <dhcp>                                                                                                                                                                                   
      <range start='192.168.121.1' end='192.168.121.254'/>                                                                                                                                   
    </dhcp>                                                                                                                                                                                  
[...]

PR #1799, opened 2024-02-02
allow setting a default lease time for networks

johanneskastl:
see #1792

### DIFF
--- a/lib/vagrant-libvirt/templates/private_network.xml.erb
+++ b/lib/vagrant-libvirt/templates/private_network.xml.erb
@@ -25,7 +25,7 @@
     <% end %>
     <% if @network_dhcp_enabled  %>
       <dhcp>
-        <range start="<%= @network_range_start %>" end="<%= @network_range_stop %>" />
+        <range start="<%= @network_range_start %>" end="<%= @network_range_stop %>">
         <% if @network_dhcp_bootp_file  %>
           <% if @network_dhcp_bootp_server  %>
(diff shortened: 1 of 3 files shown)


johanneskastl:
I think the errors are not from my changes...

lib/vagrant-libvirt/templates/private_network.xml.erb
```diff
@@ -33,6 +33,9 @@
             <bootp file="<%= @network_dhcp_bootp_file %>" />
           <% end %>
         <% end %>
+        <% if @network_dhcp_default_lease_time %>
+            <lease expiry='@network_dhcp_default_lease_time' unit='hours'/>
```
electrofelix, line 37: From the issue discussion, this looked like it should be inside the range tag, while here it is only inside the dhcp tag. Appears would need to change how the range tag is closed to add this.

johanneskastl: Thanks for checking, nice catch! I addressed this by adding a closing `</range>` tag if @network_dhcp_default_lease_time is defined.

see 1aa7ed8

electrofelix: I think it'll require changing the range tag as well to remove the trailing `/>` to prevent it being closed before this as well. Presumably this means the current bootp XML is ignored as well

johanneskastl: True. If my XML is not off, commit 4c25b52 should fix this...

PR #1809, opened 2024-02-21
Implement creating sysinfo fwcfg entries

dev-zero:
With the following configuration this can be used to provide an Ignition configuration to boxes using `ignition.machine.id=qemu`:

```
    libvirt.sysinfo_fwcfg :name => "opt/com.coreos/config", :file => '/tmp/provision.ign'
```


dev-zero:
Alternatively we could do what the Terraform libvirt provider is doing and have only a single attribute `fw_cfg_name` to tune a single name, only support `:file` and then an abstract logic to automatically create an ISO on systems/hypervisors where fw_cfg is not supported.

PR #1815, opened 2024-04-05
Ignore networks that disappear during discovery

jamie01:
On systems where we have multiple concurrent deployments and teardowns of virtual machines that overlap, there are occasions where setup can fail if teardown of another vagrant deployment is happening in parallel. Generally this is handled well, but the most common place of failure was in the network enumeration where details of all existing networks are collected prior to creating new ones.

Since a number of operations are done after `driver.list_all_networks` and in cases where the number of networks is relatively large, it was common for a network to be destroyed during the iterations of that loop and thus the XML lookup or queries for `bridge_name` etc would fail.

My change here is just to rescue any errors that occur in this loop and continue on. I understand this is a bit of a niche issue for us in the way vagrant-libvirt is used at scale, and this might mask other problems besides networks being cleaned up, but opening this PR anyway for thoughts.

### DIFF
--- a/lib/vagrant-libvirt/util/network_util.rb
+++ b/lib/vagrant-libvirt/util/network_util.rb
@@ -194,40 +194,44 @@ module VagrantPlugins
 
           # Iterate over all (active and inactive) networks.
           driver.list_all_networks.each do |libvirt_network|
-
             # Parse ip address and netmask from the network xml description.
-            xml = Nokogiri::XML(libvirt_network.xml_desc)
-            ip = xml.xpath('/network/ip/@address').first
-            ip = ip.value if ip
-            netmask = xml.xpath('/network/ip/@netmask').first
-            netmask = netmask.value if netmask
+            begin
+              xml = Nokogiri::XML(libvirt_network.xml_desc)
+              end
+              ip = xml.xpath('/network/ip/@address').first
+              ip = ip.value if ip
+              netmask = xml.xpath('/network/ip/@netmask').first
+              netmask = netmask.value if netmask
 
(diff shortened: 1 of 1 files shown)


lib/vagrant-libvirt/util/network_util.rb
```diff
@@ -194,40 +194,44 @@ def libvirt_networks(driver)
...
-              bridge_name:      libvirt_network.bridge_name,
-              domain_name:      domain_name,
-              created:          true,
-              active:           libvirt_network.active?,
-              autostart:        libvirt_network.autostart?,
-              libvirt_network:  libvirt_network
-            }
+            begin
+              xml = Nokogiri::XML(libvirt_network.xml_desc)
+              end
```
abbbi: not quite sure about that one

abbbi:
```
diff --git a/lib/vagrant-libvirt/util/network_util.rb b/lib/vagrant-libvirt/util/network_util.rb
index ac1824a..0420241 100644
--- a/lib/vagrant-libvirt/util/network_util.rb
+++ b/lib/vagrant-libvirt/util/network_util.rb
@@ -196,7 +196,17 @@ module VagrantPlugins
           driver.list_all_networks.each do |libvirt_network|
 
             # Parse ip address and netmask from the network xml description.
-            xml = Nokogiri::XML(libvirt_network.xml_desc)
+            begin
+              xml = Nokogiri::XML(libvirt_network.xml_desc)
+            rescue Libvirt::Error
+                @logger.debug "Network disappeared while iterating, skipping"
+                next
+            end

```

PR #1818, opened 2024-05-11
Fix memory setting issue when starting domain

tsuchiclay:
By passing a flag to libvirt's API to apply the memory size change to the domain's configuration file, the following issue was solved.

 -  fail to update the domain.memory on existing running vm 'Call to virDomainSetMemory failed' #1812 https://github.com/vagrant-libvirt/vagrant-libvirt/issues/1812

The problem was caused by passing only memory size to #memory= method in start_domain.rb. Specifying a flag in addition to the memory size causes libvirt's API, virDomainSetMemoryFlags, to be called with that flag, explicitly choosing whether to apply the size setting to the running domain or the config file for that domain. The missing flag was causing the above issue by attempting to change the memory size of a running domain that did not exist.

Therefore, a list of possible flags is introduced, and a flag that make changes to the domain's configuration file is selected and used.

Below are links to some related documents.

 -  https://ruby.libvirt.org/api/Libvirt/Domain.html#method-i-memory-3D

 -  https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainSetMemoryFlags

### DIFF
--- a/lib/vagrant-libvirt/util/mod_impact_flags.rb
+++ b/lib/vagrant-libvirt/util/mod_impact_flags.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+# Ripped from https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainModificationImpact
+module VagrantPlugins
+  module ProviderLibvirt
+    module Util
+      module ModImpactFlags
+        VIR_DOMAIN_AFFECT_CURRENT = 0 # affect current domain state
+        VIR_DOMAIN_AFFECT_LIVE = 1 # affect running domain state
+        VIR_DOMAIN_AFFECT_CONFIG = 2 # affect persistent domain state
+      end
+    end
+  end
+end
--- a/lib/vagrant-libvirt/action/start_domain.rb
+++ b/lib/vagrant-libvirt/action/start_domain.rb
@@ -36,7 +36,7 @@ module VagrantPlugins
           if config.numa_nodes == nil
             if config.memory.to_i * 1024 != libvirt_domain.max_memory
               libvirt_domain.max_memory = config.memory.to_i * 1024
-              libvirt_domain.memory = libvirt_domain.max_memory
+              libvirt_domain.memory = libvirt_domain.max_memory, VagrantPlugins::ProviderLibvirt::Util::ModImpactFlags::VIR_DOMAIN_AFFECT_CONFIG
             end
           end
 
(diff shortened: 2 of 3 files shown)


lib/vagrant-libvirt/util/mod_impact_flags.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# Ripped from https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainModificationImpact
+module VagrantPlugins
+  module ProviderLibvirt
+    module Util
+      module ModImpactFlags
+        VIR_DOMAIN_AFFECT_CURRENT = 0 # affect current domain state
+        VIR_DOMAIN_AFFECT_LIVE = 1 # affect running domain state
+        VIR_DOMAIN_AFFECT_CONFIG = 2 # affect persistent domain state
```
electrofelix, line 10: Can you keep these with the rest of domain_flags.rb

electrofelix (commented):
Are you able to add an acceptance test that modified the memory and reloads? If you can add it in the first commit and then have this fix in the latter so that I can see the failure and then it passing it'll make it easy for me to just approve and merge without needing to wait until I've time to test it out myself.

It should be possible to copy https://github.com/vagrant-libvirt/vagrant-libvirt/blob/a94ce0d7b6c90129a38435698ca97b364130313d/spec/acceptance/provider_settings_spec.rb#L33 to add a reload test below it, just need to access the `environment.workdir` attribute for where to find the current run test `Vagrantfile` to modify before the reload call.

lib/vagrant-libvirt/util/mod_impact_flags.rb
tsuchiclay: Yes, let's do that.

tsuchiclay:
> Are you able to add an acceptance test that modified the memory and reloads? If you can add it in the first commit and then have this fix in the latter so that I can see the failure and then it passing it'll make it easy for me to just approve and merge without needing to wait until I've time to test it out myself.

Thank you! I'm on vacation right now and will be able to work on it the day after tomorrow.
Anyway, I'll try to add it.
~~Would it be less messy to redo those as separate commits (completely different from the past ones already posted above)?
I mean re-create a new pull request and close this place.~~
Nvm. I'm an obvious newbie to git.

tsuchiclay:
I mean re-create a new pull request and close this place.

electrofelix:
As long as you're comfortable with rebase swapping the two commits it should be fine. There shouldn't be any of the same files touched by the two commits so it should be a clean swap.

Issue #1821, opened 2024-06-02
Adding support for shmem


compromyse:
Looking-glass, needs an shmem device, which I don't see a way to add currently..
```
<shmem name='looking-glass'>
  <model type='ivshmem-plain'/>
  <size unit='M'>32</size>
</shmem>
```

I'd like to have an option to add an shmem device (or) a feature to arbitrarily add xml.

electrofelix:
Having an easy callback mechanism for users to mutate various XML definitions would be immensely useful, it's something that has come up before but I haven't had a chance to experiment with what it should look like.

compromyse:
Awesome, is there any workaround I can use for the time being?

PR #1822, opened 2024-06-02
Shmem support

compromyse:
Adds partial support for `shmem` device (no `role`, `server` or `msi`)


compromyse:
Closes #1821

compromyse:
https://libvirt.org/formatdomain.html#shared-memory-device

PR #1823, opened 2024-07-11
Use jq to parse qemu-img json output

koberbe:
Using awk to parse json is not very resistant to changes in formatting of the json input. Using jq to parse the json output makes the create_box.sh script more stable to change in qemu-img output format.


koberbe:
Maybe jq also needs to be added to dependencies?

